PR #9, opened 2019-11-18
include return value from hooked chat handler; this seems to fix the context menu issues

ff-meli:
I would ask that you look into whether this makes sense and/or causes issues.  It's rather tiny and does appear to fix the problem, but I don't have any real asm tools to actually look at the hooked code in a meaningful way.

I didn't test the reported eureka issues, but I had the same problem happening with the RMT reporting option and the view party finder option, which this seems to resolve.

I defaulted the return value to 0 because it seems like most (all?) messages that aren't actual chat or emotes will return this natively already.

### DIFF
--- a/Dalamud/Game/Internal/Gui/ChatGui.cs
+++ b/Dalamud/Game/Internal/Gui/ChatGui.cs
@@ -10,7 +10,7 @@ using Serilog;
 namespace Dalamud.Game.Internal.Gui {
     public sealed class ChatGui : IDisposable {
         [UnmanagedFunctionPointer(CallingConvention.ThisCall)]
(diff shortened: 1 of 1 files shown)


ff-meli:
Updated info: I tested in Eureka and it appeared to work there too.

I did notice a problem but I think it might be unrelated - if I use a release build of dalamud and the injector, most things work (including this fix), but chats are never relayed to discord.  Other events like duty pops go to discord correctly.  With a debug build, everything works as I would expect.

I saw the official release is also using a debug build, so maybe this is a known issue, but I wanted to point it out since the first time I ran into it was testing this particular change.

goaaats:
My deploys are usually on Release - maybe I forgot yesterday, I'll have a go later. Thank you so much!

PR #21, opened 2019-12-24
5.18 updates

ff-meli:
I picked 4.4.0 for assembly versions, but I wasn't sure what scheme you used for them.  I assumed lower numbers were for game hotfixes.

This actor table location is correct for nhaama for RP info (and there is a PR for it too). I didn't check if it worked properly here as well.

### DIFF
--- a/Dalamud.Injector/Dalamud.Injector.csproj
+++ b/Dalamud.Injector/Dalamud.Injector.csproj
@@ -14,10 +14,10 @@
   </PropertyGroup>
   <PropertyGroup Label="Feature">
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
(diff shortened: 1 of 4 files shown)


ff-meli:
Karashiiro's updates taken from sapphire discord and from https://github.com/goaaats/Dalamud/commit/0749a98058c36f97efd53719179783d77e1e8b9b and https://github.com/karashiiro/MachinaWrapperJSON/commit/f270d156f1711086708c1f67a0a473350d749d1d

goaaats:
We need CID and TerriType, I'll get those later tonight - really time to sig them. 
IconReplacer needs sigs but we should really move on with it as a plugin.

Thank you!

ff-meli:
Territory type is updated in nhaama in a different PR, not sure what CID is?

PR #26, opened 2020-01-18
Hooking via VTable ( As discussed in Discord )

Minizbot2012:
This PR changes the way that the interface is hooked, away from a signature scan to a VTable method.

The major changes is the addition of DXHookD3D11.cs; which utilizes a compact method (Compact = about 100 lines of code) to gain the Present function's address. The Interface manager now uses the VTable method to hook Present.

Alas this approach does come with a side effect of requiring sharpdx.desktop due to the method used to gain the address for Present.


Minizbot2012:
Where would I go to take grievance against one of your mods?
I do not appreciate being treated in this manner

goaaats:
Definitely not GitHub issues or PRs

PR #27, opened 2020-01-20
Beginnings of ImGuiScene cleanup

ff-meli:
This doesn't fundamentally change anything, but pulls in the start of some cleanup work in ImGuiScene and removes some things that are no longer necessary.

Mostly just trying to keep the library on parity for now/the initial beta testing release, so that there are fewer extra things to merge once I do a more thorough rework and cleanup pass.

### DIFF
--- a/Dalamud/Interface/InterfaceManager.cs
+++ b/Dalamud/Interface/InterfaceManager.cs
@@ -1,4 +1,5 @@
 using System;
+using System.IO;
 using System.Runtime.InteropServices;
 using Dalamud.Game;
 using Dalamud.Game.Internal.DXGI;
@@ -35,6 +36,7 @@ namespace Dalamud.Interface
 
         private ISwapChainAddressResolver Address { get; }
 
+        private Dalamud dalamud;
         private RawDX11Scene scene;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ff-meli:
Added another commit for setting the imgui ini path.  You might want to change how that is handled, but this seemed the best location for that file to go.

Dalamud/Interface/InterfaceManager.cs
```diff
@@ -38,15 +36,18 @@ public class InterfaceManager : IDisposable
...
+        private Dalamud dalamud;
         private RawDX11Scene scene;
 
         /// <summary>
         /// This event gets called by a plugin UiBuilder when read
         /// </summary>
         public event RawDX11Scene.BuildUIDelegate OnDraw;
 
-        public InterfaceManager(SigScanner scanner)
+        public InterfaceManager(Dalamud dalamud, SigScanner scanner)
```
silasary: Why not pass just dalamud, and get the scanner from `dalamud.sigScanner`?

Is there any time where you might want to pass a different SigScanner?

ff-meli: It's private and I try not to change interfaces unnecessarily.  This also follows the pattern of other objects initialized by dalamud.

goaaats: It does follow my pattern, but my pattern is just my pattern because I'm lazy :p
Either way it's fine really, if I ever do a general cleanup pass I'll change the DI setup into something more sensible

PR #40, opened 2020-02-29
Fix: pull in ImGuiScene update to fix custom unicode typed text being automatically processed and broken by win32

ff-meli:


ff-meli:
oh.. forgot pushing again would add to the PR.

The 3rd commit is a separate issue obviously.

PR #43, opened 2020-03-04
BGM and Injection

Aida-Enna:
- /xlbgmset 0 now restores the default BGM (kind of hacky I guess?)
- Added the BGM list (technically a link to the wiki question so we can update it if need be without a dalamud commit) to the ingame command
- Dalamud Injector no longer needs a PID specified and can be run without one to determine it automatically

Please look over the Dalamud Injector code - It seems to work fine but then in random weird cases it threw an error code 15, which could very well just be EasyHook being, well, EasyHook.

### DIFF
--- a/Dalamud.Injector/Program.cs
+++ b/Dalamud.Injector/Program.cs
@@ -24,8 +24,12 @@ namespace Dalamud.Injector {
                 Environment.Exit(0);
             };
 #endif
-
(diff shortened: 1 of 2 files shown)


Aida-Enna:
dab

daemitus:
Superseded by #396

PR #44, opened 2020-03-04
fix devPlugins not being able to load

ff-meli:
This didn't diff well - it's just moving the 'raw' flag inside the exists check, so that devPlugins (which sets raw) will not trigger the else block and fail out.


ff-meli:
Actually this is both wrong, and there are other issues here... oops XD

PR #46, opened 2020-03-05
Support IDA sigs that are based off of the first CALL instruction for…

attickdoor:
… a function.

### DIFF
--- a/Dalamud/Game/SigScanner.cs
+++ b/Dalamud/Game/SigScanner.cs
@@ -155,8 +155,23 @@ namespace Dalamud.Game {
 
             if (IsCopy)
                 scanRet = new IntPtr(scanRet.ToInt64() - this.moduleCopyOffset);
+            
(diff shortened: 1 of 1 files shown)


attickdoor:
Some background: For reasons I don't entirely understand, IDA only _sometimes_ provides a function sig that's the first unique bytes of a function. Sometimes, it provides a sig that looks something like this:
`E8 ? ? ? ? [other bytes]`
E8 is the opcode for CALL, and the four ?s are indicative of an address. What this type of sig is doing is pointing towards the location of the first place a function gets called. As per x86 documentation, the next 4 bytes are a relative offset to the function starting from the instruction after the CALL instruction.

IDA loves to give me these E8 sigs, and before now I'd just get frustrated and manually punch in a sequence of bytes long enough to work as a unique identifier. This should allow for E8 sigs to not utterly fail when used.

PR #54, opened 2020-03-14
Add GetStaticAddressFromSig

attickdoor:
Adds `GetStaticAddressFromSig`, a function whose purpose is to automatically grab addresses that would need to be manually updated every patch. Since all static data must be called from a function, the goal of this function is to jump to the instruction that references the data, and scan through the instruction until a suitable address is found. All addresses found should exist in .data space.


attickdoor:
Any IDA sig should work with this function. Just find any xref for the address you want and grab a signature at that instruction. Put the sig as normal, and provide the offset as a regular number.

goaaats:
Hey, thanks a lot for this! Should improve update delays a lot.
Let's merge this for the 5.21 hotfix if there is a client patch and see how it's gonna work out!

PR #62, opened 2020-03-22
don't even attempt to load assemblies for disabled or invalid plugins…

ff-meli:
…; fixes issues where starting the game with multiple versions, even disabled, could load the wrong one and block other versions.

### DIFF
--- a/Dalamud/Configuration/PluginConfigurations.cs
+++ b/Dalamud/Configuration/PluginConfigurations.cs
@@ -16,6 +16,10 @@ namespace Dalamud.Configuration
             this.configDirectory.Create();
         }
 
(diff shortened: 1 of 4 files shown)


ff-meli:
This should be enough to "fix" all the current breaking issues with plugin updating in game.
- Fixes allowing loading assemblies with the same name
- Fixes disabling the old plugin during the update process, due to clashes with the .disabled file
- Fixes updated plugins not being able to load their config, due to newtonsoft loading the config for the old plugin assembly
- Fixes/optimizes some aspects of plugin loading to not load assemblies that are marked as disabled, or that are invalid (no plugin def)

I imagine we'll want to change some of this later, but this at least gets things to behave without changing the plugin API, and should fix some user issues.

Issue #52, opened 2020-03-10
Port Dalamud to .NET Core


Minoost:
TBD

daemitus:
Hey, I did a thing!

Now in public testing.

PR #67, opened 2020-03-29
[WIP] Port Dalamud to .NET Core

Minoost:
Related issue: #52 

Current goal is to make EntryPoint.Run() work.

## Future works
- We need to rework on plugin design. (Preferably using AssemblyLoadContext)
 https://docs.microsoft.com/en-us/dotnet/core/dependency-loading/understanding-assemblyloadcontext


goaaats:
I'll close this for now since we are working on it in another setting, thank you!

PR #69, opened 2020-03-31
chore: set log level of actor index message to verbose

karashiiro:
Reducing spam, one single-line PR at a time 😡


goaaats:
My bad, did that for debugging of the actor crap

PR #71, opened 2020-04-02
Add UIGlow and UIForeground SeString payloads; refactor some handling of integer markers in payloads, to make encoding a bit simpler

ff-meli:
Very exciting.


ff-meli:
Added map links and some additional reorganization and fixup.

Was hoping the original PR would have been merged to avoid cluttering this one with multiple changesets, but this new work depends on the previous changes so it has to be here too.

ff-meli:
Also added auto translate/completion payloads.

Next up is probably some refactor to make creating or modifying these better, and figuring out a decent way to get lumina injected into them for resolving data.  It might be nicer to have this merged in if there are no issues, to avoid having to keep mass (re)merging master in every time this gets more work done.

PR #76, opened 2020-04-09
Add GetIcon helper

fmauNeko:
Add some helpers to get any icon with Lumina, and convert it so it can be passed to LoadImageRaw from PR #73.


Dalamud/Data/LuminaExtensions/LanguageExtensions.cs
```diff
@@ -0,0 +1,33 @@
+using System.Collections.Generic;
+using Lumina.Data;
+
+namespace Dalamud.Data.LuminaExtensions
+{
+    /// <summary>
+    /// Extensions to <see cref="Language"/>.
+    /// </summary>
+    public static class LanguageExtensions
```
goaaats, line 9: Can you make this extend ClientLanguage instead of Lumina's Language? The language of the icon, if they do have languages, should be automatically chosen based on StartInfo.Language, like Excel.

fmauNeko: That should do the trick !

PR #82, opened 2020-04-21
Make some payload fields public

rreminy:
Some changes making some payload fields (specifically raw ones) setters public, and also allowing for the RawPayload to be instantiated directly.

### DIFF
--- a/Dalamud/Game/Chat/SeStringHandling/Payloads/RawPayload.cs
+++ b/Dalamud/Game/Chat/SeStringHandling/Payloads/RawPayload.cs
@@ -8,10 +8,10 @@ namespace Dalamud.Game.Chat.SeStringHandling.Payloads
     {
         public override PayloadType Type => PayloadType.Unknown;
 
-        public byte ChunkType { get; private set; }
-        public byte[] Data { get; private set; }
+        public byte ChunkType { get; set; }
+        public byte[] Data { get; set; }
 
-        public RawPayload(byte chunkType)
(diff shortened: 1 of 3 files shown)


ff-meli:
(That said, these are the relatively minor payloads so it's less of an impact - RawPayload especially probably won't get much of a change)

rreminy:
The reason I'm proposing this change is because I need to reproduce flags to post into the chat window, and as part of the flag I need to be able to set the Foreground and Glow colors. Well not really but I need to make it look as genuinely as possible it as if it came from the game itself. Its for the symbol that goes right before the flag text 

image

ff-meli:
I do agree that the functionality needs to be exposed in some fashion.  I mostly just don't want to do things like this right now, because all of this is in the process of being refactored, probably with significant internal changes.  None of this is really production-ready at the moment.

I don't mind merging these specific changes too much since these are minor payloads, but I'm also not going to guarantee that those properties will exist as-is in later versions, so it's kind of dependent on if you want to have to potentially have something break.

(Once the lower-level refactor is farther along, I'd also like to add a higher-level string class that can handle a lot of things like ui glow and those arrows for you, but that obviously won't exist until the preliminary work is done)

ff-meli:
Addressed differently in #100 

In general, properties are modifiable only where we might expect a need to modify a payload in flight - so things like changing the displayed text of a text payload before passing it further along to the game.

Other custom values should now be provided via (mostly new) constructors.

PR #84, opened 2020-04-21
Make RawX and RawY ints in MapLinkPayload

rreminy:
I found out RawX and RawY because:
- The raw coordinates (0, 0) is roughly the middle flag coordinates (21.5, 21.5)
- Going a bit to the left (or up) they jump to 4.2 billion (which makes it easy to tell its just the sign bit being set)


ff-meli:
@rreminy Using ints for these breaks on some values, which is why it was intentionally done as uints (as the casts should make obvious).

Is there a case where this does not convert to correct values for you?

PR #92, opened 2020-04-22
Implement IReadOnlyCollection<Actor> on the ActorTable to allow foreach without casting

karashiiro:
This should make the ActorTable API a bit more intuitive for consumers.


goaaats:
We should do this for attick's party list PR as well. Thank you!

PR #100, opened 2020-04-26
DRAFT for comments - Sestring payloads refactor

ff-meli:
Just looking for feedback on things that might be missing/not working/etc.

This is really just the base level of the payload refactor, to try to fix the low-level api to a point that it won't likely need breaking changes later.  I'm also working on some ways to improve string handling with formatting info (ui colors, italics, etc), but none of that is in this PR and shouldn't cause breaking changes in the lower-level types.

The only somewhat higher-level utilities are in SeStringUtils, and that is very limited for now.

This also does not attempt to clean up internal integer handling yet; that can come later and is entirely internal so it won't affect users.

@goaaats @rreminy @pmgr @Haplo064

I'm not sure who else might use payloads currently; tag anyone I've missed.


Dalamud/Game/Chat/SeStringHandling/Payloads/AutoTranslatePayload.cs
```diff
@@ -1,105 +1,156 @@
...
+        /// <remarks>
+        /// Value is evaluated lazily and cached.
+        /// </remarks>
+        public string Text
+        {
+            get
+            {
+                // wrap the text in the colored brackets that is uses in-game, since those
+                // are not actually part of any of the payloads
+                this.text ??= $"{(char)SeIconChar.AutoTranslateOpen} {Resolve()} {(char)SeIconChar.AutoTranslateClose}";
```
Haplo064, line 31: Maybe don't wrap with the brackets by default?
The game probably does some handing itself of these payloads and wraps them after, similar to how it changes the text in emote text.

ff-meli: The game just gets the binary payload, there is no text in it at all (or in any payloads that show up in chat when one of these is sent)

ff-meli: The basic idea here is that .Text should be the actual text you see in the game, that you might expect to be in the payload (it just happens that for these payloads, they are entirely binary lookups)

goaaats:
I think some important things are:

- For all payloads, we should have an implemented text rendering so anyone that calls TextValue will get a string as close as possible to what's seen in the game
- We should probably write tests for this when the implementation is nailed down

ff-meli:
>     * For all payloads, we should have an implemented text rendering so anyone that calls TextValue will get a string as close as possible to what's seen in the game
All payloads that don't implement ITextProvider do not directly display any text, even when they contain it.  It is always a following TextPayload that is actually what the user sees (items, player links, status, etc)

>     * We should probably write tests for this when the implementation is nailed down

Yeah, that will be fun >.>

ff-meli:
Note that this ended up picking up another unmerged PR, so just ignore the actortable changes in this; can always pull them out whenever this is set to be merged if they still aren't in.

Aida-Enna:
There are a lot of actortable changes in this. 🤔

ff-meli:
In lieu of formal tests, for which I can't think of way that would actually give any meaningful results, I'll probably see about porting most or all of the internal dalamud string handling to use payloads instead of the current mixed handling.

That should hopefully show what things don't work, as well as what are still overly painful to use etc.

Dalamud/Dalamud.cs
```diff
@@ -116,14 +117,16 @@ public sealed class Dalamud : IDisposable {
                 this.Data = new DataManager(this.StartInfo.Language);
                 await this.Data.Initialize(this.baseDirectory);
 
+				// TODO: better way to do this?  basically for lumina injection
+            	SeString.Dalamud = this;
```
goaaats, line 121: I think just giving DataMgr to each payload is a bit nicer, will do that after merge

PR #105, opened 2020-04-30
Update ActorTable offsets for hp/mp, add ui (display-only.. probably) status effects

ff-meli:
I didn't expose the status effects on Chara/etc since I wasn't sure if we cared yet, and I wasn't sure if you wanted raw struct accessors there (since the current way is kind of a wrapper)


ff-meli:
Is there something wrong with this?  It's mostly just fixing values that are currently broken.

goaaats:
Sorry, took a little break and forgot about this one

PR #109, opened 2020-05-07
Update TargetActorId in ActorTable

Caraxi:


attickdoor:
The new field seems to be the ActorID when a PlayerChara is the one doing the targeting. The existing field seems to be the current target of a bNPC. There may be more edge cases that these two fields combine don't catch. I'd suggest adding a `GetTarget` method that chooses the correct one based on the ActorKind.
Also, the new field doesn't get filled with an ActorID when targeting oneself. Probably a quirk of the client, as I believe targeting yourself is purely cosmetic.

goaaats:
Any update on this? What do we want to do?

KHShadowrunner:
For what it's worth this appears to also be the ActorID for bNPC's that is engaging a PlayaChara. It does need to be modified for specific types of Actor Types, but that's the case for most of the ActorTable field (Like Debuffs, etc)

Caraxi:
Still trying to determine how and when each of the fields is actually used. Both act strange at seemingly random times.

Caraxi:
I'm sure there is still issues with this... Its not the most stable offset

Caraxi:
#119 seems better and I didn't mess up the branch

PR #111, opened 2020-05-10
temporary party member resolution optimization

attickdoor:
The first 244 entries are where PCs and their minions/mounts/etc go. PCs are on even indices, minion etc is on odd indices. This is to reduce the amount of hanging when trying to resolve a party member who is not in your current zone, and thus not in the actor table.


attickdoor:
Ideally, there is a better way to resolve party members altogether. This is just to make the edge case less annoying in the meantime. In general, resolution is sufficiently fast if the actor is actually in the table.

attickdoor:
On second thought, I'd rather just leave this as is. It might break other edge cases, so I'll just try and get a better solution in instead.

PR #115, opened 2020-05-26
Update lumina, enables an option which will stop some sheets failing

NotAdam:
Updates Lumina to 1.1.4
* 'fixes' the issue where sheets would fail to parse via `LuminaOptions.ExcelSheetStrictCastingEnabled`. Defaults to on, which I assume is generally more helpful than having it throwing exceptions
* Updated excel definitions


fmauNeko:
Does that mean we can get rid of the sheets in `Dalamud.Data.TransientSheet` ?

goaaats:
People depend on them for now, but I'll notify plugin devs to switch over to the built-in ones.

PR #118, opened 2020-05-27
fix : Add non-Japanese characters that have been erased.

marimelon:
GetGlyphRangesJapanese now includes everything that was included in it.
I think we should have a separate scope definition for each language, but I can't think of a better way.
[https://github.com/ocornut/imgui/blob/master/imgui_draw.cpp#L2487-L2493](url)


goaaats:
Since this is only a problem for Japanese and we're not planning to support Chinese or Korean right now, I think we're good with this at the moment.

PR #134, opened 2020-07-02
master -> plugin_rewrite

goaaats:


goaaats:
This sucks, better to do it from scratch

PR #139, opened 2020-07-19
ActorTable performance improvements

rreminy:
**Changes summary**:
- Adds `IDisposable` pattern to free `actorMem` (which I forgot in my previous PR) and added the appropiate call in `ClientState`.
- Added `GetPointerTable()` which retrieves all actor pointers in a single Marshal call.
- Added `GetActorTable()` which retrieves all actors from the pointer table.
- Separated away the Actor memory reading function into its own function (`ReadActorFromMemory`).
- Added a cache to speed up reading the Actors Table when read multiple times. This cache needs to be reset every frame, adjustment changes have been made to do so before emitting the `Framework.OnUpdateEvent`.
- Updated all functions to reflect the usage of the new Actors Table cache.

**Performance regressions**:
- Fetching `ClientState.LocalPlayer` will cause the entire Actor Table to be constructed. (A workaround would be to fetch it the original way as a special case)

**Performance Metrics (approximates)**:
- Without this PR: `450ms`
- With this PR (no cache): `110ms`
- With this PR (with caching): `2ms`

Performance Metrics taken at my Mansion with the following code (inside the Data window)
```csharp
System.Diagnostics.Stopwatch sw = new Stopwatch();
this.dalamud.ClientState.Actors.ResetCache();
sw.Start();
for (int i = 0; i < 256; i++)
    var actors = this.dalamud.ClientState.Actors.ToList();
    //this.dalamud.ClientState.Actors.ResetCache();
}
sw.Stop();
stateString += $"Time taken: {sw.Elapsed.TotalMilliseconds}ms\n\n";
```
**Additional remarks**:
- Care is taken to avoid changing the visible API, which means nulls are preserved in `this[i]` and nulls are skipped in `GetEnumerator`.
- The `ReadProcessMemory` code if removed can recduce the no cache metric by half. I did some testing without it and had no crash but decided to preserve it, is this really needed?

### DIFF
--- a/Dalamud/Game/Internal/Framework.cs
+++ b/Dalamud/Game/Internal/Framework.cs
@@ -23,6 +23,7 @@ namespace Dalamud.Game.Internal {
         public event OnUpdateDelegate OnUpdateEvent;
         
         private Hook<OnUpdateDetour> updateHook;
+        private Dalamud dalamud;
         
         
         /// <summary>
@@ -49,6 +50,8 @@ namespace Dalamud.Game.Internal {
         #endregion
         
         public Framework(SigScanner scanner, Dalamud dalamud) {
+            this.dalamud = dalamud;
+
             Address = new FrameworkAddressResolver();
             Address.Setup(scanner);
             
@@ -101,6 +104,7 @@ namespace Dalamud.Game.Internal {
 
         private bool HandleFrameworkUpdate(IntPtr framework) {
             try {
+                dalamud.ClientState.Actors.ResetCache();
                 Gui.Chat.UpdateQueue(this);
                 Network.UpdateQueue(this);
             } catch (Exception ex) {
--- a/Dalamud/Game/Internal/Framework.cs
+++ b/Dalamud/Game/Internal/Framework.cs
@@ -104,7 +104,6 @@ namespace Dalamud.Game.Internal {
 
         private bool HandleFrameworkUpdate(IntPtr framework) {
             try {
-                dalamud.ClientState.Actors.ResetCache();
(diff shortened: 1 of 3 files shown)


Dalamud/Game/Internal/Framework.cs
```diff
@@ -101,6 +104,7 @@ public sealed class Framework : IDisposable {
 
         private bool HandleFrameworkUpdate(IntPtr framework) {
             try {
+                dalamud.ClientState.Actors.ResetCache();
```
goaaats, line 107: I think we should subscribe to this from ActorTable, instead of doing it in the hook itself

goaaats:
As far as I can remember, we left in the ReadProcessMemory as a safeguard - however, I haven't noticed any problems with it either. Will do some testing and report back!

goaaats:
Should be fine to remove, haven't seen any further issues.

Dalamud/Game/Internal/Framework.cs
rreminy: The reason I did it there is because its done all the way before the event handler is invoked. That way I can guarantee the cache is reset before any calls to the actor table functions happen. I'll see if it works fine from within and commit the change if it is.

rreminy: Surprisingly it didn't cause any side effects. Worse case would be that the actor table would be a frame behind, but I didn't see it manifesting and it wouldn't be a major issue.

rreminy:
New performance metrics with the removal of `ReadProcessMemory`:
- Uncached: `55ms`
- Cached: `~1-2ms`

PR #144, opened 2020-07-24
Massively improved discord chat bridge

witcheslive:
I have fixed up the Discord chat bridge functionality quite a bit to be MUCH nicer!

Note that while many if not most of these improvements work in the embed mode, turning it off is a much nicer experience and will enable the "Chat Prefix" functionality. I recommend changing the default to disabling embeds, but I will leave this consideration to the upstream developers.

Here is a summary of what this and its companion patch on XIVLauncher does.

- Chat formatting overall has been updated to more closely resemble the game
- - Bolding is only for chatmode tag
- - Chatmode tag is in all caps, just like in the game
- - cw1, cw2 etc have been changed to cwls1, cwls2 etc to match the in game chat
- - Free Company messages no longer include world next to the character name
- - World identifier changed from " at " to "@" for easier copy and paste
- Option on each chatmode to prepend a string for the non-embedded version, intended mainly for using Discord emoji representing things such as a linkshell or free company crest to differentiate different chats
- Every single character in the XIV special characters codepage has been given a unicode equivalent, or at least, as close as possible. No more ugly boxes any time someone links an item or uses autotranslate! 
- As the autotranslate arrows and hq icons are simultaneously iconic and options to represent them in unicode is poor, there are now additional settings in the "In-Game" configuration to use a Discord emoji of your choosing for these three icons


Dalamud/DiscordBot/DiscordBotManager.cs
```diff
@@ -315,12 +348,187 @@ public async Task ProcessCfPreferredRoleChange(string rouletteName, string prevR
...
+                {
+                    unichar = char.ConvertFromUtf32('①' + i - 1);
+                }
+                if( i >= 21 )
+                {
+                    unichar = char.ConvertFromUtf32('㉑' + i - 21);
+                }
+
+                ffxivSpecialCharacters.Add(xivchar.ToString(), unichar);
+            }
```
ff-meli, line 514: Can you pull this out into a one-time method, rather than for every message?  It's basically a const dictionary, there's no reason to do this work every time.

Also for goat - a cleaned up version of this might be useful higher up in dalamud, though now that we have the game symbol and font handling, it may not be used much.

Dalamud/DiscordBot/DiscordBotManager.cs
```diff
@@ -315,12 +348,187 @@ public async Task ProcessCfPreferredRoleChange(string rouletteName, string prevR
...
+            foreach (char c in ffxivString)
+            {
+                if ( ffxivSpecialCharacters.ContainsKey(c.ToString()) )
+                {
+                    returnString += ffxivSpecialCharacters[c.ToString()];
+                } else
+                {
+                    returnString += c;
                 }
             }
```
ff-meli, line 526: Please do this with a string builder.  This is already a pretty inefficient way to modify a string, but there's no reason to not at least use a builder to be nice to the system.

witcheslive: A string builder would be extremely inefficient. At this level of text processing there is literally nothing faster than `O(n)` for a string with `O(1)` for each lookup, which is what this is. I initially did it with a builder and a series of replaces when it was just a few characters in an experiment, but it quickly became something that wouldn't look out of place in Yandere Simulator.

ff-meli: The purpose of the builder is to not have to copy the string potentially many many times, and to avoid as much memory fragmentation.  It is absolutely going to be more efficient than a direct append

Dalamud/DiscordBot/DiscordBotManager.cs
witcheslive: I will update this to not build it every call but this is nowhere close to being called frequently enough to where this will impact literally anything

Dalamud/DiscordBot/DiscordBotManager.cs
witcheslive: This is the part where it looks at the string that is being immediately processed, as in it is preparing a chat message to send to the discord bot, and will not be doing it "many many times". I am working on a change to only build the lookup dictionary once, but for this part we are processing the incoming message for output and there is no more efficient way to do that.

ff-meli: Every _single_ character append to a string has the potential to result in a copy of the entire string.  This is one of the major reasons string builders _exist_ in many languages.

In reality it's going to be affected by alignment and padding chosen by (in this case) the managed runtime, but it's absolutely going to happen.

I literally just mean to replace your string += with a stringbuilder Append(), and don't call ToString() until the return.  It's not a rework, and it is unequivocally more efficient.

witcheslive: what

do you understand the scope of this part of the code

this is literally the part where someone says "i like :hqsymbol: butts" in fc chat, then it looks through the message to find the ffxiv special characters, in this case, swapping out the hq symbol as necessary

ff-meli: Yes I understand exactly what it does.

I'm not sure you understand what buliders are for.  If you are this confused, just let this get merged and I'll fix it.

Poliwrath: There's also the issue with formatting being weird making this PR a bit hard to see actual changes, but that's probably okay and I know some IDEs probably automatically do it

Dalamud/DiscordBot/DiscordBotManager.cs
```diff
@@ -315,12 +348,187 @@ public async Task ProcessCfPreferredRoleChange(string rouletteName, string prevR
                         }
                     }
 
-                    await channels.ElementAt(chatTypeIndex).SendMessageAsync($"**[{chatTypeDetail.Slug}]{name}**: {rawMessage}");
+                    string messageName = name;
+                    if (chatTypeDetail.Slug.Equals("fc"))
```
Poliwrath, line 352: i am hopping aboard the review train: ``Equals("fc")`` -> ``== "fc"``?

Dalamud/DiscordBot/DiscordBotManager.cs
Caraxi: string + string isn't O(1) btw, its O(n^2)

Dalamud/DiscordBot/DiscordBotManager.cs
witcheslive: ...

Dalamud/DiscordBot/DiscordBotManager.cs
witcheslive: oh so c# concatenates strings in a dumb way

alright again this is... a really really stupid thing to nitpick over as none of this is anywhere close to the volume or processing power to where this much energy and time needs to be spent obsessing over it, but nevertheless i have an idea for how to fix it

Caraxi: I don't believe there exists any language that string + string is O(1)

ff-meli: I mean.. every language with string concat works that way.

I think maybe you think I meant changing more than I'm actually suggesting.  The flow there is fine, it's just this:
`var returnString = new StringBuilder();
foreach (char c in ffxivString)
{
    if ( ffxivSpecialCharacters.ContainsKey(c.ToString()) )
    {
        returnString.Append(ffxivSpecialCharacters[c.ToString()]);
    }
    else
    {
        returnString.Append(c);
    }
}

return returnString.ToString();
`

ff-meli: Like you *could* rework things here to be greedy about in-line changes, but since the replacements are variable-sized, that really doesn't seem worth it (or clean).  I mentioned the builder here because this is literally the stereotypical use case they are for, and it's a simple fix.

witcheslive: Updated. It now creates a stringbuilder with a copy of the message, then does the loop to look for things that need replacing, then runs a stringbuilder replace when those items are found.

ff-meli:
For goat - I'm fine with these if you are, but it's not giving me the option to resolve my request for changes etc.  So don't treat that as blocking or anything.

Dalamud/DiscordBot/DiscordBotManager.cs
takhlaq: \uE05A*

NotAdam:
Is there a particular reason you went forward with implementing this in Dalamud, despite being told that this feature is better suited for a plugin? Despite that you were told that a plugin is better suited for this functionality (including the fact that you agreed that a plugin would be more suitable).

Not to mention, you are asking it to be merged faster because you are breaking your config locally, which isn't the responsibility of anyone but yourself - so why should we care?

witcheslive:
:roll_eyes: 

I did not sign up to restructure your project, I wanted to improve existing functionality. If you want it to be a plugin, feel free to submit a patch yourself.

I am tired of having to go through all the steps of rebuilding, figuring out why dalamud didn't get grabbed this time, whoopsie it ran the master version now I have to do all the settings again, etc etc. This is functionality for everyone to enjoy as well. Why are you like this.

Caraxi:
It was mentioned many times that it should be a plugin but you persisted. You were warned that it might not get accepted as a built in feature of dalamud

witcheslive:
OK great then someone who cares about that can do it. I did not add a new module, I took what was there and added improvements to it. None of these changes will make it any trickier to separate into a module for someone who is inclined. I do not know the codebase well enough nor do I have the slightest inclination to volunteer my time to learn it when I'm sure I would receive pedantic nonsense like this even if I did. I wanted the chat to not look broken when bridged to discord, and make it look nicer. That is what I am here to do, that is what I did. Being a jerk isn't going to convince me to take up the backlog of boring stuff you won't just do yourself.

Poliwrath:
There's a lot of things in Dalamud that should eventually be split into separate plugins, adding onto those things makes that job harder for people in the future... even slightly
Also, for your issue of Dalamud messing up configurations, you can disable auto update you know... I've been running 4.9.5.1 for 2 or 3 months with the same set of plugins and it worked just fine (until 5.3 obviously)

witcheslive:
I added configuration options to this, so you can put discord emoji. Every time I run the upstream version without it it removes those options and I have to redo them, which is annoying because I have to go get a bunch of discord emoji codes and I pointedly do not run discord on the machine that runs XIV. It is not corrupting anything, just supremely annoying. You would know this if you actually looked at what this did instead of engaging in enervating meta-arguments trying to get me to do things you are too lazy to do yourself.

Overall, are some very simple changes. If y'all want a module instead of a builtin so badly, please stop crying at me and make them a module. I do not know the codebase, it is not my expertise, I am not responsible for your backlog. With the way I have been received on this I have no inclination to do anything else with XL again, never mind thankless backend work in an unfamiliar backend.

Poliwrath:
😐

witcheslive:
Please merge this so I can move on with things and stop worrying about it, thanks.

goaaats:
Hey, thank you! I'm planning to rewrite Discord integration over the coming weeks and putting it in a plugin, I will include these changes in that rewrite. Hope you can run xivlauncher in ReleaseNoUpdate mode until then.

Sorry for the trouble!

witcheslive:
I would prefer to just have these now as they shouldn't complicate the work of making it a plugin, and it seems like it would be easier to just move it all in one fell swoop then to come back and retrofit these changes to the plugin version, though of course it is ultimately up to you.

goaaats:
Hey, I implemented your changes into the new Discord bridge plugin at https://github.com/goaaats/Dalamud.DiscordBridge. Thanks for your contribution!

PR #145, opened 2020-07-27
Small things (again)

Poliwrath:
* Fixed crash if using Dalamud without XIVLauncher / without the config file generated by XIVLauncher
* Moved FieldOffsets for Actor to it's own class for easier access from plugins instead of using reflection / hardcoding offsets that could change in the future (I have no idea if this happens ^__^)
* Made /xldev toggle the dev menu instead of only enabling it


goaaats:
People do actually hardcode these, so this is cool, thank you!

PR #146, opened 2020-07-28
Target Info

pohky:
add target info to ClientState


Dalamud/Game/ClientState/Targets.cs
```diff
@@ -0,0 +1,28 @@
...
+        
+        internal Targets(Dalamud dalamud, ClientStateAddressResolver addressResolver) {
+            this.dalamud = dalamud;
+            Address = addressResolver;
+        }
+
+        private Actor GetActorByOffset(int offset) {
+            if (Address.TargetManager == IntPtr.Zero) return null;
+            var actorAddress = Marshal.ReadIntPtr(Address.TargetManager + offset);
+            if (actorAddress == IntPtr.Zero) return null;
```
goaaats, line 23: We should cast this to their respective subclass, there's a method for it in ActorTable

PR #148, opened 2020-08-03
better config handling & cleanup warning noise

NotAdam:


NotAdam:
[10:07] Adam: ah actually after conversation with meli yesterday i think we came up with a better idea
[10:07] Adam: the idea being plugins get a folder instead and can manage their own data in it
[10:07] Adam: so if there's any data files or anything outside of config, that can be stored too
[10:07] goat: i'd keep the option to just save an object
[10:08] goat: but that sounds great
[10:08] goat: good for ack too
[10:08] goat: so he can store his chrome install
[10:08] Adam: yeah naturally, just let you give it any type instead of IPluginConfiguration
[10:08] Adam: cause that's what I changed it to do and it seemed to work fine but was more of a hack

Issue #151, opened 2020-08-08
Size of World


Bluefissure:
I think the size of currentWorld and homeWorld should be `ushort` instead of `byte` at [Actor.cs#L71](https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Game/ClientState/Structs/Actor.cs#L71).
You can see [World.csv](https://github.com/xivapi/ffxiv-datamining/tree/master/csv/World.csv) for ref.

Caraxi:
It can't be because theres no room for a short there

Bluefissure:
> It can't be because theres no room for a short there

`CurrentWorld = 0x1884` and `HomeWorld = 0x1886`, why theres no room for a 2-byte `ushort`?

Caraxi:
Yep I'm an idiot and didn't see offsets on the line so just assumed it was sequential and didn't notice it was definined with a constant lol

goaaats:
Thanks for fixing this!

PR #152, opened 2020-08-08
change world size to ushort

Bluefissure:
As talked in https://github.com/goatcorp/Dalamud/issues/151.


Bluefissure:
Uh I need to change it in vs instead of manually edit the source :(

PR #156, opened 2020-08-11
Some 5.3 Sigs

Caraxi:
Hopefully they're right lol

Still missing ConditionFlags and LocalContentId


Caraxi:
Added sigs for LocalContentId and ConditionFlag thanks to Mino finding the offsets

PR #158, opened 2020-08-17
Actor visibility helpers

Poliwrath:
Pretty much complete, but just needs reviews.
This is my first time ever doing stuff with unsafe, ~~so I'm 99% sure it's wrong...~~ (but it works!) (Thanks Adam!)

Should ``ObjectVisibility`` be an enum (or even named ``ObjectVisibility``? I don't know if there's any other values, CMTool and VoidList only have 0x00/0x02 and also call it "render".

Should the helper functions (Hide/ReRender) be in their own class (ActorHelpers?)? Should they even be in Dalamud at all?

https://github.com/imchillin/CM3/blob/e7664d60e33ae0afe8e3f4ad3d81c4ba5cd44d1a/ConceptMatrix3/Services/ActorRefreshService.cs#L92
https://github.com/Haplo064/VoidList/blob/master/VoidList.cs#L220

### DIFF
--- a/Dalamud/Game/ClientState/Actors/Types/Actor.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Actor.cs
@@ -71,6 +71,24 @@ namespace Dalamud.Game.ClientState.Actors.Types {
         /// </summary>
         public virtual int TargetActorID => 0;
 
+        public ObjectVisibility ObjectVisibility => this.actorStruct.ObjectVisibility;
+
         bool IEquatable<Actor>.Equals(Actor other) => this.ActorId == other.ActorId;
+
+        /// <summary>
+        /// Hides the actor
+        /// </summary>
+        public unsafe void Hide()
+        {
+
+        }
+
+        /// <summary>
+        /// Re-renders the actor
+        /// </summary>
+        public unsafe void ReRender()
+        {
+
+        }
     }
 }
--- a/Dalamud/Game/ClientState/Actors/Types/Actor.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Actor.cs
@@ -1,5 +1,6 @@
 using Dalamud.Game.ClientState.Structs;
 using System;
+using System.Threading.Tasks;
 
 namespace Dalamud.Game.ClientState.Actors.Types {
     /// <summary>
--- a/Dalamud/Game/ClientState/Actors/Types/Actor.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Actor.cs
@@ -74,7 +74,10 @@ namespace Dalamud.Game.ClientState.Actors.Types {
         /// </summary>
         public virtual int TargetActorID => 0;
 
-        public ObjectVisibility ObjectVisibility => this.actorStruct.ObjectVisibility;
+        /// <summary>
+        /// Actor render mode
+        /// </summary>
+        public RenderMode RenderMode => this.actorStruct.RenderMode;
 
         bool IEquatable<Actor>.Equals(Actor other) => this.ActorId == other.ActorId;
 
@@ -87,14 +90,14 @@ namespace Dalamud.Game.ClientState.Actors.Types {
             {
                 var player = ObjectKind == ObjectKind.Player;
                 var kind = (ObjectKind*)(Address + ActorOffsets.ObjectKind);
-                var vis = (ObjectVisibility*)(Address + ActorOffsets.ObjectVisibility);
+                var render = (RenderMode*)(Address + ActorOffsets.RenderMode);
 
                 if (player)
                 {
                     *kind = ObjectKind.BattleNpc;
                 }
 
-                *vis = ObjectVisibility.Invisible;
+                *render = RenderMode.Invisible;
 
                 if (player)
                 {
@@ -105,7 +108,7 @@ namespace Dalamud.Game.ClientState.Actors.Types {
         }
 
         /// <summary>
-        /// Re-renders the actor
+        /// "Re-Renders" the actor
         /// </summary>
         public unsafe void ReRender()
         {
@@ -113,17 +116,16 @@ namespace Dalamud.Game.ClientState.Actors.Types {
             {
                 var player = ObjectKind == ObjectKind.Player;
                 var kind = (ObjectKind*)(Address + ActorOffsets.ObjectKind);
-                var vis = (ObjectVisibility*)(Address + ActorOffsets.ObjectVisibility);
+                var render = (RenderMode*)(Address + ActorOffsets.RenderMode);
 
                 if (player)
                 {
                     *kind = ObjectKind.BattleNpc;
                 }
 
-                *vis = ObjectVisibility.Invisible;
-
+                *render = RenderMode.Invisible;
                 Thread.Sleep(100);
-                *vis = ObjectVisibility.Visible;
+                *render = RenderMode.None;
 
                 if (player)
                 {
--- a/Dalamud/Game/ClientState/Actors/Types/Actor.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Actor.cs
@@ -86,16 +86,14 @@ namespace Dalamud.Game.ClientState.Actors.Types {
 
             unsafe
             {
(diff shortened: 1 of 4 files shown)


Dalamud/Game/ClientState/Actors/Types/Actor.cs
```diff
@@ -71,6 +73,86 @@ public class Actor : IEquatable<Actor> {
...
+        /// <summary>
+        /// Hides the actor
+        /// </summary>
+        public async void Hide()
+        {
+            var player = ObjectKind == ObjectKind.Player;
+
+            unsafe
+            {
+                ObjectKind* kind = (ObjectKind*)(Address + ActorOffsets.ObjectKind);
```
NotAdam, line 89: use `var`, there's no point in specifying the type on the left hand side when it's given on the right

Dalamud/Game/ClientState/Actors/Types/Actor.cs
```diff
@@ -71,6 +73,86 @@ public class Actor : IEquatable<Actor> {
...
+                ObjectVisibility* vis = (ObjectVisibility*)(Address + ActorOffsets.ObjectVisibility);
+                {
```
NotAdam, line 91: why extra scope?

Dalamud/Game/ClientState/Actors/Types/Actor.cs
```diff
@@ -71,6 +73,86 @@ public class Actor : IEquatable<Actor> {
...
+        /// <summary>
+        /// Re-renders the actor
+        /// </summary>
+        public async void ReRender()
+        {
+            var player = ObjectKind == ObjectKind.Player;
+
+            unsafe
+            {
+                ObjectKind* kind = (ObjectKind*)(Address + ActorOffsets.ObjectKind);
```
NotAdam, line 123: you should just mark the function as unsafe and do this once

Dalamud/Game/ClientState/Actors/Types/Actor.cs
```diff
@@ -71,6 +73,86 @@ public class Actor : IEquatable<Actor> {
...
+                ObjectVisibility* vis = (ObjectVisibility*)(Address + ActorOffsets.ObjectVisibility);
+                {
+                    if (player)
+                    {
+                        *kind = ObjectKind.BattleNpc;
```
NotAdam, line 94: what you should do instead above is like 

    var kind = *(ObjectKind*)(Address + ActorOffsets.ObjectKind);

then you can just

    kind = ObjectKind.BattleNpc;

Dalamud/Game/ClientState/Actors/Types/Actor.cs
NotAdam: and again mark func as unsafe and do this all once

Dalamud/Game/ClientState/Actors/Types/Actor.cs
```diff
@@ -71,6 +73,86 @@ public class Actor : IEquatable<Actor> {
...
+                    }
+
+                    *vis = ObjectVisibility.Invisible;
+                }
+            }
+
+            if (player)
+            {
+                await Task.Delay(100);
```
NotAdam, line 103: 😢

Dalamud/Game/ClientState/Actors/Types/Actor.cs
NotAdam: or just like `var actor = *(Actor*)(Address)` and then `actor.ObjectKind = thing` and etc.

Dalamud/Game/ClientState/Actors/Types/Actor.cs
Poliwrath: can't do unsafe func + ``await Task.Delay(100)``

NotAdam: lol async is such garbage, a Task.Run then with system.threading.sleep would probably make it much nicer then

Dalamud/Game/ClientState/Actors/Types/Actor.cs
Poliwrath: leftover from when I was messing with ``fixed``, fixed

Dalamud/Game/ClientState/Actors/Types/Actor.cs
Caraxi: Thread sleep stops all tasks from running through, so its not ideal

NotAdam: shouldn't do because tasks are pooled?

Caraxi: I'm likely to be misinterpreting something, but there was issues with it in a plugin previously I think
https://github.com/attickdoor/XIVComboPlugin/commit/75b9c8e4

Poliwrath: I just tried the Thread.Sleep inside of Task.Run it and it doesn't seem to lock up my game

AmenneHolelane: test with sonar enabled as well, that was the plugin causing issues

Caraxi: If its working fine its fine, Adam usually knows what hes saying so probably safe to go with his suggestions lol

NotAdam: a regular thread would work I guess, which wouldn't cause any issues of the sort but this is probably one of those things where it'd be better to use a scheduler instead

no idea why sonar would hang the game if a task is blocking though, must be doing some real stupid shit for that to happen given that tasks wouldn't be executing on a game thread, unless they're awaiting on a game thread which is lol tier

Dalamud/Game/ClientState/Actors/Types/Actor.cs
Poliwrath: doing the first suggestion doesn't work at all, and the second one leads to this compile error (``var actor = *(Structs.Actor*)(Address);``):
```
Cannot take the address of, get the size of, or declare a pointer to a managed type ('Actor')
```

Dalamud/Game/ClientState/Actors/Types/Actor.cs
Poliwrath: yea works fine with Task.Run, I haven't tested it Sonar yet since I don't really want to run that plugin on my main computer (and it's possible that their plugin is doing something wrong, forcing everything to adjust to that isn't really a good thing I think...)

Dalamud/Game/ClientState/Actors/Types/Actor.cs
NotAdam: ah yeah, the first one wouldn't work, I misread what was there entirely

the second should have actually been something like the following:

    var actor = (Actor*)address;
    actor->ObjectKind = ...

Poliwrath: Same error, it's on this line, btw:
```
var actor = (Structs.Actor*) Address;
```

NotAdam: i give up, its probably something to do with the explicit layout or something stupid, just do it the way you have it lol

Poliwrath: 🥳

NotAdam (dismissed):

Dalamud/Game/ClientState/Actors/Types/Actor.cs
NotAdam: yeah if it's doing something stupid like awaiting tasks on a game thread, then that's sonars problem and nobody elses because that absolutely will block and cause the game to hitch

Poliwrath:
I think it's finished now? I took the RenderMode name from Nhaama, I wanted to try to figure out the rest of the options for it but I have no clue what I'm doing (a lot of them just hid my weapon or gave my character a nametag for half a second)

daemitus:
@goaaats do you want me to refresh this? being able to hide the localplayer without the additional checks visibility does seems problematic.

PR #160, opened 2020-08-21
Search on PluginInstaller window

Caraxi:


Dalamud/Plugin/PluginDefinition.cs
```diff
@@ -43,6 +43,11 @@ public class PluginDefinition
         /// </summary>
         public string RepoUrl { get; set; }
 
+        /// <summary>
+        /// List of tanks defined on the plugin.
```
Poliwrath, line 47: tanks?

Caraxi: I umm.. do not know

PR #161, opened 2020-08-25
Fix Status Effect Array to Pull from Actors Struct

r1ft4469:
Added Status Effect Array to Actor Struct.

### DIFF
--- a/Dalamud/Game/ClientState/Actors/Types/Actor.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Actor.cs
@@ -1,3 +1,4 @@
+using Dalamud.Game.ClientState.Structs;
 using System;
 
 namespace Dalamud.Game.ClientState.Actors.Types {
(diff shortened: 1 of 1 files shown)


Dalamud/Game/ClientState/Actors/Types/Actor.cs
```diff
@@ -71,6 +72,11 @@ public class Actor : IEquatable<Actor> {
         /// </summary>
         public virtual int TargetActorID => 0;
 
+        /// <summary>
+        ///  Status Effects
+        /// </summary>
+        public StatusEffect[] statusEffects => this.actorStruct.UIStatusEffects;
```
goaaats, line 78: Could you fix the style on this? Should be StatusEffects. Thanks!

goaaats (changes requested):

Dalamud/Game/ClientState/Actors/Types/Actor.cs
r1ft4469: Taken care of

goaaats (approved):

PR #163, opened 2020-08-29
Un-stub PartyList

attickdoor:
It _should_ work fine, but probably needs to be tested just to be sure. Mostly lifted straight from MOAction (thanks Winter!). Could use some better strategy to not rely on magic struct offsets in `SetupPlaceholderResolver`, but it should at least survive until 5.4, which is plenty of time.


NotAdam (approved):

goaaats:
How does this handle party members that aren't in your zone? Just returning null?

attickdoor:
> How does this handle party members that aren't in your zone? Just returning null?
Correct. `GetActorFromPlaceholder` returns `IntPtr.Zero` if it can't resolve, so it returns null in those cases. Trying to call for a party member that isn't present shouldn't fail catastrophically internally.

aers:
some notes

"Address.PlaceholderResolverObject" is just a pointer to FrameworkBase +0x10 so you don't need to resolve it
that +0x8 ptr is a pointer to UIModule+0x97738 which is the event interface for configuration changes in the config UI, then + E6D0 in that is another object inside the UIModule

probably the correct way to get there:

FrameworkBase + 0x29F8 UIModule ptr
UIModule+0xA5F08 is the "PlaceholderResolverObject" which is the class PronounModule (which contains a macro text checker etc). PronounModule name is based on 2014 client so it could have changed but shrug.

attickdoor:
Neat.

oorzkws:
Is this a candidate for unshittify dalamud? I recently spent a good amount of time trying bashing my head against something not realizing the party list in client state is nonfunctional.

goaaats:
I assume this is superceded by #273

PR #165, opened 2020-09-02
Add cutscenes and gpose to UI hide conditions

karashiiro:
Easy change, figured I'd just do it.

Unrelated, but https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Plugin/PluginInstallerWindow.cs#L61 seems to be calling a bad signature, not sure what that's supposed to be.


karashiiro:
Or not, it builds.

goaaats (approved):

PR #168, opened 2020-09-08
[WIP] Use hooks instead of opcodes for network

lojewalo:
It does what it says on the tin. If you want this sort of thing, I'll address issues below. If you'd rather stick to using opcodes, just close the PR.

- I'm not sure where these hooks should go, really. I kept them in the same file as the network handlers using opcodes, but they can be moved.
- The market board history sig does not use wildcards and is more likely to break. There is no unique signature using wildcards that I can see.
- If there's a pattern for hooking that is more appropriate, please let me know.

### DIFF
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -220,6 +220,8 @@ namespace Dalamud {
             this.SigScanner.Dispose();
             
             this.Data.Dispose();
+
+            this.NetworkHandlers.Dispose();
         }
 
 #region Interface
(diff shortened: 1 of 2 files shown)


goaaats:
Thanks, I'll think about this - initially, we decided against this, because the packet handlers do have especially weak sigs. Let's see how they fare next patch.

ascclemens:
I generated all these sigs except for the history one on the previous patch and they work on this one, at least. We'll have to see about the next one.

ascclemens:
Minus the function with no unique signature, all of these signatures still work on 5.35-HF1.

PR #171, opened 2020-09-12
Switch /xlplugins command to toggle

ms2mml:
This lets you toggle the xlplugins window with a keybind by having an in-game macro with `/xlplugins t`.


goaaats:
I think just having it toggle as the default behavior would be fine, saves on code - right?

ms2mml:
> I think just having it toggle as the default behavior would be fine, saves on code - right?

Done!

goaaats (approved):

PR #178, opened 2020-10-02
Actor MaxMp, CurrentCp, MaxCp, CurrentGp, MaxGp

daemitus:
To get things to build I had to...
- bump ImGuiScene to the top of the external dx_11_testing branch as the git submodule link would not checkout properly and does not build on the master branch. 
- Remove a ref keyword in PluginInstallerWindow as ImGui.InputTextWithHint does not allow it.

Based on the offsets in https://github.com/FFXIVAPP/sharlayan-resources/blob/master/structures/5.3/x64.json, I've added CurrentGp, CurrentCp, MaxGp, MaxCp to the Dalamud.Game.ClientState.Structs.Actor struct and public accessors to the Dalamud.Game.ClientState.Actors.Types.Chara class. 

CurrentGp gets some odd values while as a battle class, but its not used so  ¯\\_(ツ)_/¯

MaxMp has several comments as to unreliableness, however after changing it to a short and fixing the offset from what it was currently (CurrentGp) I'm getting 10000 as expected while smacking around the target dummy. There's certainly no risk to just leaving it hardcoded though, let me know. 

Otherwise, I left the rest of the offsets from the Sharlayan file as comments for reference and let the auto-formatter do some work.

### DIFF
--- a/Dalamud/Game/ClientState/Actors/Types/Chara.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Chara.cs
@@ -42,7 +42,27 @@ namespace Dalamud.Game.ClientState.Actors.Types {
         /// <summary>
         ///     The maximum MP of this Chara.
         /// </summary>
-        public int MaxMp => 10000;  // Currently hardcoded because the value in actorStruct is very questionable.
+        public int MaxMp => this.actorStruct.MaxMp;
+
+        /// <summary>
+        ///     The current GP of this Chara.
+        /// </summary>
+        public int CurrentGp => this.actorStruct.CurrentGp;
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


goaaats:
That ref bug is fixed by using the correct commit of ImGuiScene.
The correct commit is https://github.com/goatcorp/ImGuiScene/tree/97a487c795b7c8fda6ce5342c88035ba9f40e53d, but it was only on my personal fork, not on the org one - so your checkout failed, but mine succeeded. Sorry about that.

Could you try again and revert the two commits of yours that mess with this?

daemitus:
As requested, commits have been reverted.

goaaats (approved):

PR #181, opened 2020-10-03
Replace EasyHook hooking with Reloaded

goaaats:
this is a direct port of Adam's changes


NotAdam:
fyi this will break all hooks because the package on nuget doesn't include the default calling convention for x64 hooks

goaaats:
Ah oops, haven't tested myself yet cause I can't play the game atm, was waiting for Cara to test - will probably put this on hold then since only penumbra needs it? Or would you rather we submodule it for now?

NotAdam:
the problem with using a submodule is that it breaks because of the discord.net stuff using old versions of system.compilerservices.unsafe, and when I last looked, that was a huge pita to update because it was a few dependencies deep

goaaats:
Works like this now, with all deps having System.Runtime.CompilerServices.Unsafe v4.5.2

daemitus:
Mission accomplished.

PR #186, opened 2020-10-11
Expanded PluginLog. CallerMethod is hard.

daemitus:
So I just found out about PluginLog. Getting my plugin assembly name into the logfile is a great idea. But all I can do is error and information? So now you can do more. You can do all the things.

LogVerbose -> LogFatal (because it was originally using Log* styled methods)
and 
Verbose -> Fatal serilog style method names.

They all point to a generic "Emit" method, screwing up GetCallerAssembly when you're loading Dalamud in DEBUG, so some conditionals took care of that. In release inlining solves the problem.

This could probably be solved better with some sort of Fody weaving attribute, but I couldn't find anything useful.

### DIFF
--- a/Dalamud/Plugin/PluginLog.cs
+++ b/Dalamud/Plugin/PluginLog.cs
@@ -1,25 +1,99 @@
+using Serilog;
+using Serilog.Events;
 using System;
-using System.Collections.Generic;
-using System.Linq;
+using System.Diagnostics;
 using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
 
 namespace Dalamud.Plugin
 {
     public static class PluginLog
     {
+        #region "Log" prefixed Serilog style methods
+
         /// <summary>
         /// Log a templated message to the in-game debug log.
         /// </summary>
         /// <param name="messageTemplate">The message template.</param>
         /// <param name="values">Values to log.</param>
         public static void Log(string messageTemplate, params object[] values)
-        {
-            var name = Assembly.GetCallingAssembly().GetName().Name;
+            => Serilog.Log.Information($"[{Assembly.GetCallingAssembly().GetName().Name}] {messageTemplate}", values);
 
-            Serilog.Log.Information($"[{name}] {messageTemplate}", values);
-        }
+        /// <summary>
+        /// Log a templated message to the in-game debug log.
+        /// </summary>
+        /// <param name="exception">The exception that caused the error.</param>
+        /// <param name="messageTemplate">The message template.</param>
+        /// <param name="values">Values to log.</param>
+        public static void Log(Exception exception, string messageTemplate, params object[] values)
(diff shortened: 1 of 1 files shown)


Dalamud/Plugin/PluginLog.cs
```diff
@@ -12,38 +14,262 @@ public static class PluginLog
         /// <summary>
         /// Log a templated message to the in-game debug log.
         /// </summary>
+        /// <param name="level">The log level to emit.</param>
         /// <param name="messageTemplate">The message template.</param>
         /// <param name="values">Values to log.</param>
-        public static void Log(string messageTemplate, params object[] values)
+        private static void Emit(LogEventLevel level, string messageTemplate, params object[] values)
```
karashiiro, line 20: Nit, but is it possible for this to direct to the version of the function with an Exception argument, with that argument null here? As in:
```csharp
private static void Emit(LogEventLevel level, string messageTemplate, params object[] values)
    => Emit(level, null, messageTemplate, values);
```
I'm not sure if Serilog provides an overload that allows that to be done cleanly, if not then this can be disregarded.

karashiiro (changes requested):

Dalamud/Plugin/PluginLog.cs
daemitus: Null exceptions are handled by serilog properly, as in they dont show up.

```
Log.Information(null as Exception, "There is no exception here.");
Log.Information(new Exception("howdy"), "There is an exception here.");
```
```
2020-10-11 18:30:41.732 -04:00 [INF] There is no exception here.
2020-10-11 18:30:41.732 -04:00 [INF] There is an exception here.
System.Exception: howdy
```

karashiiro (approved):

Dalamud/Plugin/PluginLog.cs
```diff
@@ -12,38 +14,250 @@ public static class PluginLog
...
+        /// Log a templated message to the in-game debug log.
+        /// </summary>
+        /// <param name="level">The log level to emit.</param>
+        /// <param name="exception">The exception that caused the error.</param>
+        /// <param name="messageTemplate">The message template.</param>
+        /// <param name="values">Values to log.</param>
+        private static void Emit(LogEventLevel level, Exception exception, string messageTemplate, params object[] values)
         {
+#if DEBUG
+            // Due to DEBUG not inlining the below methods, we need the 2nd CallingAssembly which is the third frame.
```
NotAdam, line 32: this is not guaranteed to be inline at all, even with MethodImpl

daemitus: Yeah I was looking at my own logs when dalamud was in debug and it was jacked up again. Gonna take a re-look at this at some point. I do want to expand this, at the very least for my own logging.

daemitus:
@karashiiro @NotAdam  I've spent enough time on this. Same as it was originally, just more logging methods.

NotAdam (approved):
looks good to me

goaaats (approved):

PR #187, opened 2020-10-17
Remove RMT and bad word detection

ascclemens:
Functionality superseded by NoSoliciting.


goaaats:
I'm not 100% sure on this, since it's a core feature of Dalamud that many people probably don't even consciously notice - however, it's no question that your implementation is superior.
Will think about it.

goaaats (approved):

goaaats:
I decided that we should keep this in for now, as a baseline feature

PR #189, opened 2020-10-18
Catch universalis uploader errors

daemitus:
Haven't tested it, don't really know when the uploader actually fires off. 

See https://github.com/goatcorp/Dalamud/issues/188.

If there's a better way of handling this, I can throw it together.

### DIFF
--- a/Dalamud/Game/Network/MarketBoardUploaders/Universalis/UniversalisMarketBoardUploader.cs
+++ b/Dalamud/Game/Network/MarketBoardUploaders/Universalis/UniversalisMarketBoardUploader.cs
@@ -59,7 +59,15 @@ namespace Dalamud.Game.Network.Universalis.MarketBoardUploaders {
                 }
 
                 var upload = JsonConvert.SerializeObject(listingsRequestObject);
-                client.UploadString(ApiBase + $"/upload/{ApiKey}", "POST", upload);
+                try
+                {
+                    client.UploadString(ApiBase + $"/upload/{ApiKey}", "POST", upload);
+                }
+                catch (WebException ex)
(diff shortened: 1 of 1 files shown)


Dalamud/Game/Network/MarketBoardUploaders/Universalis/UniversalisMarketBoardUploader.cs
```diff
@@ -59,7 +59,15 @@ internal class UniversalisMarketBoardUploader : IMarketBoardUploader {
...
                 var upload = JsonConvert.SerializeObject(listingsRequestObject);
-                client.UploadString(ApiBase + $"/upload/{ApiKey}", "POST", upload);
+                try
+                {
+                    client.UploadString(ApiBase + $"/upload/{ApiKey}", "POST", upload);
+                }
+                catch (WebException ex)
+                {
+                    int statusCode = (int)(ex.Response as HttpWebResponse).StatusCode;
+                    Log.Verbose($"Universalis returned an HTTP {statusCode}");
```
goaaats, line 69: These should be Log.Error and have ex attached, so you can still see the stack trace

goaaats (changes requested):

Dalamud/Game/Network/MarketBoardUploaders/Universalis/UniversalisMarketBoardUploader.cs
daemitus: Changes made as requested.

Are the stacktraces really necessary in this instance? Outside of I guess the line number, the rest is pretty standard since its a non-200.

PR #195, opened 2020-11-01
Added a label and a warning for dev plugins in the installer window

Aireil:
When a plugin is loaded raw from the devPlugins folder AND is in the installer window, display a label and a warning concerning the installed location. This also removes the "Disable" button, as it's not possible to disable a dev plugins and it just produces an error.
This is what it looks like in game:
chrome_2020-11-01_17-59-08
~~It detects dev plugins when they have a dummy configuration, as they usually don't have one when a user try/install a dev plugin.
For it to work with all dev plugins, it would be easy to add a property when loading it in LoadPluginFromAssembly(), but I wasn't sure on how to do it without breaking things.
Another option would be to force raw plugins to load the dummy config.~~
The string isn't localized either as it's pretty advanced, not sure what is the "usual" rule on this.

### DIFF
--- a/Dalamud/Plugin/PluginInstallerWindow.cs
+++ b/Dalamud/Plugin/PluginInstallerWindow.cs
@@ -99,9 +99,16 @@ namespace Dalamud.Plugin
                     }
 
                     didAnyWithSearch = true;
-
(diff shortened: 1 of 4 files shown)


Aireil:
The detection of dev plugins is now reliable thanks to the PluginLoadReason mask. Let's hope I implemented it as goat thought!

PR #202, opened 2020-11-25
Localized languages in settings

Aireil:
Replacing the language 2-letter codes. There's an exception for Korean, which stayed in English, as the Korean language isn't supported. If anything goes wrong, fallbacks to the language codes. Here is how it looks in game:

ffxiv_dx11_2020-11-25_23-01-45


goaaats (approved):

goaaats:
Will have to look at KO font support soon

PR #213, opened 2020-12-09
Disable plugins with incompatible API level

Aireil:
This is "needed" due to #206, not disabling incompatible plugins that were enabled before means that they'll never get cleaned up.


goaaats:
this REALLY needs a rewrite

goaaats (approved):

PR #218, opened 2020-12-12
Fix testing plugins install

Aireil:
The json of plugins in their directories doesn't have the generated fields from the plugin master, this means that installing any testing plugins will fail. This is a not a clean fix, only a band-aid.


goaaats:
I'd rather just remove that completely then, honestly

Aireil:
Same

goaaats:
understandable, have a nice day

Aireil:
See you next time.

PR #219, opened 2020-12-13
Add an auto-updater for plugins

Aireil:
Off by default, it auto-updates plugins when logging in with a character.

I moved everything in a task, tested it with various scenarios, but please verify that it is satisfactory.

Setting and in-game message:
ffxiv_dx11_2020-12-13_04-10-01

### DIFF
--- a/Dalamud/Game/ChatHandlers.cs
+++ b/Dalamud/Game/ChatHandlers.cs
@@ -212,20 +212,34 @@ namespace Dalamud.Game {
                 this.dalamud.Configuration.Save();
             }
 
-            try {
-                var hasNeedsUpdate = this.dalamud.PluginRepository.UpdatePlugins(true).UpdatedPlugins.Count != 0;
+            Task.Run(() => this.dalamud.PluginRepository.UpdatePlugins(!this.dalamud.Configuration.AutoUpdatePlugins)).ContinueWith(t => {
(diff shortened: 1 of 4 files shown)


Dalamud/Game/ChatHandlers.cs
```diff
@@ -212,20 +212,34 @@ public class ChatHandlers {
...
-            try {
-                var hasNeedsUpdate = this.dalamud.PluginRepository.UpdatePlugins(true).UpdatedPlugins.Count != 0;
-
-                if (hasNeedsUpdate) {
-                    this.dalamud.Framework.Gui.Chat.PrintChat(new XivChatEntry
-                    {
-                        MessageBytes = Encoding.UTF8.GetBytes(Loc.Localize("DalamudPluginUpdateRequired", "One or more of your plugins needs to be updated. Please use the /xlplugins command in-game to update them!")),
-                        Type = XivChatType.Urgent
-                    });
+            Task.Run(() => this.dalamud.PluginRepository.UpdatePlugins(!this.dalamud.Configuration.AutoUpdatePlugins)).ContinueWith(t => {
```
goaaats, line 215: You should keep catching exceptions in here, in case this fails

goaaats (changes requested):

goaaats (approved):

PR #224, opened 2020-12-16
Add Payload for Quest

Caraxi:


goaaats:
I think we should add the hook/event for this as well, cool idea to use as custom chat link

goaaats (approved):

Caraxi:
Was planning on it. Was just putting this in by itself since its simpler lol

PR #228, opened 2020-12-20
Add loading failures in the welcome msg

Aireil:
Avoid being surprised when a plugin didn't load for whatever reason.


Aireil:
Updated to add API level and game version incompatibility. API-level load failures now include the .disabled file in order not to spam the fail to load until update or never (\#RIP job icons).
Errors are displayed every time, as it shouldn't happen every time.

PR #242, opened 2021-01-13
Protect users from themselves

Caraxi:
Add checks for adding blank and duplicate repo as blank repos were not deleteable and broke the installer.


goaaats (approved):

goaaats:
I wanted to make this as user-hostile as possible but I now recognize that I was misguided

PR #243, opened 2021-01-13
Update SigScanner.cs

pohky:
bit of a speed boost

### DIFF
--- a/Dalamud/Game/SigScanner.cs
+++ b/Dalamud/Game/SigScanner.cs
@@ -130,6 +130,7 @@ namespace Dalamud.Game {
             Log.Verbose("module copy START");
             // .text
             this.moduleCopyPtr = Marshal.AllocHGlobal(Module.ModuleMemorySize);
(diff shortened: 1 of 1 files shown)


Dalamud/Game/SigScanner.cs
```diff
@@ -120,170 +126,166 @@ public sealed class SigScanner : IDisposable {
             }
         }
 
-        private IntPtr moduleCopyPtr;
-        private long moduleCopyOffset;
+        private IntPtr _moduleCopyPtr;
+        private long _moduleCopyOffset;
```
goaaats, line 130: Why the style change 🤔

goaaats (changes requested):

Dalamud/Game/SigScanner.cs
Caraxi: looks like their linter went wild and changed a bunch of random things, like the 4 spaces on all the comments lol

pohky: wups yeah, think i still had it set to auto format/clean on paste

goaaats (approved):

PR #247, opened 2021-01-18
Fix misidentified market tax rates packet

Aireil:
Thanks Miu 👍 .


Aireil:
Should still be working in 5.45, didn't test it, but it looks good in ffxivmon 👍.

goaaats (approved):

PR #252, opened 2021-01-22
wip actor improvements

NotAdam:
things to do:

- [x] remove all the old actor cache garbage
- [x] why is everything an int when it should be uint?

### DIFF
--- a/Dalamud/Game/ClientState/Actors/ActorTable.cs
+++ b/Dalamud/Game/ClientState/Actors/ActorTable.cs
@@ -19,22 +19,6 @@ namespace Dalamud.Game.ClientState.Actors {
 
         private const int ActorTableLength = 424;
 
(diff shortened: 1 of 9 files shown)


Caraxi:
... I edited it by accident but 
>  why is everything an int when it should be uint?

is because fixing it is breaking

NotAdam:
should just break it then because it's not correct, `DataId` is also wrong, should be named what it actually is instead of something generic

goaaats:
We should check game state in ActorTable and remark that you should never store these for more than an update tick, if we're gonna end up going for unsafe, I guess

NotAdam:
yeah i have some incredible patent pending™ technology to deal with this in an easy way

goaaats:
We ported all of this pretty much, thanks for your work

PR #254, opened 2021-01-25
Reload plugins when change in settings + fix scaling

Aireil:
Reload plugins when saving settings in case the installer is open, avoiding to close and reopen.
Fix third repos always being saved due to a shallow copy of the list (is this how you implement a Clone() method? 👀).
Fix a missing \n in the injector (really important 👍 ).

Some ImGui witchcraft to fix scaling in settings.

Experimental tab


Aireil:
Added repo number to the installer and settings, in order to see which repo a plugin comes from without having to show the whole link (domain wasn't enough).

Before - After

Settings

goaaats (approved):

PR #257, opened 2021-01-28
Fate Table implementation

rreminy:
Fate Table implementation similar to the current Actor Table

image

Thanks to Anju and LeonBlade


rreminy:
Just had a crash on logout, please hold on before merging

rreminy:
Alright this should be ready for merge now.

- Used `SafeMemory` instead of `Marshal` where applicable
- Implemented `GameData` property for `Fate`
- Also extended `SafeMemory` to include `IntPtr` variants of `ReadBytes` and `WriteBytes`

The cause for the crashes were due to pointers causing an `AccessViolationException`, which cannot be caught the normal way. I also had to removed the `Name` properly as a result and opted to implement the `GameData` properly to retrieve additional information for `Fate`.

*(It took me about 30 crashes and extensive debugging to fix this)*

rreminy:
We're actually figuring other approaches to this problem without reducing functionality like I did. This is safe to merge but we're in the hopes that we can find a safer we can do this without the risk of crashing.

Will let us know of our progress.

goaaats:
Cool, thanks - next dalamud update will take a while still, so take your time, I'll leave this open while you're at it

rreminy:
Working with what looks like memory corruption and other problems is a pain, however we figured this out.

Usage of `SafeMemory` has been removed completely and ported the work done from Sonar into this PR. The fate table logic will no longer attempt to read corrupted / deallocated memory and should therefore be safe to use.

While I done extensive testing of this I would like for some further testing to be performed (again) to make sure, so if anyone have time please give it a time to do so. The following changes are made::

- The first/last fate pointers can change even without calling the hooked function. To take advantage of this only the manager pointer is saved now, then the first/last pointers are resolved when requesting the table.
- I also found out what looks to be a pointer 16 bytes before that coincidentally happen to be IntPtr.Zero during zone transitions. Logic has been added to take advantage of that detail and avoid scanning the table during these periods as the entries can be invalid / deallocated.
- Apparently the whole structure gets corrupted (likely as a result of destructing / deallocation and reuse) during logout. Logic has been added to detect such a case and invalidate the pointer.

rreminy:
Updated with current Dalamud changes

daemitus:
I'm taking a stab at converting this to how we're doing unsafe Actors.

rreminy:
Let me know if you need some assistance with it

The conflicts are likely caused as a result of me placing the Fate Table in the Dalamud Data window after the Actor Table. The numbers must have likely ran off again

An improvement to avoid this from happening is to use an enum instead

daemitus:
Did the enum this morning. :p

daemitus:
See #406

rreminy:
That's quite a lot of changes, I'll have to skim through them but I can't for a week or two.

I can see this PR is also technically merged so I'll be closing it as well.

daemitus:
The short version is instead of making a static object via marshalling,
make a reference to the address where it will read from memory on any
property access. In theory you'd be able to create a new Fate object every
framework tick with extremely little overhead.

It's been working well for actors so far, and the data window works like a
charm.

If you'd like me to take a stab at Sonar I'd love to peek at it too. Mostly
because I miss getting those A rank notifications for 50/60 zones :)

PR #260, opened 2021-02-14
change extension of dalamud log file

Caraxi:
so auto format detection knows what to do with it
also makes more sense having a log file be in a log file extension

### DIFF
--- a/Dalamud/EntryPoint.cs
+++ b/Dalamud/EntryPoint.cs
@@ -56,9 +56,9 @@ namespace Dalamud {
 
         private (Logger logger, LoggingLevelSwitch levelSwitch) NewLogger(string baseDirectory) {
 #if DEBUG
-            var logPath = Path.Combine(baseDirectory, "dalamud.txt");
+            var logPath = Path.Combine(baseDirectory, "dalamud.log");
 #else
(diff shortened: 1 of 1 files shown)


Dalamud/EntryPoint.cs
```diff
@@ -56,9 +56,9 @@ public sealed class EntryPoint : IEntryPoint {
 
         private (Logger logger, LoggingLevelSwitch levelSwitch) NewLogger(string baseDirectory) {
 #if DEBUG
-            var logPath = Path.Combine(baseDirectory, "dalamud.txt");
+            var logPath = Path.Combine(baseDirectory, "dalamud.log");
 #else
-            var logPath = Path.Combine(baseDirectory, "..", "..", "dalamud.txt");
+            var logPath = Path.Combine(baseDirectory, "..", "..", "dalamud.log");
```
goaaats, line 61: can you add that extra ".." while you're at it

goaaats (changes requested):

Dalamud/EntryPoint.cs
Caraxi: sure, done :)

goaaats (approved):

PR #267, opened 2021-02-27
Add Party Finder listing event

ascclemens:
The event is read-only except for the ability to hide listings. It fires once for each populated listing in the Party Finder packet. There are four listings per packet and fifty listings per in-game PF page.

The packet memory must be overwritten if any listing is hidden by a plugin. A separate chunk of memory is leased for this purpose until Dalamud is unloaded, since marshalling the struct directly into the game's memory crashes it (at least it did when I tested it some time ago).

---

I wasn't sure where to put this event and its logic. I put it under `Gui`, since that's where the same logic/events for chat is, but if there's somewhere else, let me know. I'm also not sure if the user-facing classes should go in `Structs`.

Please let me know if there any changes I should make.

### DIFF
--- a/Dalamud/Game/Internal/Gui/GameGui.cs
+++ b/Dalamud/Game/Internal/Gui/GameGui.cs
@@ -10,6 +10,7 @@ namespace Dalamud.Game.Internal.Gui {
         private GameGuiAddressResolver Address { get; }
         
         public ChatGui Chat { get; private set; }
+        public PartyFinderGui PartyFinder { get; private set; }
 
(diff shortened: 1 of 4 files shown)


ascclemens:
For additional context, I have two plugins that both work on PF stuff. One is NoSoliciting and one is BPF.

It would be nice if I could get NoSol to run first, which load priority handles for events that are actually in Dalamud. Manual hooks work in reverse (I think) in that the latest hook runs first, so the load priority actually makes it work backwards.

Adding this as an actual event in Dalamud fixes that and also exposes PF stuff for any other interested plugins.

goaaats (approved):

PR #270, opened 2021-03-03
Fix slow Any method

Aireil:
Just for Adam ❤️


NotAdam:
you should read the bytes out into a u64 and then >0 it 😤

goaaats (approved):

PR #271, opened 2021-03-07
Fix chara offsets

daemitus:
Looks like some values got updated and nobody noticed


daemitus:
Values taken from sharlayan, just like the first time I added them. Looks like they work from a glance at a test imgui window.

goaaats (approved):

PR #274, opened 2021-03-28
add current ui lang and change event

kalilistic:
Add current ui language and a ui change event to plugin interface. This would let plugins follow dalamud language settings.

### DIFF
--- a/Dalamud/Plugin/DalamudPluginInterface.cs
+++ b/Dalamud/Plugin/DalamudPluginInterface.cs
@@ -87,7 +87,23 @@ namespace Dalamud.Plugin
 #else
         public bool IsDebugging => this.dalamud.DalamudUi.IsDevMenu;
 #endif
+        
+        /// <summary>
+        /// Event that gets fired when loc is changed
+        /// </summary>
+        public event LanguageChangedDelegate OnLanguageChanged;
+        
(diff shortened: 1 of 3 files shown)


Dalamud/Plugin/DalamudPluginInterface.cs
```diff
@@ -87,7 +87,23 @@ public class DalamudPluginInterface : IDisposable {
...
+        /// <summary>
+        /// Delegate for localization change with two-letter iso lang code
+        /// </summary>
+        /// <param name="langCode"></param>
+        public delegate void LanguageChangedDelegate(string langCode);
 
+        /// <summary>
+        /// Current ui language in two-letter iso format
+        /// </summary>
+        public string UiLanguage;
```
goaaats, line 105: Can you make this a "private set" property?

goaaats (changes requested):

Dalamud/Plugin/DalamudPluginInterface.cs
kalilistic: Fixed,

goaaats: Looks good, thank you kindly

goaaats (approved):

PR #283, opened 2021-04-04
Sanitizer

kalilistic:
Sanitizer from discussion in discord. Intended for cleaning up strings to use with imgui to avoid ???. Accepts individual or lists of strings and language overrides.

### DIFF
--- a/Dalamud/Plugin/Sanitizer/Sanitizer.cs
+++ b/Dalamud/Plugin/Sanitizer/Sanitizer.cs
@@ -1,0 +1,113 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace Dalamud.Plugin.Sanitizer
+{
+    /// <summary>
+    /// Sanitize strings to remove soft hyphens and other special characters.
+    /// </summary>
+    public class Sanitizer : ISanitizer
+    {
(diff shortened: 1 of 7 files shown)


Dalamud/Plugin/Sanitizer/Sanitizer.cs
```diff
@@ -0,0 +1,113 @@
...
+using System.Text;
+
+namespace Dalamud.Plugin.Sanitizer
+{
+    /// <summary>
+    /// Sanitize strings to remove soft hyphens and other special characters.
+    /// </summary>
+    public class Sanitizer : ISanitizer
+    {
+        private readonly KeyValuePair<string, string> softHyphen = new KeyValuePair<string, string>("\u0002\u0016\u0001\u0003", string.Empty);
```
NotAdam, line 13: why? just put it in a dict by language

NotAdam:
also whats the point of tests that don't test anything that it actually does

kalilistic:
Will do a new PR later with changes.

kalilistic:
For ackwell.

kalilistic:
- Replaced most hard-coded characters with 0x20 filter
- Replaced kvp lists with dictionaries
- Fixed unit test strings so can see characters in github
- Updated unit tests to use theory/inline data

goaaats (approved):

PR #289, opened 2021-04-06
Add functions for toasting

ascclemens:
See title.

For the uninitialised, toasts are messages that pop up in the middle-ish of the screen and fade away.

image


ascclemens:
Should this contain an event similar to that for chat messages? Like `OnToast` with the ability to cancel similar to `OnChatMessage`?

ascclemens:
I added a queue similar to the one in use for printing messages to the chat log. I also added an event that can be subscribed to in order to modify or cancel any toast from the game or plugins.

I'm doing some gross array copies to ensure that the strings end in a null byte. My tests were working fine when I didn't do that, but the game does expect it. I don't think `fixed` statements add null bytes to the end. If someone else knows a less gross way, please share.

goaaats (approved):

PR #295, opened 2021-04-08
Fix circular exception in AutoTranslatePayload

Caraxi:


goaaats (approved):

goaaats:
How could we have let this happen

Caraxi:
How it managed to not cause an issue for so long is the real question

goaaats:
It wasn't me who implemented it

PR #299, opened 2021-04-09
[WIP] Auto-update dalamud-distrib/stg

daemitus:
This currently will auto-update dalamud-distrib/stg if the assembly version changes. However as per some of our comments in discord, that might not be the desirable end state.


daemitus:
This, as written will require a personal access token dropped into secrets named `PAT`. Only way cross-repo pushes work.

goaaats (approved):

goaaats (approved):

PR #310, opened 2021-04-12
[BREAKING] Use SeString for XivChatEntry

ascclemens:
I'm not sure why `XivChatEntry` was laid out like it was, but it's clunky. This change would remove `MessageBytes` and replace it with `Message`, which takes an `SeString`. Since `string`s implicitly convert into `SeString`s, users can just specify a `string` here.

This would also change the `Name` property to use an `SeString`, which is more correct.

This is a **breaking** change, as it changes the public API and will cause compilation errors for anyone who uses `XivChatEntry`.

---

Also Rider fixed some space issues and I'm too lazy to go back and force-push only my relevant changes.

### DIFF
--- a/Dalamud/Game/ChatHandlers.cs
+++ b/Dalamud/Game/ChatHandlers.cs
@@ -88,7 +88,7 @@ namespace Dalamud.Game {
 
         public ChatHandlers(Dalamud dalamud) {
             this.dalamud = dalamud;
-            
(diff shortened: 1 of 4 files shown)


daemitus:
#415 updated version here.

goaaats:
Dae merged this in with #415, thank you

PR #313, opened 2021-04-13
Actor offsets, StyleCop

daemitus:
Blame StyleCop. Did not change Actor.Address to a getter, breaking public change.


daemitus:
These are probably correct given that CompanyTag is nearby and correct, however I don't know what i'm looking for.
```        public const int Customize = 0x1898;  // Needs verification
        public const int CompanyTag = 0x18B2;  <- correct
        public const int BattleNpcTargetActorId = 0x18D8;  // Needs verification
        public const int NameId = 0x1940;  // Needs verification```

goaaats (approved):

PR #328, opened 2021-04-18
feat: Add support for gamepad control of ImGui

fitzchivalrik:
- Alongside a settings option and the ability for plugins to query gamepad state.
  Plugins cannot, however, as of now set an input as 'handled' and block it for the game.
- Switch between letting the game or ImGui handling input via L1+L3 
  (mimicking how mouse navigation is activated via L1+R3 press in game)
- Control ImGui windows via the gamepad in accordance to the proposed button mapping
  https://www.dearimgui.org/controls_sheets/imgui%20controls%20v6%20-%20Xbox.png
- Whether game or ImGui is receiving gamepad input can be be checked via
  `(ImGui.GetIO().ConfigFlags & ImGuiConfigFlags.NavEnableGamepad) > 0`

Potential points of discussion:
- Buttom combo for activating ImGui gamepad control and blocking game gamepad is not configureable
- Plugins can query button state but not subsribe to e.g. an event to e.g. 'handle'
  and supress a button to the game.
  I was not sure whether that is a desireable feature and how to got about
  implementing it so I left it out for now
- Upon activating gamepad control for ImGui, the plugin installer window is _toggled_.
  This allows the user to access almost all relevant settings and plugin configs.
  I think that intead of a toggle an explicit show/hide would be better, however, 
  there is only a ShowPluginInstaller() method, no hide, and I did not want to introduce
  such a big API change.
- There are some quirks with blocking the games input and some decisions with different
  drawbacks needs to be made. Please have a look at the comment starting `GamepadState.cs:197`
  For convenience sake, it is copied down below.

// NOTE (Chiv) Zeroing `ButtonsRaw` destroys `ButtonPressed`, `ButtonReleased`
// and `ButtonRepeat` as the game uses the RAW input to determine those (apparently).
// It does block, however, all input to the game.
// Leaving `ButtonsRaw` as it is and only zeroing the other leaves e.g. long-hold L2/R2
// and the digipad (in some situations, but thankfully not in menus) functional.
// We can either:
// (a) Explicitly only set L2/R2/Digipad to 0 (and destroy their `ButtonPressed` field) => Needs to be documented, or
// (b) ignore it as so far it seems only a 'visual' error
//      (L2/R2 being held down activates CrossHotBar but activating an ability is impossible because of the others blocked input,
//      Digipad is ignored in menus but without any menu's  one still switches target or party members, but cannot interact with them
//      because of the other blocked input)
// `ButtonPressed` is pretty useful so we opt-in to (b).
// This is debatable.
// ImGui itself does not care either way as it uses the Raw values and does its own state handling.

### DIFF
--- a/Dalamud/Configuration/DalamudConfiguration.cs
+++ b/Dalamud/Configuration/DalamudConfiguration.cs
@@ -123,6 +123,11 @@ namespace Dalamud.Configuration
         public bool IsDisableViewport { get; set; } = true;
 
         /// <summary>
+        /// Gets or sets a value indicating whether or not navigation via a gamepad should be globally enabled in ImGui.
+        /// </summary>
+        public bool IsGamepadNavigationEnabled { get; set; } = true;
+
+        /// <summary>
         /// Load a configuration from the provided path.
         /// </summary>
         /// <param name="path">The path to load the configuration file from.</param>
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -98,6 +98,9 @@ namespace Dalamud.Game.ClientState
         /// </summary>
         public KeyState KeyState;
 
+        /// <summary>
+        /// Provides access to the button state of gamepad buttons in game.
+        /// </summary>
         public GamepadState GamepadState;
         
         /// <summary>
@@ -133,7 +136,7 @@ namespace Dalamud.Game.ClientState
 
             this.KeyState = new KeyState(Address, scanner.Module.BaseAddress);
 
-            this.GamepadState = new GamepadState(scanner);
+            this.GamepadState = new GamepadState(this.Address);
 
             this.Condition = new Condition( Address );
 
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -98,6 +98,8 @@ namespace Dalamud.Game.ClientState
         /// </summary>
         public KeyState KeyState;
 
+        public GamepadState GamepadState;
+        
         /// <summary>
         /// Provides access to client conditions/player state. Allows you to check if a player is in a duty, mounted, etc.
         /// </summary>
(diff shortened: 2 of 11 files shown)


goaaats:
I will test this tonight, thank you.
Regarding your comment, I think that blocking all input to the game, regardless of what it is, while you are in the "plugin layer"(maybe make this obvious visually by putting a dimming overlay over the game?), would be best.

This would prevent confusion about what you can do when with your gamepad.

fitzchivalrik:
Sure! Should I rebase the latest changes onto this branch?

Its easy to zero out the Raw mask too and thus block everything.
I was thinking about potential plugins which could implement combos to do something and could not do that with the Pressed mask in those case. But maybe no plugins should do something anyways when the `ImGuiConfigFlags.NavEnableGamepad` bit is set and leave all the input handling to ImGui.

fitzchivalrik:
I played some more with the Gamepad controls and I tend currently towards removing the toggle of the PluginInstaller on 'mode' switch and replace that with an unused button (e.g. R3) which toggles the plugin installer, along side maybe some kinda indicator like you said that one is in 'ImGui control mode'.

The reasoning comes from that it is really convenient now to be able to e.g. 'click' the 'Try On' button of Item Search in the Examine window via gamepad with a quick mode change and back - but the popping on and off plugin window is kinda distracting.
I will first wait for your feedback on that matter though.

goaaats:
Do you think this is ready? I don't have the time right now to do any work on it myself, and it's a rather big change. Not sure if you got in that visual indicator, I feel that would be important for this.

Thanks! No worries about rebasing, I'll take care of it once you give the OK.

fitzchivalrik:
Not anymore! I already rebased on my machine and am in the process to incorporate your feedback. Sorry that it took awhile. I try to update it all this weekend.

fitzchivalrik:
- Rebased to newest dalamud
- Explicitly zero-out digipad + shoulder buttons in `ButtonsRaw` when in gamepad mode to block.
- Some fixes, like a potential race condition.
- Added background overlay when in gamepad mode + text in upper left corner

I feel like this _should_ be good enough for inclusion - but I am of course only one person and could only test it so far.

dalamud_gamepadmode_2

goaaats (approved):

PR #330, opened 2021-04-20
Fix Actor's HitboxRadius' offset

maributt:
According to [FFXIVAPP](https://github.com/FFXIVAPP/sharlayan-resources/blob/master/structures/5.4/x64.json) the proper offset for an Actor's HitboxRadius is 192 (or 0xC0)


goaaats (approved):

goaaats:
thank you for your contribution

PR #344, opened 2021-04-30
fix: skip if no json when cleaning up plugins

Aireil:


goaaats:
What does this actually fix?

Aireil:
It mostly avoids the useless exception in the log for BrowserHost (that we talked about btw, smh head 🐐's memory). Currently you just get a file not found exception and the cleanup stops if there isn't a .json.

Aireil:
someone else can do it then

PR #346, opened 2021-04-30
fix: UiLanguage null when no override

Aireil:


kalilistic:
Might be worth adding a method to Localization to get current language. Then call that from plugin. Otherwise, need to put a lot of localization logic in the plugin interface class. There's also a string constant in localization for fallback can use.

goaaats (approved):

PR #348, opened 2021-05-01
Add native fly text capability

lmcintyre:


goaaats:
I feel like this should be in its own file inside Gui, with the possibility to add events for it later

lmcintyre:
Oh, you mean like with event-handling for subscribing to flytext being added? Can sure do that

goaaats:
Yeah, not sure how useful that would be - in its own file I'd prefer for sure, you don't need to do the event handling for now

lmcintyre:
Should be all set now. I tested the event using a modified version of damage info to ensure it overwrote flytext properly and could change color, etc.

lmcintyre:
Pushed one more fix. It seems that there is a rare issue where the game may not use fly text immediately, and in its attempt to update the UI, it encounters the pointer already freed. I've added a "solution" to this by delaying the free, but if you don't think it's sufficient, I can try to find another solution.

PR #350, opened 2021-05-06
feat: expose chat type in plugin interface

kalilistic:
Plugins can get dalamud chat channel by using pluginInterface.GeneralChatType.


kalilistic:
Noticed an extra save/line break in there. Will add another commit now to fix.

goaaats (approved):

PR #375, opened 2021-06-16
The ultimate fix (for actor names)

daemitus:
Read the conversation [here](https://discord.com/channels/581875019861328007/653504487352303619/854685863002243092)

In .NET5 the previous band-aid actually causes an access violation, because LPUTF8Str will first attempt to read a pointer. It is an LP after all. Why this works in net472 at all is a mystery. Ultimately, Cara's fix was the correct one to begin with. I've added some caching of the name on top of that and an "oh-no" incase there is no `\0` in the byte array.

### DIFF
--- a/Dalamud/Game/ClientState/Actors/Types/Actor.cs
+++ b/Dalamud/Game/ClientState/Actors/Types/Actor.cs
@@ -1,6 +1,7 @@
 using System;
-
+using System.Text;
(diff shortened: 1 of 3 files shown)


goaaats:
Thank god I don't have to deal with this myself

goaaats (approved):

goaaats (approved):

PR #379, opened 2021-06-25
net5 Prep work

daemitus:
Some prep work that can be backported to the current Dalamud. Nothing breaking except MAYBE IconButton, I'm not sure too many people actually use it though.  

* Update CreateHashList.ps1 to create a list with the full relative path. When I last checked, this wasn't tracking items properly with subfolder recursing.
* Fixes up Dalamud.Test. SDK style, formatting, blah blah. Nothing special.
* Add "HookManager" to Dalamud, currently only has the static TrackedHooks from HookInfo, once Reloaded is implemented it will revert to the original assembly pre-hook. 
* Made the debug address log msgs print in hex. Actually useful now.
* Reworked the Interface folder structure. Internal stuff was moved to an internal folder. Added some new component methods. **Changes to IconButton may? be breaking.** To get some of the overloads working I needed to fiddle with the signature.
* Fiddled with the DXGI vtbl resolver. Added the full vtbl as an enum from Reloaded. Works the same as before.
* Added some ScaledThings to ImGuiHelpers, had Dalamud windows make use of them. 
* Internal representation of IPC subscriptions  changed from valuetuple to a simple class. Not exposed, no end user changes.
* ModuleLog, basically a copy of the Serilog Log.* entries, that get prefixed with a ... well prefix. For example, any messages that do [PLUGINR] can be simplified. 
* Removed messagebox stuff from NativeFunctions, we've got a messagebox library, may as well use it.
* UPGRADED most dependencies.


daemitus:
I hate msbuild tests. will fix in morning.

daemitus:
Turns out Get-FileHash was broken with powershell, using pwsh fixes it.

daemitus:
@goaaats ready for review.

daemitus:
mmm. changed my mind. it can wait.

Issue #390, opened 2021-07-11
Move code into /src


goaaats:
We should move the whole solution into /src like XL to clean up the repo a little bit.

daemitus:
So then...

```
api					?
bin					?
build				yes?
Dalamud				yes
Dalamud.Boot			yes
Dalamud.CorePlugin	yes
Dalamud.Injector		yes
Dalamud.Injector.Boot	yes	
Dalamud.Test			yes
DalamudDebugStub	yes
docs				?
lib					?
.editorconfig			?
.gitattributes			no
.gitignore				no
.gitmodules			no
build.cmd			no?
build.ps1				no?
build.sh				no?
CreateHashList.ps1		?
Dalamud.sln			no
Dalamud.sln.DotSettings no
docfx.json			?
filterConfig.yml		?	
index.md				?
LICENSE				no
README.md			no
stylecop.json			yes?
```

daemitus:
#402

daemitus:
Let me know when you'd like to move forward with this and I'll rebase one final time.

Aireil:
I will let you know for sure!

PR #402, opened 2021-07-13
[On Hold] src folder reorganization

daemitus:
#390


goaaats:
Thought about this again and we should probably hold off until we work through some pending PRs.

goaaats:
I'll do this myself once we're caught up

PR #403, opened 2021-07-13
Fix 3rd party install button not working if also on main repo

### LINKED ISSUE #399: Install plugin button from 3rd party repo doesnt work if on main repo
Try to install ezMiniCactpot from my repo.

daemitus:
Fixes #399 

Also:
- after uninstall, plugins should show as available again
- Remove unused property in 3rd party plugin object
- Double stacktrace on plugin load during boot
- Excess whitespace linter complained about

EDIT
Committed on top of the fix plugin cleanup PR, will wait for that to be merged to I can rebase.


goaaats:
Merged into #407 while I was at it

PR #405, opened 2021-07-14
Add 'Disable Auto Login' button to title menu.

SoyaX:
hopefully will make a few less people ask how to disable it
not all, of course.


goaaats:
I'll probably add a general "TitleScreenAction" that plugins and Dalamud can register to.

PR #406, opened 2021-07-14
Feat: Unsafe actors AND Fate table

daemitus:
This is going to thoroughly break things. What else is a new API level for though?

- Disable "make field readonly" IDE suggestion.
- Moved ActorTable collection implementations into a partial class. It's mostly boilerplate anyways.
- Added ActorTable[indexer] by IntPtr. Solves the issue of people needed to reflect Dalamud to make new actors.
- Made BaseResolver.Dalamud private, like it should be (was internal).
- ClassJob and World Resolvers "Id" field made into a property.
- Exposed the casting actor offsets as properties in Chara
- Name and FCTag are now `SeString`
- Any remaining fields in ClientState are now Properties
- Fixed 2 bugs in MemoryHelper
- Finished making properties out of any remaining fields in all the Job Gauges.
(EDIT)
- Resolvers renamed to NameResolver, moved to ClientState namespace. 
  - This is so as not to collide with Lumina types, and they aren't used by only Actors
- Position3 moved to Game namespace. Not just an Actor thing.
- DataWindow now uses an Enum for the types of data, no more fiddling with ints
- *FEAT* Fate table in the same unsafe way as Actors
- Some fiddling with CorePlugin
- MemoryHelper.ReadSeString that takes a ClientStructs.Utf8String*


daemitus:
Any field renaming was from Adam's branch

Dalamud/Game/ClientState/Actors/ActorTable.cs
```diff
@@ -56,37 +53,34 @@ public int Length
...
+            }
+        }
 
-                return null;
+        /// <summary>
+        /// Get an actor at the specified address.
+        /// </summary>
+        /// <param name="address">The actor address.</param>
+        /// <returns>An <see cref="Actor"/> at the specified address.</returns>
+        public Actor this[IntPtr address]
```
goaaats, line 76: Are you sure this is not confusing? Technically, you're not indexing anything.

daemitus: It could be. would you prefer a GetActor  or CreateActor(IntPtr)

daemitus: Well, we could just expose CreateActorReference too

goaaats: I just wanted to ask what the problem with making that public was, I think that'd be more sensible

daemitus: pushed

goaaats (approved):

goaaats (approved):

PR #413, opened 2021-07-16
Expose ChatHandler.MakeItalics as a little helper method

daemitus:
Also add a `params` to the SeString ctor, to allow for `new SeString(payload1, payload2, ...)` without using an explicit array or list.


daemitus:
MakeItalics  might be more appropriate in SeStringManager, but this was where it was originally.

goaaats (approved):

PR #414, opened 2021-07-16
Dev plugin bugs

daemitus:
Dev plugins if they were configured to not load at boot, or failed, would be in a not-loaded not-disabled state where they would fail manual loading later because "enabling" would fail.

Dev plugin settings in the dalamud config were not updating. Maybe the ref was broken. Changed to a dictionary of dll-path : devSettings.


daemitus:
@goaaats get this before you go vacation or whatever crazy thing's taking you out of town please :)

goaaats (approved):

PR #419, opened 2021-07-19
Big ol' pile of dooty

daemitus:
- SharpDX.Desktop is no longer required due to previous changes in the dxgi finder
- Added a bunch of debug logging to the injector
- Added vector extensions
- Moved common extensions into a more common namespace
- Ripped out PartyList until it is re-implemted. No more confusion.
- Broke a bunch of things fixing the remaining StyleCop suppressions
- Added overloads to sharpDX methods in GameGui
- Rename some structs to <Thing>Struct to  not collide with the usage of the struct. e.g. AddonStruct
- PartyFinder types got split up into individual files
- WebClient -> HttpClient


daemitus:
Should EnumExtensions and VectorExtensions get moved into a Dalamud.Extensions namespace?

daemitus:
Will break up later.

PR #420, opened 2021-07-23
Job Gauges

daemitus:
This implements live job gauges similar to how the actor table rework is going. Due to using Activator for instantiation, caching is implemented in the gauge getter. It's reading from live memory now, so there isn't much downside to doing this.

ClientStructs is pointing to my fork, pending inclusion in master from Aers. The job gauge structs were ported over to there.


daemitus:
This will likely break some testing builds people have been working on. Hold until I can get some more PR's ready so we can slam them all at once.

PR #422, opened 2021-07-23
Injector logging, cleanup

daemitus:
Add a bunch of logging to the injector, in-case more debugging is needed for Linux.

Remove `RemotePinnedData` since we're using the buffers from Reloaded.


daemitus:
Weird, build got cancelled.

PR #434, opened 2021-07-23
ImGui fixes: 1.83 and mouseup fix

lmcintyre:
This PR contains updated references to
ImGuiScene:
- Fixes an issue where ImGui would handle a mouseup event for a corresponding mousedown event that the game handled, resulting in a "stuck mouse".
- Updated reference to ImGui.NET:
  - Updated ImGui to 1.83, resulting in a fix for pasting private-use unicode characters from the clipboard. Game symbols can now be pasted into the ImGui interface.
  - [All updates and features for 1.83 can be viewed here.](https://github.com/ocornut/imgui/releases/tag/v1.83) I don't believe there are any plugin-API-breaking changes.


goaaats:
You need to PR your changes to ImGuiScene and ImGui.NET @lmcintyre

goaaats (approved):

PR #435, opened 2021-07-23
Add ImGuiFileDialog

0ceal0t:


daemitus:
Should this get added under components maybe?

0ceal0t:
I think it's fine either way. I put this in its own folder since the components are all just static ImGui draw calls. If it makes more sense though, I'll move it

goaaats (approved):

PR #445, opened 2021-07-31
Serilog -> NLog

daemitus:
This is mostly logging related now.

This PR includes:
- Some tweaks to the core plugin. 
- Switching from Serilog to NLog, to support a persistent dalamud.log that archives itself into XXX suffixed number files. Serilog specifically does NOT support this per an issue I found.
- Additional verbose injector logging for troubleshooting.
- ~Ability to use a COREHOOK build flag for switching to/from Corehook and Reloaded hooks. The default is currently reloaded which has been working well so far.~
- ~Environment variable XL_NO_WIN32_HOOKS which disables the User32.SetCursor and ws2_32.socket hooks, allowing for reloaded to work on Linux/Wine.~
- ~Reloaded does not fully dispose of hooks, so the HookManager now rewrites the original code back into place when Dalamud disposes of itself. This allows for Dalamud to re-inject.~
- Removed obsoleted DalamudPluginInterface.LogX methods.
- Moved the new Log class, ModuleLog, and PluginLog into the Dalamud.Logging namespace. Log is designed to be the new Serilog.Log replacement, as NLog's design pattern is to get a logger object per class. 
- Abused NLog layout renderers to rename NLog.Trace as Verbose, and made the UI reflect this. Looks almost the same as serilog log output now.


goaaats:
> Environment variable XL_NO_WIN32_HOOKS which disables the User32.SetCursor and ws2_32.socket hooks, allowing for reloaded to work on Linux/Wine.

I really don't like this added complexity - what's the point of reloaded in the first place? Why can we not just stay on corehook for now until reloaded is fixed?

goaaats:
This will also break like every Linux user since no one has the env var set

daemitus:
Corehook has the same problem as easyhook, it uses the threadacl to enable/disable hooks, and it just doesnt work on Wine 6. 

Reloaded hooks perfectly fine, but for whatever reason doesn't work on Win32 hooks in Wine 6. it immediately blows up. I've reversed the code, there are no jumps, the export goes straight to code rather than through 2 jumps ala Windows.  Disabling these 2 hooks that we use allows Wine to boot up fine. 

It also solves Aireil's problem with reloaded. Something on their system is already hooking SetCursor, maybe an nvidia DLL, I'm not sure. and Reloaded can't hook the hook properly.  If there's an environment variable already in place that we can detect in Wine, that would solve the linux user problem.

daemitus:
The problem with Reloaded hooks not working on Win32 in general, is that Reloaded does not automatically follow any jumps at the start of the hooked method. You have to do them manually, hence the additional code in the Hook constructor.

goaaats:
Maybe Wine just needs to be not shit

daemitus:
I completely agree

daemitus:
I recall franz talking in the Linux channel with some dude about hooks working in a new proton/wine build, might have some solutions there if i can dig it up. It was within the past day or two.

daemitus:
If you'd like, I can rip that stuff out for now, keep it on the side and use corehook, so we can push ahead with the rest.

karashiiro:
> Environment variable XL_NO_WIN32_HOOKS which disables the User32.SetCursor and ws2_32.socket hooks, allowing for reloaded to work on Linux/Wine.

Can't this just be done with the platform API? [There's a thing for this.](https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?view=net-5.0)

daemitus:
Well the whole point is it emulates windows. But there are plenty of ways
to detect wine. Maybe

karashiiro:
https://forum.winehq.org/viewtopic.php?t=4988#p25888 this post has a possible detection method, along with a lot of encouragement to PR a fix to wine instead of doing this.

daemitus:
Ok. Here's a best/worst of both options push.

If it detects you are a dirty linux user, you get reloaded, otherwise nice and stable corehook. HookManager.DirtyLinuxUser disables the SetCursor and socket hooks. 

This solves the aireil can't hook with reloaded problem due to an existing hook problem and the linux users cant hook with corehook at all problem. 

The detection method is looking for the ntdll.get_wine_version export. I still need to test it, but its not going into master so who cares (for now).

goaaats:
> ntdll.get_wine_version export
This doesn't work on proton since it's hidden

daemitus:
I'll figure something out that works for all the variations of detecting Linux, how about the idea in general.

reiichi001:
> I'll figure something out that works for all the variations of detecting Linux, how about the idea in general.

We can/should stick to making Linux users set XL_WINEONLINUX. It's part of the Lutris installer now. Worst case, we can adjust the injection error message to add a note for Linux users (set this, hop on Discord, whatever). 

It also works for any version of wine, so users still on 5.7 are covered up to bleeding edge 6.14

daemitus:
I'll roll the XL_WINEONLINUX into a later batch after #446

daemitus:
I've removed the hook changes and will submit a separate PR for discussion. Please re-review the remaining changes.

daemitus:
👀 @goaaats

daemitus:
@goaaats I've rebased to fix the conflicts caused by #448

PR #448, opened 2021-08-02
Proposed hook changes

daemitus:
As mentioned in #445, this proposes a dual track of hooks. If XL_WINEONLINUX is defined, reloaded is used which works well enough with some slight modifications, namely following JMP statements in Win32 style hooks. However as of right now win32 hooks don't work in general, so the 2 that Dalamud uses are disabled, seemingly without penalty. If the variable is NOT defined, CoreHook is used, which while slightly outdated, works exactly as we've come to expect with EasyHook.

### DIFF
--- a/Dalamud/Dalamud.csproj
+++ b/Dalamud/Dalamud.csproj
@@ -74,6 +74,7 @@
         <PackageReference Include="Serilog" Version="2.10.0" />
         <PackageReference Include="Serilog.Sinks.Async" Version="1.5.0" />
         <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
+        <PackageReference Include="Reloaded.Hooks" Version="3.2.3" />
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.333">
             <PrivateAssets>all</PrivateAssets>
             <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
--- a/Dalamud/Game/Network/Internal/WinSockHandlers.cs
+++ b/Dalamud/Game/Network/Internal/WinSockHandlers.cs
@@ -3,8 +3,9 @@ using System.Net.Sockets;
 using System.Runtime.InteropServices;
 
 using Dalamud.Hooking;
+using Dalamud.Hooking.Internal;
 
-namespace Dalamud.Game
+namespace Dalamud.Game.Network.Internal
 {
     /// <summary>
     /// This class enables TCP optimizations in the game socket for better performance.
@@ -18,8 +19,9 @@ namespace Dalamud.Game
         /// </summary>
         public WinSockHandlers()
         {
-            this.ws2SocketHook = Hook<SocketDelegate>.FromSymbol("ws2_32.dll", "socket", new SocketDelegate(this.OnSocket));
-            this.ws2SocketHook.Enable();
+            this.ws2SocketHook = HookManager.DirtyLinuxUser ? null
+                : Hook<SocketDelegate>.FromSymbol("ws2_32.dll", "socket", this.OnSocket);
+            this.ws2SocketHook?.Enable();
         }
 
         [UnmanagedFunctionPointer(CallingConvention.Winapi)]
@@ -30,7 +32,7 @@ namespace Dalamud.Game
         /// </summary>
         public void Dispose()
         {
-            this.ws2SocketHook.Dispose();
+            this.ws2SocketHook?.Dispose();
         }
 
         private IntPtr OnSocket(int af, int type, int protocol)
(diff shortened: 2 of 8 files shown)


daemitus:
CoreHook also does not have problems with existing hooks from other systems. Namely the issue Aireil has with something hooking their SetCursor function before Dalamud gets to it. From reviewing their system, I'm blaming something from NVidia.

goaaats:
I'm a bit worried that this will turn into a nightmare when hooking behaviour is slightly different between CoreHook/Reloaded.
No Linux user has that env var set at the moment you can't rely on it.

How much work would actually fixing reloaded be?

daemitus:
Franz was telling me that all linux users should currently have `XL_WINEONLINUX` set. This is not the different variable that I proposed before.

goaaats:
We only started setting it by default on new Linux installs via Lutris like two weeks ago

daemitus:
Here are the current things I'm unhappy with regarding Reloaded, and what I've done to fix them. 
- Does not revert hook on dispose preventing re-injection.
  - HookManager rewrites the original assembly during its own dispose after all hooks are released.
- Does not follow JMP statements, i.e. Win32 styled hooks
  - Hook.cs manually follows 0xE9 and 0xFF 25 jumps if directed to do so, there is an additional ctor to deal with this.

UNRESOLVED ISSUES
- Blows up if something else has already hooked the hook. Aireil has this issue, something hooks their SetCursor. 
  - We can probably work around this by adding more follow-the-hook logic, or copying how CoreHook establishes itself
- Reloaded just doesn't work on Linux Win32 hooks.  Don't know why, don't have a solution. I assume it's a Wine issue. Maybe it works with the Proton build Franz was discussing with Eorn, would need to test. As of right now, my solution is to not hook SetCursor and ws2_32.socket.

daemitus:
Otherwise, Reloaded works fine so far in the limited testing I've done with the plugins that I use daily. We could just go all-in on Reloaded, offer XL_WINEONLINUX, some other checks to detect older linux installs without that variable, and an optional "help me setcursor blows up" variable to just disable it too.

reiichi001:
> We only started setting it by default on new Linux installs via Lutris like two weeks ago

We only started enforcing it a couple weeks ago, but plenty more Linux users have set it. (Mostly the ones who came in for aria issues). 

Unless we get reloaded working perfectly, this might be our best option. We could maybe add a check for on boarding to look for wine-related environment variables if we want to remind a Linux user to set it.

daemitus:
I got aireil's hook problem sorted out by using this logic to follow JMPs. The hook might not actually work, but whatever its hooking is hooked. If anyone knows what the difference in IPRelativeMemoryOperand jmps vs not is, I'd love to hear about it.

```
IntPtr newAddress;
switch (inst.Op0Kind)
{
    case OpKind.NearBranch64:
    case OpKind.NearBranch32:
    case OpKind.NearBranch16:
        newAddress = (IntPtr)inst.NearBranchTarget;
        break;
    case OpKind.Immediate16:
    case OpKind.Immediate8to16:
    case OpKind.Immediate8to32:
    case OpKind.Immediate8to64:
    case OpKind.Immediate32to64:
    case OpKind.Immediate32 when IntPtr.Size == 4:
    case OpKind.Immediate64:
        newAddress = (IntPtr)inst.GetImmediate(0);
        break;
    case OpKind.Memory when inst.IsIPRelativeMemoryOperand:
        // What is the difference with this?
        newAddress = (IntPtr)inst.IPRelativeMemoryAddress;
        newAddress = Marshal.ReadIntPtr(newAddress);
        break;
    case OpKind.Memory:
        newAddress = (IntPtr)inst.MemoryDisplacement64;
        newAddress = Marshal.ReadIntPtr(newAddress);
        break;
    default:
        throw new Exception("Unknown OpKind");
}
```

daemitus:
Then the question is, do we try to follow jumps always or only when the developer requests it from the hook ctor. 

If we unconditionally follow JMP operators, we'll need to check with HookManager so that we don't jump into a reloaded hook trampoline.

I think that makes sense anyways

daemitus:
Ok. Added in the Iced disassembler jump follow code, as far as I can tell, it works great. 
Threw in a few more Wine checks, add/remove more, whatever. 

Launched Dalamud with 2 test addons that both try to hook SetCursor.
```
[VRB] Resolving assembly jump (NearBranch64) from 7FFDA381B120 to 7FFD9FC90CCE
[VRB] Resolving assembly jump (Memory) from 7FFD9FC90CCE to 7FFCACC2F4F0
[VRB] Detected hook trampoline at 7FFCACC2F4F0, stopping jump resolution.
```

I am extremely pleased with how this turned out.

daemitus:
👀 @goaaats

goaaats (approved):

goaaats:
In the end not being able to hook win32 stuff on Linux is not ideal since all plugins that depend on it will not work either, but it'll have to do for now. This will e.g. crash character sync and vfx editor at the moment.

PR #450, opened 2021-08-04
Replace plugin init with IoC & general clean up and improvements

NotAdam:
wip garbage

- [ ] properly expose plugin api
- [x] cleanup references to the dalamud instance everywhere
- [ ] unbreak the scratchpad that i broke
- [ ] better way of passing through the assembly location cause I broke that too (pass string in ctor? lol)
- [ ] some general unshittifying
- [x] break every plugin very hard
- [x] properly separate what should be a dalamud member and what should just be available via DI/service locator
- [x] fix editorconfig some more so it properly enforces stylecop spacing
- [x] fix localisation in plugins
- [ ] add new IPC
- [x] fix internal ctors not being trivially constructable via ioc container


goaaats:
We merged this ourselves, thanks Adam

PR #451, opened 2021-08-08
Job gauges

daemitus:
Job gauge rework, reads live memory instead of marshalling. Dependent on structs that were moved into ClientStructs from Dalamud. Eventually we'll just update our own instead of pointing to Aers'.

Dependent on #445.


daemitus:
Also handles a fix in UiDebug due to updating client structs.

PR #454, opened 2021-08-09
Plugins, bugs, utility

daemitus:
- Moved various extensions to a new utility namespace
- Update plugin installer window to show each dev plugin's filepath
- Don't immediately cleanup outdated GameVersion plugins, allowing for updating. Still enforce not loading.
- Fixes bug with the plugin load/unload button because of duplicate ImGui ID.
- Implement devPlugin load location list. Load your dev plugins from anywhere. Visible in the installer settings window under 3rd party repos.
- Update Assembly.Location and CodeBase hooks due to plugin dependencies breaking.
- More WebClient/HttpClient conversions.


daemitus:
I changed my mind. im going to keep the plugin stuff in a separate FEAT pr.

PR #485, opened 2021-08-19
[WIP] Crash handler

daemitus:
The dalamud.log stuff is very untested atm.


daemitus:
If you copy it all into injector.boot, you can immediately cause a crash in c++ land and test it out rather than having to wait for the game.

PR #488, opened 2021-08-21
ScratchPad

daemitus:
This adds DI to the scratchpad plugin, however it is still BROKEN. For some reason, types referenced by scratchpad are not equal to those supplied by Dalamud. Names and assemblies are the same, however the hashCode is NOT.

I would still merge this though, and continue troubleshooting later.


daemitus:
Also, the LibcFunction was never setup in Service<T>. So added that.

goaaats (approved):

PR #489, opened 2021-08-22
Bugs

daemitus:
- ClientState.Set() was called twice
  - Added a check in Service<T> to prevent it.
- Re-implement Character.NameId
- Fix GameObject.Name length
- Fix FlyTextHook.Dispose breaking on unload/game close
- Gracefully fail when IDalamudPlugin is not found being implemented


daemitus:
I also exposed a list of Plugin names and internal names in the plugin interface. QolBar was reflecting to access these, so there's clearly a need of some kind.

goaaats (approved):

PR #499, opened 2021-08-25
Filter disabled repos

daemitus:
Fair warning, I haven't tested this. It makes sense though. 

SetPluginReposFromConfig is called in 2 places
 - On PluginManager init
 - When the SettingsWindow has changes saved.


Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -1290,6 +1289,7 @@ private void DrawPluginControlButton(LocalPlugin plugin)
                             if (!plugin.IsDev)
                             {
                                 pluginManager.RemovePlugin(plugin);
+                                this.OnAvailablePluginsChanged();
```
daemitus, line 1292: If we move the filtering of what's installed into this method, freshly installed plugins won't remain in the available tab immediately after installation.

However, if we re-filter after uninstalling, the pluginmaster doesnt re-fetch anything, we only re-filter once, and it seems to just work. This I DID test.

Dalamud/Plugin/Internal/PluginManager.cs
```diff
@@ -148,6 +148,7 @@ public void SetPluginReposFromConfig(bool notify)
 
             var repos = new List<PluginRepository>() { PluginRepository.MainRepo };
             repos.AddRange(configuration.ThirdRepoList
+                .Where(repo => repo.IsEnabled)
```
daemitus, line 151: Filter disabled repos here.

PR #502, opened 2021-08-26
feat: Condition change event

Ottermandias:
Introduce an event in the Condition class that should fire whenever a condition is changed by the game.
(also exposed the ConditionArray pointer, but that could be skipped if unwanted)


daemitus:
Ok. caching and update on framework tick implemented, but untested. Somebody mind giving this a joyride?

goaaats (approved):

PR #516, opened 2021-08-28
IPC with cross-plugin objects

daemitus:
I got lazy and did serde with newtonsoft. But it works and is arguably cleaner than before. All the conversion is in a single function, so if you want to do reflection, that's the place to do it.


daemitus:
this is BREAKING due to moving everything into dalamud.plugin.ipc

daemitus:
actually. hold that thought... it might not be.

daemitus:
nah it is. whee

goaaats (approved):

PR #519, opened 2021-08-30
More IPC 

daemitus:


daemitus:
@goaaats this should be a good bandaid for now. At some point we'd probably want to throw some events inside the channel class so that it deserializes from the other assembly, but so long as we dont force typing into the json this worked like a charm in testing.

goaaats (approved):

PR #520, opened 2021-08-30
feat: AsmHook

daemitus:
Requested by unknown.


daemitus:
This should be good to go. Unknown tested it earlier, and I added a name parameter so we know what's being hooked in the stats window.

goaaats (approved):

PR #524, opened 2021-09-01
Add window.preConditionals virtual method

daemitus:
#523


goaaats:
Might want a PostConditionals as well, so you can style the window.

goaaats (approved):

Issue #525, opened 2021-09-01
KeyState stops corresponding to VK enums after 127


UnknownX7:
```cs
private static byte* keyStateIndexArray = (byte*)(DalamudApi.SigScanner.Module.BaseAddress + *(int*)(DalamudApi.SigScanner.ScanModule("0F B6 94 33 ?? ?? ?? ?? 84 D2") + 4));
public static byte GetKeyStateIndex(int key) => key is >= 0 and < 240 ? keyStateIndexArray[key] : (byte)0;
```
will give you the real index of a VK inside KeyState (seems to be a bit bigger than 240, but I think the game blocks keys past 239 anyway)

PR #527, opened 2021-09-01
feat: keystate in data window, fix keyState after 127 not indexed correctly

daemitus:
#525

### DIFF
--- a/Dalamud/Game/ClientState/ClientStateAddressResolver.cs
+++ b/Dalamud/Game/ClientState/ClientStateAddressResolver.cs
@@ -48,6 +48,11 @@ namespace Dalamud.Game.ClientState
         public IntPtr KeyboardState { get; private set; }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


daemitus:
@UnknownX7 can you verify this please

daemitus:
I removed the individual indexers for &1, &2, &3, since theyre largely useless and consumed each tick. Implemented a get/set raw value instead, so people who want to, can set the explicit value themselves.

I'm happy with this PR now, and can be considered for merging.

goaaats (approved):

daemitus:
@goaaats better plan after  talking with Unknown. plugins use the setter to disable game input for custom hotkeys. so now this will throw when setting non-zero values.
```throw new ArgumentOutOfRangeException(nameof(value), "Dalamud does not support pressing keys, only preventing them via zero or False. If you have a valid use-case for this, please contact the dev team.");```

goaaats (approved):

PR #530, opened 2021-09-04
Hook Process.Handle [Discussion needed]

daemitus:
This was developed while troubleshooting pokhy's injector issues. It forces anything using Process.Handle (Reloaded) to use a pseudo-handle of 0xFFFFFFFF. No permission issues. 

Its not necessarily required anymore, but might not be a bad idea to merge.

Opinions?


daemitus:
@NotAdam @goaaats

goaaats (approved):

goaaats:
I assume this won't do any harm, we can always go back and reevaluate if there's an issue

PR #556, opened 2021-09-13
fix: populate available plugins in SetPluginReposFromConfigAsync

daemitus:
When Dalamud boots up, the availablePlugins list is never populated. Hence auto-updates fail.


daemitus:
This also gets rid of the private buffered plugin lists in pluginmanager. Just use the immutable list. This also solves any possible iteration errors as a side effect.

goaaats (approved):

PR #572, opened 2021-09-15
Fix MaxKeyCodeIndex

UnknownX7:


daemitus (approved):
240 == 0xF0, this streamlines things a bit. No issues from me. Does the checking in ConvertKeyCode instead.

goaaats (approved):

PR #575, opened 2021-09-16
Load dev plugin icons from disk, load all icons as needed instead of one by one

daemitus:
Will comment major changes in the PR code.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
@@ -25,7 +25,6 @@ using Dalamud.Plugin.Internal.Types;
 using Dalamud.Utility;
 using ImGuiNET;
 using ImGuiScene;
-using Microsoft.VisualBasic;
 
 namespace Dalamud.Interface.Internal.Windows
 {
@@ -56,7 +55,7 @@ namespace Dalamud.Interface.Internal.Windows
         private string[] testerImagePaths = new string[5];
         private string testerIconPath = string.Empty;
 
-        private TextureWrap?[]? testerImages;
+        private TextureWrap?[] testerImages;
         private TextureWrap? testerIcon;
 
         private bool testerError = false;
@@ -76,9 +75,8 @@ namespace Dalamud.Interface.Internal.Windows
         private List<AvailablePluginUpdate> pluginListUpdatable = new();
         private bool hasDevPlugins = false;
 
-        private bool downloadingIcons = false;
-        private Dictionary<string, (bool IsDownloaded, TextureWrap?[]? Textures)> pluginImagesMap = new();
-        private Dictionary<string, (bool IsDownloaded, TextureWrap? Texture)> pluginIconMap = new();
+        private Dictionary<string, TextureWrap?> pluginIconMap = new();
+        private Dictionary<string, TextureWrap?[]> pluginImagesMap = new();
 
         private string searchText = string.Empty;
 
@@ -197,8 +195,6 @@ namespace Dalamud.Interface.Internal.Windows
         {
             this.pluginIconMap.Clear();
             this.pluginImagesMap.Clear();
-
-            this.DownloadPluginIcons();
         }
 
         private static Vector2 GetButtonSize(string text) => ImGui.CalcTextSize(text) + (ImGui.GetStyle().FramePadding * 2);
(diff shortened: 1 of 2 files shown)


Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -76,9 +75,8 @@ internal class PluginInstallerWindow : Window, IDisposable
         private List<AvailablePluginUpdate> pluginListUpdatable = new();
         private bool hasDevPlugins = false;
 
-        private bool downloadingIcons = false;
-        private Dictionary<string, (bool IsDownloaded, TextureWrap?[]? Textures)> pluginImagesMap = new();
-        private Dictionary<string, (bool IsDownloaded, TextureWrap? Texture)> pluginIconMap = new();
+        private Dictionary<string, TextureWrap?> pluginIconMap = new();
```
daemitus, line 78: No need to save IsDownloaded now. If there is no value, it needs to be downloaded and a placeholder is inserted beforehand. If there is a value, its either null/failed, or a valid tex.

Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -211,10 +207,18 @@ public void ClearIconCache()
...
             if (isThirdParty)
+            {
+                if (manifest.ImageUrls.Count > 5)
+                {
+                    Log.Warning($"Plugin {manifest.InternalName} has too many images");
+                    return manifest.ImageUrls.Take(5).ToList();
+                }
+
                 return manifest.ImageUrls;
+            }
```
daemitus, line 221: Don't error out. Just take 5 and throw a warning.

Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -802,7 +845,7 @@ private bool DrawPluginListLoading()
             return ready;
         }
 
-        private bool DrawPluginCollapsingHeader(string label, PluginManifest manifest, bool trouble, bool updateAvailable, bool isNew, Action drawContextMenuAction, int index)
+        private bool DrawPluginCollapsingHeader(string label, LocalPlugin? plugin, PluginManifest manifest, bool isThirdParty, bool trouble, bool updateAvailable, bool isNew, Action drawContextMenuAction, int index)
```
daemitus, line 848: We need to pass around a nullable plugin so that we can detect devPlugins. For available plugins, we just pass in null here.

Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -875,8 +926,8 @@ private bool DrawPluginCollapsingHeader(string label, PluginManifest manifest, b
 
             // Download count
             var downloadCountText = manifest.DownloadCount > 0
-                                        ? Locs.PluginBody_AuthorWithDownloadCount(manifest.Author, manifest.DownloadCount)
-                                        : Locs.PluginBody_AuthorWithDownloadCountUnavailable(manifest.Author);
+                ? Locs.PluginBody_AuthorWithDownloadCount(manifest.Author, manifest.DownloadCount)
+                : Locs.PluginBody_AuthorWithDownloadCountUnavailable(manifest.Author);
```
daemitus, line 930: formatting.

Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -1512,18 +1563,18 @@ private void DrawVisitRepoUrlButton(string? repoUrl)
...
-            }
+                this.pluginImagesMap.Add(manifest.InternalName, Array.Empty<TextureWrap>());
+                Task.Run(async () => await this.DownloadPluginImagesAsync(plugin, manifest, isThirdParty));
 
-            if (!images.IsDownloaded)
                 return false;
+            }
 
-            if (images.Textures == null)
+            if (imageTextures.Length == 0)
```
daemitus, line 1577: the images array will never be null, either an empty array, or an array with possible nulls in it.

daemitus:
How images were downloaded largely remains the same. Icons on the other hand are now downloaded like the images were. If ImGui gets to the icon and is missing it, it inserts a placeholder in the iconMap and kicks off a task. This results in however many async tasks running all at the same time to download the icons, rather than it being sequential like it was before.

Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
```diff
@@ -225,6 +229,46 @@ public void ClearIconCache()
             return output;
         }
 
+        private static FileInfo? GetPluginIconFileInfo(LocalPlugin? plugin)
```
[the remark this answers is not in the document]
goaaats: I think we should only check the images folder. Will prevent people from pushing wrong setups to the main repo.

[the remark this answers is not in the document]
daemitus: done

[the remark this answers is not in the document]
daemitus: also done in the other usage of "images" too

goaaats (approved):

PR #576, opened 2021-09-19
feat: add an option to disable RMT Filtering

Caraxi:


goaaats:
Why? Are there false positives?

Caraxi:
Occasionally

Caraxi:
any time anyone talks about merch with bonus codes it false positives.

goaaats:
Think we should consider either removing it altogether or improving the regex, even though I think there's still some value in having a basic filter built-in

Caraxi:
improved or left how it is, it should still have an option to disable it if it doesn't get removed.

daemitus (approved):
Looks solid

Issue #578, opened 2021-09-20
Banned plugins still load


Aireil:
They should not load, but be allowed to update (same as outdated plugins).

PR #579, opened 2021-09-20
Delete button for outdated/banned, do banned plugins like outdated so they update.

### LINKED ISSUE #577: Disable/Uninstall outdated plugins from the installer
To be able to remove them from the installer without going in files.

daemitus:
Closes #577 #578


daemitus:
Ok. we extract the deletePlugin button from the dev plugin controls so it can be used if banned or outdated as they never get marked for cleanup, instead they remain so they can be updated.

Add some help text here and there, make shortcuts for IsOutdated and IsBanned, few formatting fixes and ezpz

goaaats (approved):

PR #581, opened 2021-09-21
Fix plugin entries staying open across tabs

attickdoor:
Current behavior: Plugin listing at index i will be open on all tabs if opened on any tabs.
Fix: Push ID of current tab before drawing.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstallerWindow.cs
@@ -440,6 +440,7 @@ namespace Dalamud.Interface.Internal.Windows
 
         private void DrawPluginTab(string title, Action drawPluginList)
         {
+            ImGui.PushID(title);
             if (ImGui.BeginTabItem(title))
             {
                 ImGui.BeginChild($"Scrolling{title}", ImGuiHelpers.ScaledVector2(0, -30), true, ImGuiWindowFlags.HorizontalScrollbar | ImGuiWindowFlags.NoBackground);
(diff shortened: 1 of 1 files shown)


goaaats:
Did you test this? Because the issue is actually caused by the `openPluginCollapsibles` list being shared across tabs (I was lazy).

An easy way to fix this would be to call Clear() on the list whenever tabs are switched.

daemitus:
the quick fix is to probably change openPluginCollapsibles to be internalName + some other identifier

daemitus:
or internalName+version, and append to the dictionary on update so it stays open.

attickdoor:
> An easy way to fix this would be to call Clear() on the list whenever tabs are switched.

I think having persistence between tabs is a good idea.

> or internalName+version, and append to the dictionary on update so it stays open.

Right now I'm doing internalName + version, though this introduces a minor edge case where an installed dev plugin can also exist in other tabs, causing it to mirror its state on both tabs. If that's acceptable then I'll push again.

attickdoor:
I guess the general issue is "a plugin that can exist in multiple tabs at the same time will have its state mirrored", but I can't see a good way around that with the information available to the function.

daemitus:
In the pending pr I've got open, I'm passing the plugin in too, so we could
conditionally add in the dll file path, and other stuff to differentiate
between dev, normal, available and installed

attickdoor:
That would work. I don't think it'd be particularly worth it to pass that data in *just* to fix the entries being open, but if you're using it for something else then I'll gladly make use of it. I suppose I'll wait until your PR is merged.

KazWolfe:
Given that this PR seems to have gone stale, I think it makes sense to close it for now. If this is still a problem and/or we still want this in the future, feel free to reopen it.

PR #582, opened 2021-09-22
api3 plugins will have a null AssemblyName value, needs handling during removal

daemitus:


daemitus:
If a plugin does not load fully, or bails early during Load(), AssemblyName is likely to be null. so we null coalesce everything and mark the field as nullable now.

goaaats (approved):

PR #584, opened 2021-09-22
Add fallible methods to SigScanner

ascclemens:
I make a static class with `TryScanText` in every plugin. I figure other devs might find this useful.


ascclemens:
Force-pushed to fix a method missing `Try` prefix.

daemitus:
May/may not use it myself, but it's got my seal of approval.

goaaats (approved):

PR #592, opened 2021-09-26
Save repo manifest before continuing

daemitus:


daemitus:
Ok, so in this situation, we basically disregard the manifest included in the zipfile. Should we require it and/or update dalamudpackager?

goaaats:
This is sensible, don't think we need to change anything else.

goaaats (approved):

PR #597, opened 2021-09-27
Copy over Bluefissure's support for Simplified Chinese regex

reiichi001:
Copying the Simplified Chinese language regexes from https://github.com/Bluefissure/Dalamud/blob/cn/Dalamud/Game/Command/CommandManager.cs#L23

### DIFF
--- a/Dalamud/Game/Command/CommandManager.cs
+++ b/Dalamud/Game/Command/CommandManager.cs
@@ -24,6 +24,7 @@ namespace Dalamud.Game.Command
         private readonly Regex commandRegexJp = new(@"^そのコマンドはありません。： (?<command>.+)$", RegexOptions.Compiled);
         private readonly Regex commandRegexDe = new(@"^„(?<command>.+)“ existiert nicht als Textkommando\.$", RegexOptions.Compiled);
         private readonly Regex commandRegexFr = new(@"^La commande texte “(?<command>.+)” n'existe pas\.$", RegexOptions.Compiled);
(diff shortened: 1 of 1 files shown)


Dalamud/Game/Command/CommandManager.cs
```diff
@@ -39,6 +40,7 @@ internal CommandManager()
                 ClientLanguage.English => this.commandRegexEn,
                 ClientLanguage.German => this.commandRegexDe,
                 ClientLanguage.French => this.commandRegexFr,
+                ClientLanguage.ChineseSimplified => this.commandRegexCn,
```
Bluefissure, line 43: No `ClientLanguage.ChineseSimplified`

Bluefissure (commented):
There's no `ClientLanguage.ChineseSimplified` unless add ChineseSimplified to the Language enum.
Not sure whether that's good to go, and I don't even know whether the `LocalPlayer.Language` is ChineseSImplified after installing the Language pack (because I never used that).
Or we can simply add a `commandRegexCn` check  to the fail fallback?

goaaats:
Yeah, this won't work. You're going to have to check them for all languages.

Dalamud/Game/Command/CommandManager.cs
goaaats: Yeah, this won't work. You're going to have to check them for all languages.

reiichi001:
I have some new WIP based on the how the mod works. Will be testing it out later when I have time and will update the PR then.

PR #607, opened 2021-09-29
Adding plugin categories

MgAl2O4:
Needs:

- design decisions for PluginCategoryManager.cs - categoryList
- values in plugin manifests (CategoryTags: [ categories... ]) or making fallback maps more accurate
- localization

### DIFF
--- a/Dalamud/Interface/Internal/PluginCategoryManager.cs
+++ b/Dalamud/Interface/Internal/PluginCategoryManager.cs
@@ -48,9 +48,23 @@ namespace Dalamud.Interface.Internal
         private Dictionary<PluginManifest, int[]> mapPluginCategories = new();
         private List<int> highlightedCategoryIds = new();
 
+        /// <summary>
+        /// Forces plugin category tags, overrides settings from manifest.
+        ///   key: PluginManifest.InternalName (lowercase, no spaces),
+        ///   value: list of category tags, <see cref="categoryList"/>.
+        /// </summary>
+        private Dictionary<string, string[]> mapPluginCategoryTagOverrides = new();
+
+        /// <summary>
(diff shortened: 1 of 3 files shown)


Dalamud/Interface/Internal/PluginCategoryManager.cs
```diff
@@ -0,0 +1,496 @@
...
+            ["vfxeditor"] = new string[] { "UI" },
+            ["visibility"] = new string[] { "UI" },
+            ["voidlist"] = new string[] { "UI" },
+            ["waymarkpresetplugin"] = new string[] { "UI" },
+            ["wintitle"] = new string[] { "UI" },
+            ["woldo"] = new string[] { "UI" },
+            ["wondroustailssolver"] = new string[] { "minigames" },
+            ["xivchat"] = new string[] { "social" },
+            ["xivcombo"] = new string[] { "jobs" },
+        };
```
goaaats, line 149: I kinda want to avoid hardcoding this in Dalamud, at least in here - would be better to do it when generating the pluginmaster additionally to having a "uncategorized"

goaaats (changes requested):

Dalamud/Interface/Internal/PluginCategoryManager.cs
kalilistic: Not sure if casing matters but there's "other" and "Other".

MgAl2O4: right, forgot that pluginmaster is also a script that can be messed with. 

can move fallback category generation in there (it's supposed to be temporary anyway, just to avoid forcing people to update manifests right now)

goaaats (approved):

PR #612, opened 2021-10-02
scrollbar, implicit types

daemitus:


daemitus:
Ok. here's what we've fixed so far. 
 - some explicit type usage. whatever.
 - vertical scrollbar not showing in the plugin list (usage of incorrect variable instead of -1)
 - An EndChild inside the BeginChild closure.
 - nulls as image5 added a spacer at the end of the list. Filter nulls now.
 - Commented unused methods from category view superceding them.
 - Take the ItemSpacing from said unused method, roll into the new view.
 - Remove code from the category view, and merge into DrawAvailablePlugins instead.

There is an existing issue now where when you install a plugin and have that plugin's collapsible open, the installed version is closed again. It's minor, but shouldn't be happening and is a regression. Since we're still using index as isOpen for the collapsible, i'm not sure why it's happening but I've got stuff todo this morning. So problem for later.

goaaats (approved):

PR #628, opened 2021-10-08
feat: add ban reason in the installer

Aireil:
Pulled from assets.
Thoughts? 🤔

Ban reason on top, bottom is the default message if no reason.
Demo


daemitus:
If this has been tested, i like the idea

PR #638, opened 2021-10-10
Replace Harmony with up-to-date RuntimeDetour

0x0ade:
This PR replaces Harmony and the manual MonoMod.Common update to instead fully switch to MonoMod.RuntimeDetour.  
(Disclaimer: I'm a contributor to Harmony and the lead maintainer of the MonoMod libraries.)

Both Harmony and MonoMod.RuntimeDetour use MonoMod.Common under the hood, but latter is a bit "closer to the metal" so to speak. And given how [Wine's library overriding feature seems to interfere in weird ways](https://github.com/MonoMod/MonoMod.Common/commit/22d6fccb8068c5fb17a38065b3e42611a709766e), I personally feel like it might be worth it to cut down on anything that could introduce further weird interactions. (Having `HARMONY_DEBUG` act as a workaround for that issue was *really unexpected*, but in hindsight, it makes some sense for the affected hook specifically.)

I've tried to match the existing codestyle somewhat, and only replace as little as necessary. This also meant replacing the `AccessTools` calls with bare reflection, and updating the documentation comments with hopefully adequate replacements. (Just wondering: Is my usage of `_ = new` fine by your standards?)

Harmony is an amazing library. I want to make sure that all further issues can be fixed ASAP though, so no matter if this PR gets merged or not, feel free to poke me at any time and I'll try to help out whenever I can, even with Harmony problems 😅

By the way, I'm sorry to say that **MonoMod.Common 21.10.08.11 comes with a regression that went unnoticed in all previous testing, which means that even if you're not going to merge this PR, I still highly advise you to update to 21.10.10.01+ [which contains a fix for it](https://github.com/MonoMod/MonoMod.Common/commit/9c150835aba243e62606f4a6f8650946d5aba503).** Again, sorry for the inconvenience - I'll try to figure out why testing didn't end up triggering that stupidly obvious oversight.

And finally, thanks y'all for this and the quick launcher! <3

### DIFF
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Reflection;
 using System.Runtime.CompilerServices;
@@ -24,7 +23,6 @@ using Dalamud.Logging.Internal;
 using Dalamud.Plugin.Internal;
 using Dalamud.Plugin.Ipc.Internal;
 using Dalamud.Support;
-using MonoMod.RuntimeDetour;
 using Serilog;
 using Serilog.Core;
 using Serilog.Events;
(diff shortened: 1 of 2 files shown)


daemitus:
Going to give this a test drive... tomorrow probably. See if I can't rope in a nix user too with some plugin that uses assembly location.

daemitus:
I've generally only added "_" for pinvoke stuff when the linter complains about it. either or is fine imo.

PR #646, opened 2021-10-14
Corehook? what?

daemitus:
feat: EnvironmentSettings.cs, gather up our environment variables in a single location.
feat: CoreHook hooks, new bool parameter on a new ctor/method to use CoreHook instead.
feat: DALAMUD_FORCE_COREHOOK environment variable incase Reloaded REALLLLY doesnt work. Also added to troubleshooting.
move TaskTracker below framework, it depends on it.
migrate our 2 existing win32 hooks to corehook hooks.


goaaats:
Let's run this for a day or two in testing, see if an issue crops up.

PR #677, opened 2021-11-04
Fixes

daemitus:
Couple things I've found.

- [FIX] MinHook currently crashes on plugin reload. This is because the Hook.Dispose() method nulls the list itself in the multi-hook tracker, rather than the hook index in the list.
- [EDIT] Change DALAMUD_FORCE_COREHOOK to DALAMUD_FORCE_MINHOOK
- [FEAT] Add DALAMUD_FORCE_RELOADED
- [FEAT/FIX] TaskTracker is currently a "memory leak" in that it grows forever, quickly, with certain plugins (Sonar) and you can never clear it once it goes over the "crash imgui" size. This limits the size to 1000, and adds a variable so that viewed/expanded tasks are not cleared. 
- [FIX] TaskTracker monohook does not bail when the hook fails. It will now.
- [FIX] A previous change of mine to add isMinHook broke Assembly.CallingAssembly semantics. This is a fix.

### DIFF
--- a/Dalamud/Configuration/Internal/EnvironmentConfiguration.cs
+++ b/Dalamud/Configuration/Internal/EnvironmentConfiguration.cs
@@ -23,9 +23,14 @@ namespace Dalamud.Configuration.Internal
         public static bool DalamudNoPlugins { get; } = GetEnvironmentVariable("DALAMUD_NOT_HAVE_PLUGINS");
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


daemitus:
Also change the log window to always have a horizontal scrollbar, fixes a graphical issue when a long line is near the top of the window.

PR #680, opened 2021-11-11
Fixes, formatting, http oh my

daemitus:
- Extract all usages of httpclient into a static global in Util. This follows best usage patterns by not reinstantiating over and over.
- Marketboard history packet: should be 20 not 10
- Marketboard history packet: listing.CatalogId is better defined as NextCatalogId, its 0 when there are no more items.
- Some formatting, internal->private stuff. Use the MBType.Read, dont make your own.


daemitus:
- Used continuation methods for the async upload tasks for error logging.
- Extracted the cfPop code into its own method.

goaaats (dismissed):

goaaats (approved):

PR #711, opened 2021-12-09
Aetheryte List

pohky:
added a list of available aetherytes as a service


goaaats:
Would you mind adding a small test for this to the self-tests here?
https://github.com/goatcorp/Dalamud/tree/master/Dalamud/Interface/Internal/Windows/SelfTest/AgingSteps

goaaats (approved):

PR #714, opened 2021-12-09
All sorts of gauge stuff

daemitus:
Illion had to clarify what  he said to me earlier about the chakra count field. So another bump is due.


goaaats (approved):

daemitus:
DRG "BOTD" values have been renamed to LOTD, as BOTD is always on and has no timer anymore.

goaaats (approved):

PR #716, opened 2021-12-10
Update Dalamud.Game.ClientState.Objects.Enums.StatusFlags for 6.0

SheepGoMeh:


Dalamud/Game/ClientState/Objects/Enums/StatusFlags.cs
```diff
@@ -34,23 +34,23 @@ public enum StatusFlags : byte
...
+        /// Character is a alliance member.
         /// </summary>
-        Friend = 64,
+        AllianceMember = 64,
 
         /// <summary>
-        /// Character is casting.
+        /// Character is in friend list.
         /// </summary>
-        IsCasting = 128,
```
goaaats, line 54: Why is IsCasting gone? Is it crafting instead now?
This would break the API.

goaaats (changes requested):

Dalamud/Game/ClientState/Objects/Enums/StatusFlags.cs
SheepGoMeh: For whatever odd reason, it seems gone, and everything from PartyMember onwards has been shifted.
What was PartyMember was lighting up when people were crafting, so assuming it's that.

UnknownX7:
OffhandOut is something else too, seems to become true the moment you gain WeaponOut and never returns to false until you change zones (or the character is reloaded). Additionally, I think PartyMember and AllianceMember are only true for people in your party/alliance now.

Dalamud/Game/ClientState/Objects/Enums/StatusFlags.cs
WesselKuipers: I was looking at this just now and noticed that one of my friends had the flag of `IsCasting` and nothing else set to true. It looks like everything got shifted?

PR #720, opened 2021-12-24
Print dotnet stack trace on veh error msgbox, and more

Soreepeong:
* Added function on EntryPoint that will be called from VEH exception handler, which will add stack trace to whatever was given from handler and prompt it to user.
* Use precompiled headers.
* Removed `_wcsdup` as it was unnecessary.

image

Example log file (different crash): [dalamud_appcrash.log](https://github.com/goatcorp/Dalamud/files/7772443/dalamud_appcrash.log)


goaaats:
Looks good, i'll merge this so people on testing can have a go at it

goaaats (approved):

PR #739, opened 2022-01-25
Add SignatureHelper

ascclemens:
It's just [Siggingway](https://git.annaclemens.io/ascclemens/Siggingway) ported into Dalamud.

This PR makes every plugin object have `SignatureHelper.Initialise` called on it when loaded, as well. Public so users may call it in other classes that need it.

Adds the `Signature` attribute, allowing for reduction of boilerplate code when working with signatures. Throw `[Signature("DE AD ?? EF")]` on a property or field and `SignatureHelper.Initialise(this)` will fill it. Comes with support for fallible signatures, hooks, pointers, functions, delegates, and offsets.


ascclemens:
`SignatureHelper.Initialise` can throw when signatures fail and aren't marked/inferred to be fallible, but I'm not sure if `LocalPlugin.Load` should catch that or not. It should be as if the plugin did `SigScanner.ScanText` and that threw in their constructor.

goaaats (approved):

PR #749, opened 2022-01-31
(fix) Don't show properties that require parameters in ShowStruct

Caraxi:
Stop exploding


goaaats (approved):

daemitus:
Properties can have parameters?

Caraxi:
only place I've seen it is in the index getters yea

PR #769, opened 2022-03-03
Add support for font gamma

Soreepeong:
Dalamud should be using Noto Sans **Regular** instead of Medium after this PR is merged.

### Noto Sans Medium with gamma value of 1.0 (linear space)
image

This is how it is without this PR.

### Noto Sans Medium with gamma value of 1.4 (Windows default)
image

The game's prebaked font files are also based on gamma value of 1.4.

### AXIS Regular with gamma value of 0.8
image

## Misc
* Screenshot compare: http://www.framecompare.com/screenshotcomparison/WDPG7NNX
* Also, fixed GameFontHandle.Text functions not counting in window scrolls.

### DIFF
--- a/Dalamud/Configuration/Internal/DalamudConfiguration.cs
+++ b/Dalamud/Configuration/Internal/DalamudConfiguration.cs
@@ -135,6 +135,15 @@ namespace Dalamud.Configuration.Internal
         public bool UseAxisFontsFromGame { get; set; } = false;
 
         /// <summary>
+        /// Gets or sets the gamma value to apply for Dalamud fonts. Effects text thickness.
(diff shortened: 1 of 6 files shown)


goaaats:
So I'll replace Noto in assets with the regular version?

Soreepeong:
Please check whether Regular looks fine after this - I tried replacing files in all folders in `%APPDATA%\XivLauncher\dalamudAssets` but I failed to see any difference.

goaaats:
Will check, all assets there are hash checked when you start the game so it probably overrode it immediately

goaaats (approved):

PR #771, opened 2022-03-07
Added GetWindows() function

LadyDefile:
# Dalamad.Interface.Windowing.WindowSystem.cs
* Added GetWindows() function to return the window collection as an array. Returning as array not List<Window> to protect encapsulation.


goaaats:
In a more idiomatic way, foregoing the array conversion, you can write this as:

```cs
public IReadOnlyList Windows => this.windows;
```

LadyDefile:
@goaaats I changed it to use IReadOnlyList.

goaaats:
You should make it a property actually - makes more sense here as the function doesn't actually do anything and we don't usually do getter functions like that. Sorry, should have been more clear.

LadyDefile:
@goaaats Done. Sorry to make you work so hard. LOL

goaaats (approved):

PR #773, opened 2022-03-10
Disable normalizing to dos path and workaround for WINE

Soreepeong:
Instead of checking which file did memory region got mapped from, this will compare the memory content of the loaded executable, and compare it with what's on the disk. Works on Windows 11 at least, but requires testing on other platforms and need to attempt this on WINE.


goaaats (approved):

goaaats:
Let's merge this and see how people on wine fare

PR #787, opened 2022-04-09
ContextMenu null check

Ottermandias:


goaaats:
Could you target the net5 branch instead, if it's not too much trouble? Thanks!

PR #794, opened 2022-04-15
Do not change plugins' custom font sizes

### LINKED ISSUE #792: Dalamud overrides custom plugin font sizes
https://github.com/goatcorp/Dalamud/blob/9a2e2f1024c697bdc446d9f5d815117cfe7643e2/Dalamud/Interface/Internal/InterfaceManager.cs#L815-L821

See the last line of the snippet above. This breaks plugins that load their own fonts, making them blurry and unreadable. Dalamud should not be arbitrarily deciding custom font sizes.

ascclemens:
Close #792.


ascclemens:
https://github.com/goatcorp/Dalamud/pull/796

PR #818, opened 2022-04-23
Add Chinese Language Code

jackiexyz:


goaaats:
This won't work out of the box because of fonts, but we might get ImGui that can handle big fonts soon, so we'll see

jackiexyz:
I think it can display Chinese just fine?
Screenshot 2022-04-24 063542

goaaats (approved):

goaaats:
Yeah, you're right, should be fine

PR #825, opened 2022-04-29
Check punchlines while searching for plugins

### LINKED ISSUE #684: Add option to include Plugin Description Text in Search bar
**What is your suggestion/feature request?**
Add ability to search text descriptions of plugins.

**Describe how you'd like this to work**
Add a checkbox next to Search field of main window with "Include Descrptions" as text. Selecting this checkbox will alter the search to include the title AND the description. 

**Additional context**
As a use case example.. a plugin search for "chat" will display 7 results that have "chat" in the plugin name, but MapLinker. teleporter and owo plugin are not shown.

redstrate:
Small change, but this fixes #684 and also a minor gripe of mine. For example, now searching for "chat" shows the Teleporter plugin which the issue mentioned:

image

And this also fixes another issue where some plugins might look like they're missing because they can't be searched easily:

image
image

The issue mentions searching descriptions, but they probably really only meant the punchline since that's whats visible by default in the list anyway.


goaaats (dismissed):

goaaats:
I cherry picked this onto our dev branch, since master is on .NET 6 - thanks!

PR #833, opened 2022-05-05
Fix MapLinkPayload to account for offset

electr0sheep:
If you like it, great, but mostly wanted to point out that I think this would be pretty easy to fix, without breaking anything.

Also hoping the comments make it easier to understand.

I was staring at this for too long I think trying to make sense of it, so be merciful 🙂


goaaats:
You may want to rebase this one against the net5 branch, master is already on net6 and not mainline, currently

electr0sheep:
@goaaats Sorry, life kinda got in the way. It's been a while since your comment, let me know if you need me to do anything.

goaaats (approved):

PR #835, opened 2022-05-07
Add dev state / notice

kalilistic:
This adds the feature to let users see warnings about plugins and also if a given plugin is waiting for an update or has been deprecated.

This introduces the following plugin fields:
- DevSupportState
- DevSupportStateReason
- Notice

The fields can be loaded from either dalamud assets or the plugin manifest. If both are set, the dalamud assets version will take precedence. In dalamud assets, these are defined in a new pluginmetadata file which can be extended to add more fields that would benefit from central control.

DevSupportState is an enum that defines the level of support that is currently provided for a plugin. There were discussions in discord/issue#814 with a few more options but I've started with a smaller subset for now.
- Active - new updates/features (default)
- MaintenanceOnly - only bug fixes, stable
- Adoptable - waiting for new dev to give it a home
- Obsolete - replaced by vanilla feature, other plugin
- Discontinued - against current rules or dev reasons

DevSupportStateReason is a string to use to provide more detail for the state. For example, you can reference the replacing plugin or describe why a plugin was discontinued.

Notice can be used to show extra warnings for a plugin. This can be used for cases where we don't want to ban a version but help to try to avoid some commmon support issue.

**Incompatible Plugins**
image

**Notice**
image

**Related PRs**
https://github.com/goatcorp/DalamudAssets/pull/50
https://github.com/goatcorp/DalamudPackager/pull/6

### DIFF
--- a/Dalamud.sln.DotSettings
+++ b/Dalamud.sln.DotSettings
@@ -55,6 +55,7 @@
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=lumina/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Materia/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=PLUGINM/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=pluginmetadata/@EntryIndexedValue">True</s:Boolean>
 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Refilter/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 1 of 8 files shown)


goaaats:
This is a good feature, but I feel like it doesn't always need to be in the installer as in the end, it's inconsequential/visual clutter to users - if the plugin is in the installer, it should be good and usable.
Should there be a toggle? Only in dev mode?

kalilistic:
Sure, I'd say there's two areas of display right now.
1) Message on impacted plugins in existing views
2) Separate dedicated view for incompatible plugins.

Could add toggle(s) for these. Were you thinking #1 or #2 would be off by default?

kalilistic:
Or a second thought, moving the separate incompatible view outside of the plugin installer.

kalilistic:
The merge conflicts are kind of a pain on this one. Given the feedback, I think I'll re-implement this instead as part of the bigger plugin installer refresh. The side window can then perhaps be a separate window.

PR #841, opened 2022-05-14
Better IMEWindow

Loskh:
Moving the IME position to current cursor position.
ffxiv_dx11_g3qSTTcniv

### DIFF
--- a/Dalamud/Game/Gui/Internal/DalamudIME.cs
+++ b/Dalamud/Game/Gui/Internal/DalamudIME.cs
@@ -1,8 +1,12 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Numerics;
 using System.Runtime.InteropServices;
 using System.Text;
 
+using Dalamud.Hooking;
 using Dalamud.Interface.Internal;
 using Dalamud.Logging.Internal;
 using ImGuiNET;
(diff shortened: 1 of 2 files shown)


Dalamud/Game/Gui/Internal/DalamudIME.cs
```diff
@@ -73,6 +91,31 @@ internal void Enable()
                 this.interfaceHandle = Service<InterfaceManager>.Get().WindowHandlePtr;
                 this.wndProcPtr = Marshal.GetFunctionPointerForDelegate(this.wndProcDelegate);
                 this.oldWndProcPtr = SetWindowLongPtrW(this.interfaceHandle, WindowLongType.WndProc, this.wndProcPtr);
+
+                var module = Process.GetCurrentProcess().Modules.Cast<ProcessModule>().First(m => m.ModuleName == "cimgui.dll");
```
goaaats, line 95: Did you consider sending a patch to expose this to upstream ImGui? I'm sure he'd be happy to merge/implement that, considering it's essential for a good IME implementation.

Not that I don't wanna merge it like this, but I'd be happier knowing that there was gonna be an official way to do it at some point.

Loskh: okey, I'll try.

Loskh: I found that **ImGui::GetCurrentContext()->PlatformImePos** is the pos of the cursor. But  ImGui.Net did not generate the struct  **ImGuiContext**.

goaaats: @perchbird how annoying and terrible would it be to generate that? I think I need it in some other places too, would be cool to have

PR #821, opened 2022-04-26
Port XIVCommon context menu

kalilistic:
~~This breaks the existing context menu in it's current form so I marked as WIP. I hope to have time later this week to look at it so thought I'd share in the meantime in case anyone wanted to look.~~

I copied over the last version of the XivCommon (thanks Anna) context menu and set it's context menu class as the plugin service. I tested and it's currently able to add inventory and game object context menus. ~~It also supports sub menus for game objects (but not for inventory).~~
https://git.annaclemens.io/ascclemens/XivCommon/commit/2be0da1bda9d2ac6cdc01743e144fbc1c8f76a70

PR #846, opened 2022-05-16
feat: hide system menu buttons when using /xlui

Aireil:


Aireil:
Only includes Dalamud buttons, didn't do it for plugins context menus as #821 is still pending and I don't want kal to kill me if there's a conflict.

goaaats (approved):

Issue #853, opened 2022-05-25
Custom /Slash commands are not working for ANY plugins, including Dalamud


EmilyLove26:
None of the /slash commands are working as of the latest update:

### This includes commands from Dalamud itself as well as other Plugins

/xlsettings
/xlhelp
/xlplugins
/xldev
/xpiv - Visibility Plugin
/tweaks
/target <PC> - Simple Tweeks - Fix /Target Command
/pmb - Market Board Plugin

The list is just a sample of what doesn't work, as far as I know, none of the custom /slash commands are working

EmilyLove26:
Seems to be a problem with InnerSpace, not Dalamud. Closing.

EmilyLove26:
A pending pull request

LaxLacks:
I heard about this issue alongside a report that Sonar was also producing Japanese output. This occurs when `DalamudStartupInfo` is not initialized and contains default values, which means `DalamudStartupInfo.Language` is `Japanese` -- the commands are working, they are just in Japanese instead of the expected English in this case.

The referenced commit allows the use of an environment variable `DALAMUD_STARTUP_INFO` to populate the data. This means any injector can load `Dalamud.Boot.dll` without explicitly calling `Initialize`. It will initialize itself if provided.

goaaats:
Closing this because it's user error, not an actual issue

PR #879, opened 2022-06-06
Dalamud.Boot can auto-Initialize given DALAMUD_STARTUP_INFO environment variable

LaxLacks:
- fixed memory leak from using _wcsdup (without a subsequent free) to assign to std::wstring
- Dalamud.Boot checks for DALAMUD_STARTUP_INFO environment variable, and automatically initializes Dalamud, as an alternative to requiring an explicit call to Initialize() with the data

### DIFF
--- a/Dalamud.Boot/dllmain.cpp
+++ b/Dalamud.Boot/dllmain.cpp
@@ -61,8 +61,8 @@ DllExport DWORD WINAPI Initialize(LPVOID lpParam)
     GetModuleFileNameW(g_hModule, _module_path, sizeof _module_path / 2);
     std::filesystem::path fs_module_path(_module_path);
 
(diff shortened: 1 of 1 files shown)


EmilyLove26:
This fixes #853, btw

LaxLacks:
The issue #853 occurs because of compatibility issues with recent changes to the Dalamud injector system. This pull request helps resolve that by simplifying the injection process. Thanks for considering.

iCry1337:
Push!

NotNite:
Is this important to be merged soon such that it required a bump? I'm not familiar with Dalamud.Boot, so I'm likely missing something, but I can't infer what benefit this provides besides not having to type out args for injector.

PR #880, opened 2022-06-10
fix: Change GamepadState interface version for consistency

fitzchivalrik:
All other `PluginInterface`s are using a two-numbers version.


goaaats:
This is good, but if we do it now, we'll break every plugin that gets it. Let's hold it for a new API level!

goaaats (approved):

PR #881, opened 2022-06-12
FileDialog extensions

Ottermandias:
- Some cleanup with TextUnformatted instead of Text
- Use FontAwesomeIcon instead of chars
- Customization point for window flags (e.g. for no docking | no collapsing)
- Customization point for setting up quick access folders.

### DIFF
--- a/Dalamud/Interface/ImGuiFileDialog/ImGuiFileDialogFlags.cs
+++ b/Dalamud/Interface/ImGuiFileDialog/ImGuiFileDialogFlags.cs
@@ -16,41 +16,41 @@ namespace Dalamud.Interface.ImGuiFileDialog
         /// <summary>
         /// Confirm the selection when choosing a file which already exists.
         /// </summary>
-        ConfirmOverwrite = 1,
+        ConfirmOverwrite = 0x01,
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


Dalamud/Interface/ImGuiFileDialog/ImGuiFileDialogFlags.cs
```diff
@@ -16,41 +16,41 @@ public enum ImGuiFileDialogFlags
         /// <summary>
         /// Confirm the selection when choosing a file which already exists.
         /// </summary>
-        ConfirmOverwrite = 1,
```
goaaats, line 19: Are these breaking changes?

goaaats (changes requested):

Dalamud/Interface/ImGuiFileDialog/ImGuiFileDialogFlags.cs
Ottermandias: Technically, yes. But a) as far as I know nobody uses those flags at all yet (the manager handles them for you anyway), and b) if they are, they're absolutely not working as expected, in that you can't actually use them as flags and any try to do so will result in very weird consequences.
But yeah, breaking - but they have to be.

PR #890, opened 2022-06-21
Increase timeout for game window

marzent:
This raises the timeout to find the game window to effectively 21s since on a cold Wine start the window can take more than 6s to spawn (sometimes drastically longer depending on log level)


goaaats:
Figured we'd have to tune this, thanks.

PR #893, opened 2022-06-23
Load services asynchronously whenever possible

Soreepeong:
Fire all services to construct, but block when one load requires something else to be loaded. 

## Startup time taken
Time measured from pressing enter in command line with `Dalamud.Injector.exe launch -f` to seeing SQUARE ENIX logo.
* Without Dalamud: 3.452s
* With this PR: 5.400s
* Current net5: 7.688s


Dalamud/ServiceManager.cs
```diff
@@ -0,0 +1,52 @@
...
+                    null,
+                    null,
+                    null);
+            }
+        }
+
+        /// <summary>
+        /// Tag class to identify constructors to be called from Dalamud core.
+        /// </summary>
+        public class Tag
```
goaaats, line 42: Is there a reason this isn't an attribute with `AttributeTargets.Constructor`?

goaaats (commented):
This generally is a good idea, but I don't really like that it basically forces you to get what you would have passed in as a ctor arg as a service instead. 
Maybe as a QoL feature, ctor args could be injected if satisfied? We have code for this in Service<T> already, I believe.

Dalamud/ServiceManager.cs
Soreepeong: Parameterless constructors may be mistakenly called for whatever reason, but this marks that you aren't supposed to construct something manually using that service initializing constructor.

Soreepeong:
```cs
[ServiceAttribute]
private DalamudConfiguration Configuration { get; set; } = null!;

private SomeServiceCtor(ServiceManager.Tag tag, PluginManager pm) { ... }
```

DI now works for service object construction.

Soreepeong:
* Added `--msgbox3` that shows a msgbox from C# code, for attaching debugger.
* Added `--no-wait` to prevent Dalamud.Injector from force quitting the game while working with debugger.
* Converted service markers to be attributes instead.
* Added `BlockingEarlyLoadedService`, which will block game main thread until startup plugins have been loaded, if EntryPoint mode is used and a configuration option is turned on.

Dalamud/Game/ClientState/Conditions/Condition.cs
```diff
@@ -11,21 +11,21 @@ namespace Dalamud.Game.ClientState.Conditions
...
     [InterfaceVersion("1.0")]
+    [ServiceManager.BlockingEarlyLoadedService]
     public sealed partial class Condition
     {
         /// <summary>
         /// The current max number of conditions. You can get this just by looking at the condition sheet and how many rows it has.
         /// </summary>
         public const int MaxConditionEntries = 100;
 
+        [ServiceManager.ServiceDependency]
```
goaaats, line 22: Apparently leftover

Dalamud/IoC/Internal/ObjectInstance.cs
```diff
@@ -11,11 +12,11 @@ internal class ObjectInstance
...
         /// </summary>
-        /// <param name="instance">The underlying instance.</param>
-        public ObjectInstance(object instance)
+        /// <param name="instanceTask">The underlying instance.</param>
+        public ObjectInstance(Task<WeakReference> instanceTask)
         {
-            this.Instance = new WeakReference(instance);
-            this.Version = instance.GetType().GetCustomAttribute<InterfaceVersionAttribute>();
+            this.InstanceTask = instanceTask;
+            this.Version = instanceTask.GetType().GetCustomAttribute<InterfaceVersionAttribute>();
```
goaaats, line 19: Would this not get the attribute on Task<T>?

goaaats (changes requested):

PR #903, opened 2022-06-28
Improvements

Soreepeong:
* Added Clear button next to search input text box in Plugin Installer.
* Added an option to make Serilog log synchronously without any additional buffering else than operating system level buffers.
* Added an xivfix that will prevent the game from ending up corrupting configuration data when it crashes while saving.
    * This will make xiv write to .new files when it's trying to write to .cfg or .dat files, and then rename those .new files into original filenames after the game is done writing.
* Made Plugin Installer download up to 8 images concurrently, but only for visible images. Images for the items that aren't visible won't be downloaded until scrolled.
* Made TitleScreenMenu order stable.
* Made Dalamud Unload reliably work, and made it quicker, by forcing all Service{T}.Dispose to run in Framework.Tick if invoked manually, or during Framework.Destroy if invoked by game quit.
* Made hook create/dispose/enable/disable run in synchronized blocks.
* Made Dalamud wait 0.5 seconds by default for plugins after Plugin.Dispose before Loader.Dispose. Wait time is configurable via Experimental tab.
    * If a plugin results in a crash when this value is set to zero, that plugin needs to be fixed to properly cancel and wait for all Tasks that plugin has fired.
* Ensured plugin to load in the context of Framework.Tick if synchronous load is requested, except when plugin's load mode is 2 (anytime).

### DIFF
--- a/Dalamud/Interface/Internal/InterfaceManager.cs
+++ b/Dalamud/Interface/Internal/InterfaceManager.cs
@@ -57,6 +57,9 @@ namespace Dalamud.Interface.Internal
         private readonly HashSet<SpecialGlyphRequest> glyphRequests = new();
         private readonly Dictionary<ImFontPtr, TargetFontModification> loadedFontInfo = new();
 
+        [ServiceManager.ServiceDependency]
+        private readonly Framework framework = Service<Framework>.Get();
+
         private readonly ManualResetEvent fontBuildSignal;
         private readonly SwapChainVtableResolver address;
         private readonly Hook<DispatchMessageWDelegate> dispatchMessageWHook;
@@ -73,12 +76,12 @@ namespace Dalamud.Interface.Internal
         private bool isOverrideGameCursor = true;
 
         [ServiceManager.ServiceConstructor]
-        private InterfaceManager(SigScanner sigScanner)
+        private InterfaceManager()
         {
             this.dispatchMessageWHook = Hook<DispatchMessageWDelegate>.FromImport(
-                Process.GetCurrentProcess().MainModule, "user32.dll", "DispatchMessageW", 0, this.DispatchMessageWDetour);
+                null, "user32.dll", "DispatchMessageW", 0, this.DispatchMessageWDetour);
             this.setCursorHook = Hook<SetCursorDelegate>.FromImport(
-                Process.GetCurrentProcess().MainModule, "user32.dll", "SetCursor", 0, this.SetCursorDetour);
+                null, "user32.dll", "SetCursor", 0, this.SetCursorDetour);
 
             this.fontBuildSignal = new ManualResetEvent(false);
 
@@ -90,12 +93,6 @@ namespace Dalamud.Interface.Internal
 
         [UnmanagedFunctionPointer(CallingConvention.ThisCall)]
         private delegate IntPtr ResizeBuffersDelegate(IntPtr swapChain, uint bufferCount, uint width, uint height, uint newFormat, uint swapChainFlags);
-
-        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
-        private delegate int CreateDXGIFactoryDelegate(Guid riid, out IntPtr ppFactory);
-
-        [UnmanagedFunctionPointer(CallingConvention.StdCall)]
-        private delegate int IDXGIFactory_CreateSwapChainDelegate(IntPtr pFactory, IntPtr pDevice, IntPtr pDesc, out IntPtr ppSwapChain);
 
         [UnmanagedFunctionPointer(CallingConvention.StdCall)]
         private delegate IntPtr SetCursorDelegate(IntPtr hCursor);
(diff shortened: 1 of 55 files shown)


Dalamud/Interface/Internal/InterfaceManager.cs
```diff
@@ -248,14 +245,7 @@ internal set
...
         {
-            // HACK: this is usually called on a separate thread from PresentDetour (likely on a dedicated render thread)
-            // and if we aren't already disabled, disposing of the scene and hook can frequently crash due to the hook
-            // being disposed of in this thread while it is actively in use in the render thread.
-            // This is a terrible way to prevent issues, but should basically always work to ensure that all outstanding
-            // calls to PresentDetour have finished (and Disable means no new ones will start), before we try to cleanup
-            // So... not great, but much better than constantly crashing on unload
-            this.Disable();
-            Thread.Sleep(500);
+            this.framework.RunOnFrameworkThread(this.Disable).Wait();
```
goaaats, line 248: setCursorHook is disabled in Disable(), yet we Dispose it immediately here... Dispose() automatically disables it, so aren't you double disposing?

Dalamud/Interface/Internal/Windows/PluginImageCache.cs
```diff
@@ -234,14 +287,43 @@ private void FrameworkOnUpdate(Framework framework)
 
         private async void DownloadTask()
         {
-            while (!this.downloadToken.Token.IsCancellationRequested)
+            var token = this.downloadToken.Token;
+            var runningTasks = new List<Task>();
+            var pendingFuncs = new List<Tuple<ulong, Func<Task>>>();
+            while (true)
```
goaaats, line 293: Maybe we want to sleep in this loop, to not use up CPU for something insignificant - last iteration didn't have one either though

Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
```diff
@@ -1686,7 +1701,7 @@ private void DrawPluginControlButton(LocalPlugin plugin)
                     {
                         Task.Run(() =>
                         {
-                            var unloadTask = Task.Run(() => plugin.Unload())
+                            var unloadTask = Task.Run(() => plugin.UnloadAsync())
```
goaaats, line 1704: Does UnloadAsync() need to be awaited in the task? Why wrap it in Task.Run()?

Dalamud/Interface/Internal/Windows/SettingsWindow.cs
```diff
@@ -561,6 +565,34 @@ private void DrawExperimentalTab()
             var configuration = Service<DalamudConfiguration>.Get();
             var pluginManager = Service<PluginManager>.Get();
 
+            var useCustomPluginWaitBeforeFree = this.pluginWaitBeforeFree.HasValue;
+            if (ImGui.Checkbox(
```
goaaats, line 569: I'm a little wary of having this customizable... It has no meaning to users, ideally we'd have a default value that always works. What's the actual problem here - why can wait time affect disposes? Back before the async workaround, we'd just call dispose synchronously on every plugin, and it worked fine.

Dalamud/Plugin/Internal/PluginManager.cs
```diff
@@ -207,6 +212,46 @@ public static bool IsManifestVisible(RemotePluginManifest manifest)
     /// <inheritdoc/>
     public void Dispose()
     {
+        if (!this.InstalledPlugins.Any())
```
goaaats, line 215: Why do you only dispose this if there are any plugins? It's applied every time - not that it matters, just wanna understant the thinking behind it

goaaats (changes requested):

Dalamud/Interface/Internal/Windows/PluginImageCache.cs
Soreepeong: 1. If there aren't any further pending tasks, TryTake will block.
2. If there are any, then it will block when there already are 8 tasks running.

This loop shouldn't be hogging CPU times, so sleep won't be necessary, but probably can limit concurrent decoding to the number of cores available, while having separate limit of 8 for image download.

Dalamud/Plugin/Internal/PluginManager.cs
Soreepeong: Same code exists also at the end of Dispose, but now that I think about it again, less duplicates would be more preferrable than less indents in this case. Will invert conditionals.

Dalamud/Interface/Internal/Windows/SettingsWindow.cs
Soreepeong: It also used to wait for quite a long time, and that usually gave Tasks ran by plugins to finish. Without this wait time, tasks would keep running even after successful plugin/loader.Dispose, and TE would crash the game when this is followed by Dalamud unload.

goaaats: Fair enough, I guess this is where implementing our own task scheduler would be useful, but that's a topic for another day

Dalamud/Interface/Internal/InterfaceManager.cs
Soreepeong: Double disable is acceptable via how Hook<...> is implemented, and it does Dispose them only once. Since we do not want game threads to have code running when we dispose the hook, we disable it from Framework.Tick, which is the time when the game shouldn't be running any of the hooked code. It is guaranteed that the game is the one to run our hook code for winapi hooks, as we're hooking by overwriting game's import table.

Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
goaaats: All good

Dalamud/Interface/Internal/InterfaceManager.cs
Soreepeong: Changed to make intentions clear.

goaaats (approved):

PR #907, opened 2022-07-03
adding pf condition "duty complete (weekly reward unclaimed)"

pot0to:
image


goaaats (changes requested):
Since this is a mask, I would recommend just using 1 << 0, 1 << 1, 1 << 2, etc.

It's how we do it elsewhere as well and looks a bit nicer.

goaaats (approved):

PR #917, opened 2022-07-13
Add VirtualKey -> ImGuiKey helper method to ImGuiHelpers

lmcintyre:


goaaats (approved):

lmcintyre:
im a quality dev

lmcintyre:
ok I fixed this because it was obviously failing build thanks to my web editor usage but also I fixed the modifiers being held after alt+tab and added a reverse map (imguikey to virtualkey). i didn't want to have to copy around multiple key maps, so both of the massive switch statements are in imguiscene which i can move if you want

PR #921, opened 2022-07-15
fix: wrong version displaying in the installer

Aireil:
Assembly version sometimes doesn't reflect the actual plugin version. The manifest should always be correct though, unless I'm mistaken.


goaaats:
I think I accidentally just did this myself, sorry, didn't even see your PR

PR #924, opened 2022-07-16
Change visibility for MarketBoardPurchase and MarketBoardPurchaseHandler Struct

Infiziert90:
Makes it easier to check what item an user just purchased from the marketboard, with details like price, without having to refresh the history again, reducing requests to the ff servers


goaaats (approved):

Aireil:
Fyi, this was PR'd on master (net6), not on net5.

PR #926, opened 2022-07-17
Call RebuildFonts on next Framework.Tick on NewFontRef

Soreepeong:


goaaats:
Should be safer in any case

PR #927, opened 2022-07-18
Rework plugin deletion + misc

Aireil:
Allows to always use the trash button as long as the plugin is disabled, but only schedules the deletion (except for dev plugins).
If the plugin has no error, it is simply removed from the plugin list.
If it does have an error, the plugin stays in the plugin list with a (scheduled for deletion) tag in the installer.

I've tried to limit big code changes, so it's easy to read and tested it with multiple scenarios, but if you see anything weird, don't hesitate to ask.


goaaats:
Looks good, ideally the "scheduling" status would be completely transparent to users. Fine for now though.

PR #943, opened 2022-07-30
Add Per-Plugin Log Filtering Support

KazWolfe:
This commit adds first-pass support for more robust plugin log filtering. No changes were made that affect PluginLog's APIs nor log format. This change works by making use of Serilog's `SourceContext` property to attach plugin information to all log messages. The in-game log UI can then be used to filter based on this property.

Future expansions for this system include the ability to set different plugins to different log levels (something that already can technically be done, but requires those plugins be hard-coded through `MinimumLevel.Override`), creating new root loggers for each plugin (thereby giving plugin devs more control over their logging should they want to use it), plus other potential improvements in the way of adding context or rich information.

- Update `PluginLog` to attach a `SourceContext` property to all log messages.
- Tweak the `SerilogEventSink` to pass the original log event around.
- Suppress Info/Debug/Verbose exceptions from Troubleshooting reports.
- Fix the `ConsoleWindow` log filter to use _all_ filters, rather than just one.
- Add `ConsoleWindow` dropdown to select plugins to filter logs by
- Add support for multiple log levels to `ConsoleWindow` filtering

### DIFF
--- a/Dalamud/EntryPoint.cs
+++ b/Dalamud/EntryPoint.cs
@@ -182,12 +182,17 @@ namespace Dalamud
             }
         }
 
-        private static void SerilogOnLogLine(object? sender, (string Line, LogEventLevel Level, DateTimeOffset TimeStamp, Exception? Exception) e)
+        private static void SerilogOnLogLine(object? sender, (string Line, LogEvent LogEvent) ev)
         {
-            if (e.Exception == null)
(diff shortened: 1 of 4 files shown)


goaaats:
Sorry, could you rebase this to master? Shouldn't be a huge deal

goaaats (approved):

PR #953, opened 2022-08-03
Installer spacing + reduce search bar length

Aireil:
Search bar was way bigger than needed, which is weird at smallest/default size, overlapping the text.


Aireil:
Added fixes for two overlay issues as well. Will probably merge conflict with another PR, but that's not my fault, too slow!! 🏃

goaaats (approved):

PR #976, opened 2022-08-24
Remove "Reload All Plugins" Command

KazWolfe:
RIP AND TEAR.

- Remove the "Reload All Plugins" MenuItem from `/xldev`
- Remove the `/xldreloadplugins` Command
- Mark `PluginManager#ReloadAllPluginsAsync()` as obsolete for removal in the future

### DIFF
--- a/Dalamud/Plugin/Internal/PluginManager.cs
+++ b/Dalamud/Plugin/Internal/PluginManager.cs
@@ -535,6 +535,7 @@ internal partial class PluginManager : IDisposable, IServiceType
     /// Reload all loaded plugins.
     /// </summary>
     /// <returns>A task.</returns>
+    [Obsolete("This method should no longer be used and will be removed in a future release.")]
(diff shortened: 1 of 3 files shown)


Dalamud/Plugin/Internal/PluginManager.cs
```diff
@@ -535,6 +535,7 @@ async Task LoadPluginsAsync(string logPrefix, IEnumerable<PluginDef> pluginDefsL
     /// Reload all loaded plugins.
     /// </summary>
     /// <returns>A task.</returns>
+    [Obsolete("This method should no longer be used and will be removed in a future release.")]
```
KazWolfe, line 538: Maybe worth making this throw a compile-time error just to _really_ drive it home that people should not be touching this? Won't break existing plugins but it will stop new builds from working until people fix things.

goaaats (approved):

PR #988, opened 2022-08-28
fix: allow control if orphaned is loaded

Aireil:


goaaats:
It shouldn't be possible to load orphaned plugins, is deleting them not possible at the moment?

Aireil:
If you remove a repo during gameplay, you can have a loaded orphaned, but you can't disable it, which also means you cannot delete it.

goaaats:
Deprecated, thanks

PR #994, opened 2022-09-01
Add internal exception handling for event delegates

rootdarkarchon:
This PR adds internal exception handling for `Invoke()` by replacing it with the `Raise()` extension method.
The idea is to continue plugin execution in case of a bad plugin causing an Exception in a subscribed event.

Raise is iterating over all event subscribers from the invocation list and in case of failure, will log an Exception, preventing it from bubbling upwards and continuing with the next subscriber, instead of halting after one Exception.

This implementation is per se incomplete and does not cover all existing event delegates in Dalamud. The principle to implement those would be the same, however I only covered the low hanging fruits and "most important" events.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -9,6 +9,7 @@ using Dalamud.Game.Network.Internal;
 using Dalamud.Hooking;
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
+using Dalamud.Utility;
 using FFXIVClientStructs.FFXIV.Client.Game;
 using Serilog;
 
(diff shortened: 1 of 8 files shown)


philpax:
Had a quick look over and it looks good (and I'm a big fan of the change - an exception in one plugin shouldn't bring down everything else). Only thing I'd suggest is using a different name; `Raise` makes me think that it's going to throw unconditionally. Maybe `InvokeSafely` or some variant thereof?

philpax (approved):

rootdarkarchon:
Adjusted naming to InvokeSafely, you are correct it makes more sense

philpax (approved):

goaaats (approved):

goaaats:
Logging could probably be better but this is good for now, thanks

PR #1015, opened 2022-09-19
fix RunOnFrameworkThread/RunOnNextTick race condition

Soreepeong:


KazWolfe:
Sorry, but would it be possible to have a maintainer take a look at this? I'm getting a few users able to reproduce this problem with some regularity, albeit it's probably my fault for doing multi-threaded shenanigans again.

goaaats (approved):

PR #1025, opened 2022-10-11
Add SeString.Parse(byte* ) method

MidoriKami:
Add a method to generate a SeString from a byte* of unknown length by searching for and counting the number of characters until the null terminator.


goaaats:
What about the ones in MemoryHelper? Am I missing something?

MidoriKami:
I wasn't aware of `MemoryHelper.ReadSeStringNullTerminated(new IntPtr(ptr))` until Caraxi pointed it out after this was submitted.

goaaats:
Will close then, thanks though

PR #1026, opened 2022-10-11
fix: make sure style is popped on installer exceptions

Aireil:
It's for the famous concurrency issue, proper fix will wait @kalilistic's rework of the installer. It doesn't really cause any visible issues other than breaking the style.


goaaats:
We should have support for that in the window class

Aireil:
True, feel free!

kalilistic:
rip

PR #1032, opened 2022-11-08
make Framework updateHook and destroyHook nullable

marzent:
This fixes 
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Dalamud.Game.Framework.HandleFrameworkUpdate(IntPtr framework) in C:\goatsoft\companysecrets\dalamud\Game\Framework.cs:line 456
```
which seemed to happen semi-reliably for some wine users.

### DIFF
--- a/Dalamud/Game/Framework.cs
+++ b/Dalamud/Game/Framework.cs
@@ -30,8 +30,8 @@ public sealed class Framework : IDisposable, IServiceType
 
     private readonly Stopwatch updateStopwatch = new();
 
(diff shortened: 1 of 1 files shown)


goaaats:
Why does this happen, I wonder?

KazWolfe:
Is this PR still valid?

marzent:
I haven't seen this in a long time, although I am also not certain what the root cause of the issue was and under what circumstances the hooks could be `null` there...
Maybe this resolved itself with the reworked service dependency initialization.

PR #1033, opened 2022-11-14
Add IsDead boolean property to Character

SheepGoMeh:
`EventState == 2` is used to check that a character is dead.
While functionally similar to checking whether the target's HP is zero, this is the correct way to check it based on how it is handled in the client (refer to `Client::Game::Character::Character_IsDead`).


Caraxi:
Would be far better to call the virtual function on the object

SheepGoMeh:
> Would be far better to call the virtual function on the object
> 
> ```
> this.Struct->GameObject.IsDead()
> ```

Would it not be better to move this to GameObject in that case?

Caraxi:
> 
> Would it not be better to move this to GameObject in that case?

Yes, I'm just going from where you already had it

goaaats (approved):

PR #1045, opened 2022-12-14
Indicate major dalamud config in crash tspack filename

Soreepeong:
Crash file names will have formats of `crash[3][D][P][L]-YYYYYMMDDHHMMSS.tspack`.
* `3`: At least one third party plugin is active.
* `D`: Dalamud testing is enabled.
* `P`: Plugin test is enabled.
* `L`: "Load All API Levels" is enabled.

### DIFF
--- a/DalamudCrashHandler/DalamudCrashHandler.cpp
+++ b/DalamudCrashHandler/DalamudCrashHandler.cpp
@@ -1,4 +1,5 @@
 #include <array>
+#include <cassert>
 #include <chrono>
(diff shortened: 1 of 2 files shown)


goaaats:
I honestly don't think that we should be doing this kinda stuff in dalamud, it defeats the point of having the packs at all. There's no reason

KazWolfe:
Targeting net7 branch, I think it's safe to say this PR has gone stale.

If this is still desired at a later time, we can re-open.

PR #1046, opened 2022-12-20
Fix runtime_present for ReShade 5/GShade 4

NotNite:
they call me the less-than-10-lines-diff contributor


goaaats:
You should try-catch the product info code and have it fall back to the original signature to preserve the behavior we've had up until now

NotNite:
Oh, didn't occur to me that could throw an exception. Will try-catch and add a log in a second.

goaaats (approved):

PR #1052, opened 2023-01-07
fix: Fix broken case for relative symlinks

KazWolfe:
This pull request aims to fix a bug that occurs when trying to save a file that is a relative symbolic link.

The existing behavior of `read_symlink` does not resolve canonical paths for targets, so a returned target may be `..\path\to\target.DAT`, which cannot be used to create a file handle later. This will cause the file handle to fail to open and (as a result) cause file writes to fail with the normal cryptic in-game "could not be saved" error message.

Instead of resolving symbolic links in this manner, using [`weakly_canonical()`](https://en.cppreference.com/w/cpp/filesystem/canonical) allows full canonical file paths to be resolved, making a guarantee that the output path will have no symbolic links. This should additionally simplify the logic loop and error handling in this case.

`weakly_canonical()` will succeed even for nonexistent files, similar to the behavior of breaking on `ec`, so this change should not have any negative impact. If, however, an error code is thrown, the game will fall back to native file save methods.

This was tested locally and _seemed_ to behave, but I'd like to propose this change proper (especially @Soreepeong) to verify that this won't have any side effects or other issues.

### DIFF
--- a/Dalamud.Boot/xivfixes.cpp
+++ b/Dalamud.Boot/xivfixes.cpp
@@ -437,13 +437,13 @@ void xivfixes::backup_userdata_save(bool bApply) {
             if (ext != ".dat" && ext != ".cfg")
                 return s_hookCreateFileW->call_original(lpFileName, dwDesiredAccess, dwShareMode, lpSecurityAttributes, dwCreationDisposition, dwFlagsAndAttributes, hTemplateFile);
 
-            for (auto i = 0; i < 16; i++) {
-                std::error_code ec;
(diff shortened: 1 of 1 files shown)


KazWolfe:
Marking as ready for review. Works on my machine, probably, so I'll let goat et al tear it apart.

goaaats:
I think this is a bit dangerous to merge just before 6.3 because save files, but we'll have it a week or two after the rush for the patch, just so we can isolate potential issues.

goaaats (approved):

PR #1054, opened 2023-01-09
Change GameGui.GetAddonByName default parameter

MidoriKami:
Changes 
```cs
public unsafe IntPtr GetAddonByName(string name, int index)
```

to
```cs
public unsafe IntPtr GetAddonByName(string name, int index = 1)
```

To reflect most common usage. Very few addons need anything other than a 1 index by default. This also reflects the ClientStructs implementation as it also has a default value of 1.


MidoriKami:
Targeted wrong branch. Will fix in a moment.

PR #1060, opened 2023-01-10
Improve WorldToScreen-ing points in front but off to the side

0x0ade:
Would've loved to show an example, but I didn't get any pics or gifs before the patch hit, so...

This PR splits the logic statement at the end of `WorldToScreen` apart into two, allowing for the user of that method to only care if something's "in front". This is important if you want to f.e. draw a line that goes from a world point that's on-screen to another one that's still in front of the camera, but off to the side of the screen. The previous (current) "screen culling" behavior of that method makes it impossible.

I've tested this change myself for months now using a plugin that's still in-dev and a stupid reflection hack to get the matrix singleton myself, with a slightly different signature:

```cs
public void WorldToScreen(Vector3 worldPos, out Vector2 screenPos, out bool inView, out bool inFront)
```

```cs
inFront = pCoords.Z > 0;
inView = inFront &&
    screenPos.X > windowPos.X && screenPos.X < windowPos.X + width &&
    screenPos.Y > windowPos.Y && screenPos.Y < windowPos.Y + height;
```

```cs
Service.GameGuiExt.WorldToScreen(target.Position, out Vector2 to, out bool isToInView, out bool isToInFront);

if ((Service.Config.CullAggressively && !isToInView) || !isToInFront)
{
    continue;
}
```

I've decided to keep the old behavior in tact with the existing overload and adjusted the documentation to match the existing behavior accordingly, to try to avoid surprising anyone else who used that method and relied on the "screen culling" behavior. Using the overload with the additional out param and even just discarding it works fine for getting offscreen-but-infront points.

**TL;DR:** in front != in view, ~~valve~~ pls fix

### DIFF
--- a/Dalamud/Game/Gui/GameGui.cs
+++ b/Dalamud/Game/Gui/GameGui.cs
@@ -184,8 +184,18 @@ public sealed unsafe class GameGui : IDisposable, IServiceType
     /// </summary>
     /// <param name="worldPos">Coordinates in the world.</param>
     /// <param name="screenPos">Converted coordinates.</param>
-    /// <returns>True if worldPos corresponds to a position in front of the camera.</returns>
+    /// <returns>True if worldPos corresponds to a position in front of the camera and screenPos is in the viewport.</returns>
     public bool WorldToScreen(Vector3 worldPos, out Vector2 screenPos)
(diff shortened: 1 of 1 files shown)


0x0ade:
Force-pushed as I rebased against latest master.

goaaats (approved):

PR #1069, opened 2023-01-12
Improve IconButton

MidoriKami:
As per the discussion in the dalamud-dev discord channel, we believe the cause of the caravan issue is the missing `##` tags to indicate the string has an ID component.

Additionally a couple overloads that allow for strings to be the buttons id have been added for ease of use.


MidoriKami:
Please hold... working on git issues...

PR #1078, opened 2023-01-14
Add Instance number into ClientState

rreminy:
The address pointed by the signature is where the instance number is stored at, as read by the `/instance` command.

This instance number corresponds to the current instance you're at (1, 2 or 3). It'll be zero for non-instanced zones.

By itself its a byte but there are places that read this value as 32-bit

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -9,6 +9,7 @@ using Dalamud.Game.Network.Internal;
 using Dalamud.Hooking;
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
+using Dalamud.Memory;
 using Dalamud.Utility;
 using FFXIVClientStructs.FFXIV.Client.Game;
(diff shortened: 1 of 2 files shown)


MidoriKami:
Wouldn't this be more appropriate to be in ClientStructs?

rreminy:
Wherever its more appropriate. I kind of associate this with TerritoryType so I placed it next to it.

I have not worked with ClientStructs yet

goaaats:
Is this not directly in CS? I try to avoid adding sigs directly into Dalamud nowadays, since they can't benefit from their infrastructure.

Caraxi:
It looks like its part of UiState. If its not already in ClientStructs I'll add it now

Caraxi:
As of https://github.com/aers/FFXIVClientStructs/commit/4aa2ee2bdbdeba578d0010dee4fa178b77023ae3 
```
UIState.Instance()->AreaInstance.Instance
```

rreminy:
alright

PR #1086, opened 2023-01-15
Update ClientStructs

MidoriKami:
It seems the github action for updating client structs broke.

git said that the ClientStructs commit was "detached from head"
so I checked out "main" and pulled clientstructs

I don't know if this is correct, I have never tried to do this before.
It seems correct, but please double check 😱 

The updates to client structs will let me remove the last sig from wotsit, and remove a sig from WAIA


goaaats:
I usually do this myself, it's easier, but thanks anyhow.

Issue #1087, opened 2023-01-16
Plugin installer: very rare InvalidOperationException


Limiana:
Once in a while, there will be this error when working with plugin installer window:
```
2023-01-16 09:34:01.610 +03:00 [ERR] [Dalamud] Error during OnDraw
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at Dalamud.Interface.Internal.Windows.PluginInstaller.PluginInstallerWindow.DrawAvailablePluginList() in C:\goatsoft\companysecrets\dalamud\Interface\Internal\Windows\PluginInstaller\PluginInstallerWindow.cs:line 881
   at Dalamud.Interface.Internal.Windows.PluginInstaller.PluginInstallerWindow.DrawPluginCategoryContent() in C:\goatsoft\companysecrets\dalamud\Interface\Internal\Windows\PluginInstaller\PluginInstallerWindow.cs:line 1094
   at Dalamud.Interface.Internal.Windows.PluginInstaller.PluginInstallerWindow.DrawPluginCategories() in C:\goatsoft\companysecrets\dalamud\Interface\Internal\Windows\PluginInstaller\PluginInstallerWindow.cs:line 1013
   at Dalamud.Interface.Internal.Windows.PluginInstaller.PluginInstallerWindow.Draw() in C:\goatsoft\companysecrets\dalamud\Interface\Internal\Windows\PluginInstaller\PluginInstallerWindow.cs:line 224
   at Dalamud.Interface.Windowing.Window.DrawInternal() in C:\goatsoft\companysecrets\dalamud\Interface\Windowing\Window.cs:line 274
   at Dalamud.Interface.Windowing.WindowSystem.Draw() in C:\goatsoft\companysecrets\dalamud\Interface\Windowing\WindowSystem.cs:line 144
   at Dalamud.Interface.Internal.DalamudInterface.OnDraw() in C:\goatsoft\companysecrets\dalamud\Interface\Internal\DalamudInterface.cs:line 455
```
The error has been happening since a long while but it is rare and I can't reliably reproduce it too. I've been rarely seeing it in some other people's log. I'm personally most likely to encounter it when enabling/disabling dev plugins.
It also causes style var leaking:
[...]

PR #1090, opened 2023-01-18
fix: don't hide the plugin list when operations are in progress

Aireil:
Fix the scrolling breaking when installing/enabling/disabling a plugin. I don't think this was necessary, as the installer keeps copies. This pretty much reverts the change made a while ago attempting to fix concurrency issues that still exist #1087.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
@@ -1091,7 +1091,7 @@ internal class PluginInstallerWindow : Window, IDisposable
 
     private void DrawPluginCategoryContent()
     {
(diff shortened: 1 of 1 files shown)


goaaats:
If this doesn't work, we'll have to be smarter about locks or copying, soon mayhaps

goaaats (approved):

Aireil:
What do you mean if this doesn't work? This only restores the old behavior to avoid the bump up in the installer, but doesn't try to fix the concurrency issue at all.
For the concurrency issue, after looking at it again, I feel like it's these lists that are getting modified asynchronously when PluginManager sends a refresh update, but didn't confirm it 🔍: https://github.com/goatcorp/Dalamud/blob/379a4422a2195ec57d73aefa14f088ffec884bf8/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs#L88-L90

PR #1104, opened 2023-01-29
Add enable and disable commands

Infiziert90:
Adds two simple chat commands, /xlenable and /xldisable.
The commands take the full plugin name as argument and try to enable/disable said plugin.

The user will be informed with a chat message if the load/unload was successful.

### DIFF
--- a/Dalamud/Interface/Internal/DalamudCommands.cs
+++ b/Dalamud/Interface/Internal/DalamudCommands.cs
@@ -23,6 +23,10 @@ namespace Dalamud.Interface.Internal;
 [ServiceManager.EarlyLoadedService]
 internal class DalamudCommands : IServiceType
 {
+    // in milliseconds
+    private const int CooldownTime = 10000;
+    private Dictionary<string, long> cooldown = new();
+
     [ServiceManager.ServiceConstructor]
     private DalamudCommands(CommandManager commandManager)
     {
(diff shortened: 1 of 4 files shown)


goaaats:
You need to make sure that you cannot load a plugin multiple times, and that you cannot enable/disable a plugin while it's still being enabled/disabled. I think an additional cooldown on top of this would also be well-advised.

A lot of plugins take quite a bit of time for this, nowadays. Will cause bugs.

Infiziert90:
> You need to make sure that you cannot load a plugin multiple times, and that you cannot enable/disable a plugin while it's still being enabled/disabled. 

I think i already do that, or is there more cases that should be checked / aren't checked by LoadAsync / UnloadAsync?

Infiziert90:
Added a simple cooldown to both commands, using TickCount64.
Cooldown is currently set to 10s, but can be adjusted by changing the const value.

goaaats:
LoadAsync/UnloadAsync don't actually check if the plugin is already loading or unloading, so you can't rely on them blocking. Probably have to track it separately and also block operations on that plugin in the installer, ideally.

Infiziert90:
> LoadAsync/UnloadAsync don't actually check if the plugin is already loading or unloading, so you can't rely on them blocking. Probably have to track it separately and also block operations on that plugin in the installer, ideally.

Both commands share the same cooldown now, this will prevent unload and load at the same time.

The commands now also check if the plugin installer is currently busy with something.
After this check they set the installer to "InProgress" themself, this should prevent the installer from unload and load.

PR #1110, opened 2023-02-11
Minor Logging QoL

KazWolfe:
This pull request aims to add the following:

- A new `LogRaw()` method in `PluginLog` will allow plugin developers to specify log levels at runtime, allowing easier interoperability with other logging systems.
- Update all `Exception` arguments to be nullable, suppressing compiler warnings in cases where an exception may or may not be present.

If necessary, I have no qualms about dropping the Exception nullability change (will it affect API?).


goaaats:
Not sure if this breaks the API; don't think so. We'll see

goaaats (approved):

PR #1111, opened 2023-02-11
Changes to Dalamud Boot DLL so that it works in WINE

Soreepeong:
1. ~~Instead of traversing through allocated memory regions, which tends to be flakey under WINE because it reports non-pagesize-aligned allocation size and makes traverse routing loop infinitely, we now use RCX obtained from `GetThreadContext(process_information.hThread, &ctx)` which should contain the address to game entry point. If some other tool decides to inject it this way then it won't work, so if it turns out to be the case that happens often, we may have to support both ways for injecting in entrypoint mode.~~ Just traverse memory region, but forcing it to advance by round-by page size.
2. Minimized handwritten assembly code, and while doing that, fixed stack alignment problems that happens under WINE.

### DIFF
--- a/Dalamud.Boot/Dalamud.Boot.vcxproj
+++ b/Dalamud.Boot/Dalamud.Boot.vcxproj
@@ -32,6 +32,9 @@
     <IntDir>obj\$(Configuration)\</IntDir>
   </PropertyGroup>
   <Import Project="$(VCTargetsPath)\Microsoft.Cpp.props" />
+  <ImportGroup Label="ExtensionSettings">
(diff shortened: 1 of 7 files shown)


goaaats:
Is this good to go?

Soreepeong:
Just in case the answer was lost in a huge amount of scrolling up in random places, the answer is yes.

It seems like it consistently causes the game to not start up **only for the first time after update** but subsequent startups would always work.

KazWolfe:
Windows fakeboot tested ok (win11 pro, release branch), succeeded on first launch. Live boot ok on Steam Deck with stock config.

Yoloing it because maint tonight.

KazWolfe (approved):

Issue #721, opened 2021-12-25
PrintChat breaks context menu items


Aireil:
Using this event:
`public delegate void OnMessageDelegate(XivChatType type, uint senderId, ref SeString sender, ref SeString message, ref bool isHandled);`
Just resending the message like this:
`Service.Chat.PrintChat(new XivChatEntry { Message = message, Name = sender, Type = type, SenderId = senderId });`

The message from Dalamud is missing some context menu items for some reasons.
First is normal chat, second is PrintChat.
https://streamable.com/8w9cfn

PunishedPineapple:
This is because the field/parameter that's sender ID right now is actually a unix timestamp (unless it's zero, then it seems to set time automatically).  I had an issue with setting sender on my messages as a way to track them in the `ChatMessage` event where all my chat messages had the same timestamp.  Setting it to any unix timestamp gives the appropriate time in the log.  Idk if there's more going on.  My guess is that the `parameter` field actually contains *something* that identifies the sender, but I haven't figured it out yet.

PunishedPineapple:
Digging into this and REing a bit with one of Anna's sigs too, some of the menu items are from the sender field being a link string (Player + Text + Link Terminator payloads), but other menu options are from the content ID that gets stored when the message comes in.

I can get the all of the context menu items working on copied and re-printed messages by writing in the correct content ID to the log module's store, but I don't know whether this is something that would get accepted into Dalamud even if I find a function that does it instead of writing memory directly.  I could see this being potentially dangerous if you were to stick in an invalid content ID.

Anyway, I just don't have enough knowledge to decide whether it's wise to pursue this further or not.

PunishedPineapple:
Sig for that function was `4C 8B D1 48 8B 89 ?? ?? ?? ?? 48 85 C9` so I don't have to dig it up again.

PR #1116, opened 2023-02-14
ChatGui2

PunishedPineapple:
### Description:
Implements most of the proposed [DIP 39](https://github.com/goatcorp/DIPs/pull/39):
- Fills out `XivChatType` with all known channels and adds all known source/target masks.  Until the next API level change, this is done as `XivChatType2` in order to avoid breakage of naive plugins.
- Creates a `ChatGui2` class/service that implements all of the old `ChatGui` functionality, with knowledge and handling of chat message sources, chat message printed callback, and corrected `senderID` to `timestamp`.
- Changes `ChatGui` to be a simple wrapper for `ChatGui2` to preserve backward compatability with existing plugins on the current API level.
- Switches all of Dalamud to use `ChatGui2` internally.
- Adds extension methods to `XivChatType2` to obtain the various masked parts of the chat type, collections of all chat types of a kind (channel, source, target), and localized name of a given (un-masked) chat type.

### Benefits for Plugin Devs:

- Chat event handlers now receive data on the origin of a chat message (game, Dalamud, or plugin, as well as which plugin, if applicable).
- More complete chat type listing, including mask values.
- Extension and/or helper methods for common `XivChatType` operations (masking, enumeration, and localization).
- The ability to get your message's index in `RaptureLogModule`.  This is not needed by many plugins, but was previously extremely tedious to obtain when required.

### Porting your plugin to ChatGui2:

- Change any instances of `ChatGui` to `ChatGui2`.  If you specify a required version for your dependency injection of the ChatGui, update it to 2.0 (this is done so that any plugins that are not updated when `ChatGui2` replaces `ChatGui` break in the manner expected by the plugin dev).
	- Update any `ChatGui` event subscribers that you have to use the new parameter lists.
- Change any instances of `XivChatEntry` to `XivChatEntry2`.  If you previously used the `SenderId` field, this has now been renamed `Timestamp` to reflect what it *actually* does.
- Add your `DalamudPluginInterface` as a parameter when calling `ChatGui2.Print()`, `ChatGui2.PrintError()`, or `ChatGui2.PrintChat()`.  You *can* pass null, but it will result in your plugin being unnamed as the message source, so is not recommended.
- Change any instances of `XivChatType` to `XivChatType2`.  (The old XivChatType was left intact until the next API breakage to avoid issues with Haplo-type plugins and naive enumeration of available chat channels.)
	- The `SystemError` type has been renamed to `BattleSystemMessage` to reflect what the channel actually is.
- The new `XivChatType2` type has extensions for things that you'd previously manually do.
	- Call `someChatType.GetChatChannel()` if you were previously manually masking (i.e., `someChatType & 0x7F`).
	- Call `XivChatType2Extensions.GetAllChatChannels()` instead of calling `Enum.GetValues(...)`.  This is strongly recommended because `XivChatType2` now also contains values for the target and source masks of a chat type.

### Dalamud changes at next API break:
- Remove the following files: ChatGui.cs, XivChatEntry.cs, XivChatType.cs, XivChatTypeExtensions.cs.
- All instances of ChatGui2, XivChatEntry2, XivChatType2, etc. should be renamed to remove the '2'.
- All plugins using ChatGui2 perform the renaming in the step above.
- All plugins using the old ChatGui update to the new one.

### Uncertainties/Questions/Comments:
- I'm not a huge fan of asking people to pass in their `DalamudPluginInterface` to get the plugin name when printing to chat.  Should having the plugin name just be removed from this implementation, and consider the source enum good enough?  `GetCallingAssembly` is insufficient because of the libraries out there that people use to call chat functions for them.
- Enum localization could be improved as discussed in discord.  We shouldn't need to add values to the enum very often, so the big switch isn't the end of the world, but it's not great.
- At some point, the Dalamud config entry for the default chat channel should be made to use the extensions to get only chat channels and to get the loalized names.


Dalamud/Game/Gui/ChatGui2.cs
```diff
@@ -0,0 +1,580 @@
...
+    }
+
+    /// <summary>
+    /// Queue a chat message. While method is named as PrintChat, it only adds an entry to the queue,
+    /// later to be processed when UpdateQueue() is called.
+    /// </summary>
+    /// <param name="chat">A message to send.</param>
+    /// <param name="pluginInterface">Your plugin's <see cref="DalamudPluginInterface"/>.</param>
+    /// <param name="messagePrintedCallback">A delegate that is invoked once the message is actually printed to the chat log.  The parameter is the RaptureLogModule message index.</param>
+    public void PrintChat(XivChatEntry2 chat, DalamudPluginInterface pluginInterface, Action<nint> messagePrintedCallback = null)
```
goaaats, line 158: You should get the calling plugin via `FindCallingPlugin()` in `PluginManager` instead. If possible, inject it as a dependency into this.

Dalamud/Game/Gui/ChatGui2.cs
```diff
@@ -0,0 +1,580 @@
...
+            this.HandlePrintMessage(this.baseAddress, chat.Type, senderOwned.Address, messageOwned.Address, chat.Timestamp, chat.Parameters, chat.MessageSource, chat.SourceName, chat.MessagePrintedCallback);
+        }
+    }
+
+    /// <summary>
+    /// Queue a chat message. While method is named as Print, it only adds an entry to the queue,
+    /// later to be processed when UpdateQueue() is called.  This is the internal version that should be used by all Dalamud calls to properly set the message source.
+    /// </summary>
+    /// <param name="message">A message to send.</param>
+    internal void Print_Internal(string message)
```
goaaats, line 265: Would probably be happier with `PrintRaw` as a name instead of this.

goaaats (changes requested):
I like this a lot, good job.

goaaats:
I like the auto plugin name in the chat, but some plugins probably won't be able to use this as they intercept + modify actual chat messages and just e.g. inject links and colors after processing them. Might be good to have an escape hatch for that.

goaaats:
Open question: there have been some issues with additional data passed through chat, which Dalamud manages to lose, for things like context menus on senders. Is this still the case? This new system should be able to handle that, might need some RE.

@Aireil I remember that you used to complain about this regularly, does it still happen?

Aireil:
> @Aireil I remember that you used to complain about this regularly, does it still happen?

If you mean on current Dalamud, probably? I don't remember anyone changing anything regarding that. I can test this and see if the issue is still there.

Aireil:
Tested it real quick before my raid, didn't seem to change anything, still missing context menus. For reference, it's this issue #721 .

Dalamud/Game/Gui/ChatGui2.cs
PunishedPineapple: As long as it gets the point across that these, and only these, are the versions that should be called internally by Dalamud.  I guess depending on `FindCallingPlugin`s behavior when it's just Dalamud calling, it would eliminate the need for the internal versions period.

PunishedPineapple:
> I like the auto plugin name in the chat, but some plugins probably won't be able to use this as they intercept + modify actual chat messages and just e.g. inject links and colors after processing them. Might be good to have an escape hatch for that.

The intent wasn't so much plugin name in chat, but just providing more information for chat event subscribers exactly where a chat message originated.  If we want it to be chat entries get auto-tagged in the log too, let me know.

I'll work on the other stuff later tonight.

PunishedPineapple:
I give up honestly.  I'm just too damn stupid to figure out how to properly use git I guess.  Tried to rebase to get `FindCallingPlugin` on my branch, and now I have like 40 of other people's commits in this PR.  It looked like it made sense in the commit log before I pushed, but I guess not.  You should probably just have someone that can actually tell their left hand from their right do the chat stuff, because trying to fix this mess is beyond me.  That's unacceptable for changes to Dalamud itself.

The only things from this disaster of a PR that might be of use if someone else picks this up are probably the researched missing channel names and source/target masks, and the brief discussion in discord of why propagating sender info to messages printed in response to a ChatGui event is going to be kind of gross (https://discord.com/channels/581875019861328007/860813266468732938/1075929572656304289).

reiichi001:
This seems a bit rash? The PR itself looks good and Dalamud PRs are already monumental tasks, especially for features instead of fixes. You're by no means stupid. Frustrated with git, maybe. (And that's valid.)

@ackwell might be able to help with some git tips to get the commits cleaned up, but I don't think that's something you necessarily should worry about if the code itself is looking good and on the track toward merge readiness.

PunishedPineapple:
The code is probably not in a sufficient state to cleanly implement the additional new features that had been discussed over the last couple days in the discord.

Regardless, the bottom line is that I don't have the intelligence to be contributing to this community, and I *especially* don't trust myself enough to be contributing to the underlying framework where errors could potentially result in risking Dalamud users' accounts. It is unpleasant to rip that bandaid off, but I'm fortunate that it was just SCM issues that truly drove the point home, and not something more serious.

reiichi001:
> I don't have the intelligence to be contributing to this community

While a closed PR is absolutely not the ideal place for this conversation, I feel the need to reiterate that I disagree with this sentiment and it sucks to read that you think that. 

I'm personally taking a break from the discord and most XL/Dalamud/Plugin related things myself due to various circumstances (mostly IRL) and the break has been restful and invigorating. So I can recommend taking a break if things are stressful, especially if there are any other factors at play. You've made good plugins and you aren't dumb. If you feel like you've bitten off more than you can chew, it's more than fine to give yourself some time and space to sort that out. It doesn't undermine your work and past works.

goaaats:
I agree, I think you did a great job with this, it'd be a shame to see you go. You should reach out on Discord to fix the git stuff, it usually just tends to be one or two commands to sort it out - if not, I hope you don't mind if we run with this PR anyway as it stands.

I sadly didn't have a lot of time to support you with the RE tasks, but I know that there's plenty of people that would be happy to help. Sometimes you have to dig a bit and ask directly, though - again, if you feel up to that, I can set it up. Not that there's any rush with this either.

PR #1129, opened 2023-02-20
Better local time for local builds

Aireil:
Using 24-hour clock and just prettifying💅.
Not sure if this has other implications, didn't find anything using it except the dev bar when using debug builds.


KazWolfe:
Would it make more sense to store the local build time as a unix timestamp and just localize to the client?

Aireil:
It's for debug builds, doesn't matter, just triggered me reading it like it was lol.

goaaats (approved):

PR #1131, opened 2023-02-22
Update Font Awesome

kalilistic:
**Changes**
- Update enum to 6.3.0 which adds a new ~400 icons.
- Mark bad enums (e.g. brands) with Obsolete attribute
- Add search terms and category attributes from font awesome metadata
- Separate SE font / FA font into separate data windows.
- Built out FA Font window with search and category filtering.

**Notes**
- Created https://github.com/goatcorp/Dalamud.FASharpGen to make future updates easier.
- Kept original enum names and unicode values to avoid breakage.
- Width slightly decreased on icons but don't see any obvious issues after looking at a few plugins.
- Excluded new number/alphabet icons since weren't displaying (maybe conflict? can revisit later).
- Performance could be a little better on data screen when actively searching (but seemed good enough for dev menu).

**Next API Level**
- Can remove Obsolete enum members
- Can consider aligning name/values to modern FA (if worthwhile?)

**Data Screen Screenshot**
image


kalilistic:
Pre-Req: https://github.com/goatcorp/DalamudAssets/pull/108

PR #1137, opened 2023-02-25
Unelevate on game launch unless --no-unelevate is provided

Soreepeong:
See: https://devblogs.microsoft.com/oldnewthing/20190425-00/?p=102443


avafloww:
Has this been tested on Wine?

KazWolfe:
Another question that came up: what would the impact be on users that have already _been_ launching the game as admin? While I wholeheartedly endorse this feature being added, I do have some concerns of the increased support burden _especially_ given how frustrating it can be to fix file ownership at times.

How difficult would it be to disable this for users who are already in a "tainted" state, but apply to everyone normally (unless explicitly specified)?

KazWolfe:
Alright, finally got around to testing this, at least on macOS. While the game does launch (yay), the system does throw an exception:

```
2023-04-19 20:16:34.637 -07:00 [WRN] [GameStart] Failed to prepare for launching the game unelevated.
System.Runtime.InteropServices.COMException (0x80070578): Invalid window handle. (0x80070578)
   at System.Runtime.InteropServices.Marshal.ThrowExceptionForHR(Int32 errorCode)
   at Dalamud.Injector.GameStart.LaunchGame(String workingDir, String exePath, String arguments, Boolean dontFixAcl, Boolean dontUnelevate, Action`1 beforeResume, Boolean waitForGameWindow) in C:\goatsoft\companysecrets\injector\GameStart.cs:line 79
```

This honestly... seems expected on Wine? The [relevant failing code snippet](https://github.com/goatcorp/Dalamud/pull/1137/files#diff-6eb7a6a1d37fe632b6bb957a587af09f56f684cf020558df10ce67789466e94aR78) is:

```cs
if (PInvoke.GetWindowThreadProcessId(PInvoke.GetShellWindow(), out var shellProcessId) == 0)
    Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
```

If I had to guess (and Google around), Wine returns a `NULL` for `GetShellWindow()` (at least until [something calls `SetShellWindow()`](https://www.winehq.org/pipermail/wine-bugs/2015-March/409651.html), which I don't think is likely). 

Given that the permission (de)escalation thing is more or less irrelevant on non-Windows systems (and *should be*), would it be possible to totally ignore this behavior on Linux systems and/or bail gracefully if there's no shell PID? This also seems in line with what we'd expect per Raymond Chen.

Dalamud.Injector/GameStart.cs
```diff
@@ -64,6 +69,63 @@ public static Process LaunchGame(string workingDir, string exePath, string argum
...
             }
 
+            var procAttrListBuffer = nint.Zero;
+            var parentProcessHandle = nint.Zero;
+            if (!dontUnelevate)
+            {
+                try
+                {
+                    if (PInvoke.GetWindowThreadProcessId(PInvoke.GetShellWindow(), out var shellProcessId) == 0)
+                        Marshal.ThrowExceptionForHR(Marshal.GetHRForLastWin32Error());
```
KazWolfe, line 79: See comment in PR chain - can this (gracefully) stop the unelevation process and/or somehow not get run at all on *nix operating systems?

KazWolfe (dismissed):

PR #1140, opened 2023-02-26
pluginInterface: Add function to open the plugin installer

Infiziert90:


goaaats:
This is probably only worth it if you also make it focus the calling plugin in the installed plugins list.

Infiziert90:
I'm sadly out of ideas on how to focus it in the installer.
Opening the CollapsingHeader doesn't seem like a bigger problem, but getting to focus ...

goaaats:
Can't you calculate the scroll position based on the position in the list? All of the headers have the same size, you would collapse all of them but the plugin that's asking for focus

Infiziert90:
Mhm that is possible, would require making some changes to prevent updates from just breaking it. 
And I would have to find out how you can grab the imgui window and change its ScrollBar position from outside.

Other question that came to my mind, what to do if the user already has the installer open, this would reset there current position to something completely else?

Aireil:
Maybe swap it to installed plugins tab and have the search set with the plugin name? Or with anything for that matter, I was thinking of adding that to /xlplugins actually, to be able to fill the search bar as an argument 🤔.

goaaats:
Search is probably a good compromise.

Infiziert90:
Added 2 new function to open the plugin installer in the correct category and search for the plugin name.

A bool is used to prevent it from clearing the search on open and also used to clear the search again on close.

goaaats (approved):

PR #1141, opened 2023-02-26
feat: expose plugin name on interface

kalilistic:
Expose plugin name so don't need to pass it around to service classes.


goaaats:
This should probably clarify name vs. internal name. What's there right now is the internal name.

kalilistic:
Renamed property and comment to internal name.

goaaats (approved):

PR #1143, opened 2023-02-26
feat: validate player name

kalilistic:
**Test Cases (Input => Result)**
```
Rycharde Leatherby => True
Leatherby Goodfellow => True
Lodewicus Barlow => True
Hi Hi => True
J Smith => False
John S => False
Johnnnnnnnnnnn Sm => True
John Smithhhhhhhhhhhh => False
Johnnnnnnnnnnn Smithhhhhhh => False
Ke$ha Smith => False
Joh``n Smith => False
-John Smith => False
John `Smith => False
Jo`-hn Smith => False
Jo-`hn Smith => False
```

### DIFF
--- a/Dalamud/Utility/SeStringExtensions.cs
+++ b/Dalamud/Utility/SeStringExtensions.cs
@@ -14,4 +14,18 @@ public static class SeStringExtensions
     /// <param name="originalString">The original Lumina SeString.</param>
     /// <returns>The re-parsed Dalamud SeString.</returns>
     public static SeString ToDalamudString(this Lumina.Text.SeString originalString) => SeString.Parse(originalString.RawData);
+
+    /// <summary>
+    /// Validate if character name is valid.
+    /// Both forename and surname must be between 2 and 15 characters and not total more than 20 characters combined.
+    /// Only letters, hyphens, and apostrophes can be used.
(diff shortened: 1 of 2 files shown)


karashiiro:
This will probably need to be patched over in the CN fork - I think CN has slightly different restrictions.

kalilistic:
> This will probably need to be patched over in the CN fork - I think CN has slightly different restrictions.

Anything I can change to make that easier?

karashiiro:
@Bluefissure

Bluefissure:
There's no last name in CN character names, you can go ahead and we will patch it in our fork.

Thanks for letting me know btw. :)

Loskh:
I found the function used in the game to validate player names, and tested it in both CN client and Global client. It works fine.
I think using this function can improve non-global client compatibility.
Code is here: https://gist.github.com/Loskh/c0c075d1494c9b1c63402e4bf5d06485

kalilistic:
Very nice. I think this would require a CS update? I'd suggest I can close this PR and you can submit the function instead.

kalilistic:
Maybe still worth merging? Someone could enhance this in the future using a func without breaking the API.

goaaats (approved):

PR #1147, opened 2023-03-02
feat: remove obsolete icons [breaking change]

kalilistic:
Remove obsolete icons. Potentially breaking change for the next API level.

### DIFF
--- a/Dalamud/Interface/FontAwesome/FontAwesomeIcon.cs
+++ b/Dalamud/Interface/FontAwesome/FontAwesomeIcon.cs
@@ -20,28 +20,6 @@ public enum FontAwesomeIcon
     None = 0,
 
     /// <summary>
-    /// The Font Awesome "500px" icon unicode character.
-    /// </summary>
-    [Obsolete]
-    [FontAwesomeSearchTerms(new[] { "500px" })]
-    _500Px = 0xF26E,
-
-    /// <summary>
-    /// The Font Awesome "accessible-icon" icon unicode character.
-    /// </summary>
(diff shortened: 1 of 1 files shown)


Dalamud/Interface/FontAwesome/FontAwesomeIcon.cs
```diff
@@ -3375,20 +2619,6 @@ public enum FontAwesomeIcon
     [FontAwesomeCategoriesAttribute(new[] { "Maps" })]
     Directions = 0xF5EB,
 
-    /// <summary>
-    /// The Font Awesome "discord" icon unicode character.
-    /// </summary>
-    [Obsolete]
-    [FontAwesomeSearchTerms(new[] { "discord" })]
-    Discord = 0xF392,
```
avafloww, line 3383: Do we want to make an exception to keep the Discord icon, i.e. for Discord links within plugins?

avafloww (approved):
@kalilistic do we still want to do this? I'm ok with bringing it into v9 if so, just had a question about one icon.

Dalamud/Interface/FontAwesome/FontAwesomeIcon.cs
kalilistic: None of these icons actually render today (including Discord). Font Awesome includes them in a separate "brands" file. This was excluded to keep the atlas manageable.

avafloww: Ah, got it!

PR #1151, opened 2023-03-08
Walk dependencies to unload services

goaaats:


Dalamud/Game/Framework.cs
```diff
@@ -492,7 +492,10 @@ private bool HandleFrameworkDestroy(IntPtr framework)
         Log.Information("Framework::Destroy!");
         Service<Dalamud>.Get().Unload();
         this.RunPendingTickTasks();
-        ServiceManager.UnloadAllServices();
+
+        // why did we do this here? EntryPoint also does it when the signal is set, what sense does that make
+        // we should definitely wait for pending tick tasks though
+        // ServiceManager.UnloadAllServices();
```
Soreepeong, line 498: `Dalamud.Unload` sets unloadSignal, and then returns immediately. Even if there was nothing to do while RunPendingTasks happen, game unload will continue if we return early, and the game will call `ExitProcess` as soon as it's done cleaning up, but Dalamud might be not done unloading yet. Framework::Destroy would be a good place to ensure that Dalamud is unloaded before game continues to unload and exit the process.

PR #1154, opened 2023-03-11
Add GameConfig service

Caraxi:
This PR implements a new `GameConfig` service which will allow plugins to easily read and write config values from the game.

### DIFF
--- a/Dalamud/Game/Config/GameConfig.cs
+++ b/Dalamud/Game/Config/GameConfig.cs
@@ -16,7 +16,7 @@ namespace Dalamud.Game.Config;
 /// </summary>
 [InterfaceVersion("1.0")]
 [PluginInterface]
-[ServiceManager.BlockingEarlyLoadedService]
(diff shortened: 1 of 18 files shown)


Dalamud/Game/Config/GameConfig.cs
```diff
@@ -0,0 +1,450 @@
...
+using Serilog;
+
+namespace Dalamud.Game.Config;
+
+/// <summary>
+/// This class represents the game's configuration.
+/// </summary>
+[InterfaceVersion("1.0")]
+[PluginInterface]
+[ServiceManager.BlockingEarlyLoadedService]
```
goaaats, line 19: Is there a reason this is BlockingEarlyLoadedService instead of a regular EarlyLoadedService?

Caraxi: because every other service was and I have no idea how any of this works

Caraxi:
git why you do that

PR #1170, opened 2023-03-31
[Fools23] Add exception handling to the activation of AU plugins.

Berna-L:
Also changed Hey, Dalamud! to let FoolsManager... manage the exception.

### DIFF
--- a/Dalamud/Fools/FoolsManager.cs
+++ b/Dalamud/Fools/FoolsManager.cs
@@ -188,8 +188,15 @@ internal class FoolsManager : IDisposable, IServiceType
             return;
         }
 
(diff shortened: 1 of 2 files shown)


Berna-L:
With this, the install button doesn't change to disable if an exception occurs during initialization, instead of registering as enabled and messing up the UI a bit if the user tries to disable it (since calling Dispose on the plugin object will NPE).

Since Hey, Dalamud! is the only AU plugin that explicitly catches an exception and it prints a message in chat, I didn't implement any kind of pop-up or message that it couldn't be enabled, since I hope the user will check the message in chat.

goaaats (approved):

PR #1181, opened 2023-04-11
Profiles in master

goaaats:


KazWolfe:
Everything seems stable on my end. I'm a bit "hm" about the name difference between API/UI, but I guess that's not particularly relevant as it's just end-user naming

In what case would a plugin not want to support ~~profiles~~ collections?

KazWolfe (approved):

goaaats:
The name change was a last minute thing, since I got feedback that "profiles" doesn't make much sense for what they actually are.

goaaats:
> In what case would a plugin not want to support profiles collections?

I imagined that something like penumbra, that might not support toggling at runtime very well, might want to enable that. Not sure if it's actually needed though, if it turns out no one's using it we can remove it.

PR #1187, opened 2023-04-21
Add Happy Eyeballs Support

### LINKED ISSUE #1186: Add Happy Eyeballs to our HttpClient
The Discord has recently been absolutely flooded by users experiencing issues connecting to Kamori and other sites that use IPv6. Generally, this is caused by local ISP hiccups or other problems where a user's IPv6 connection is unstable. Our typical solution has been to tell users to enable a VPN or disable IPv6 entirely, neither of which are particularly _good_ solutions, especially as browsers don't have this problem. Enter: [Happy Eyeballs](https://www.rfc-editor.org/rfc/rfc8305).

While it would really be better for [Microsoft to do this](https://github.com/dotnet/runtime/issues/26177) for everyone and everything, it may make sense for us to just do it ourselves for now. Osu! [has an implementation](https://github.com/ppy/osu-framework/pull/4191/files) that we can look at borrowing as well.

The scope of this Issue is to ensure that Happy Eyeballs isn't already available in .NET (and just needs us to pass an argument somewhere), and if not, to borrow Osu's implementation (and others) for our own use case.

KazWolfe:
This pull request aims to solve a number of issues related to bad IPv6 compatibility, especially an inability to load Kamori when one's ISP hiccups. This is done by implementing a variant of the Happy Eyeballs algorithm, as described in [RFC 8305](https://www.rfc-editor.org/rfc/rfc8305). The specific implementation we're using here was derived from [a pull request to the `jellyfin` project](https://github.com/jellyfin/jellyfin/pull/8598), with some tweaks that hopefully would help our use case.

This pull request provides a new class meant for public consumption: `HappyEyeballsCallback`. This class provides a `ConnectCallback` method that can be assigned to a [`SocketsHttpHandler`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler?view=net-7.0) when creating an [`HttpClient`](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=net-7.0). `HappyEyeballsCallback` implements the following behavior:

1. If set, the `forcedAddressFamily` will be used and all evaluation will be skipped.
2. The callback will check for a cached `AddressFamily` for the DNS name being used. If it exists, the cached value is used and future steps will be skipped.
3. The callback will race an IPv4 and IPv6 connection attempt (with the IPv4 connection delayed by a configurable time period, defaulting to 100ms).
4. The first successful connection attempt to complete will return and be cached for that hostname for the rest of the session. If no connection succeeds, the error propagates up the chain.

When a connection is successfully established, the resulting `AddressFamily` is cached for that DNS hostname and all future connections for the life of the application will use that `AddressFamily`. 

The `HappyEyeballsCallback` class provides state in the form of the DNS/AddressFamily cache, which can (but does not have to be) shared between multiple `HttpClient`s if desired. Plugin developers should normally create and maintain their own callback with their own state.

This pull request also adds a new service, the `HappyHttpClient`, which is intended to be shared between all Dalamud subsystems that can use a common `HttpClient` and obsoletes `Util.HttpClient`. Further, all uses of `Util.HttpClient` and any new non-customized `HttpClient`s are replaced with references to the `HttpClient` provided by this service.

This pull request, if merged, fixes #1186.

### DIFF
--- a/Dalamud/Networking/Http/HappyEyeballsCallback.cs
+++ b/Dalamud/Networking/Http/HappyEyeballsCallback.cs
@@ -87,6 +87,12 @@ public class HappyEyeballsCallback : IDisposable
             return this.forcedAddressFamily.Value;
         }
 
+        // Force IPv4 if IPv6 support isn't detected to avoid the resolution delay.
+        if (!Socket.OSSupportsIPv6)
+        {
+            return AddressFamily.InterNetwork;
+        }
+
         if (this.addressFamilyCache.TryGetValue(context.DnsEndPoint, out var cachedValue))
         {
             // TODO: Find some way to delete this after a while.
--- a/Dalamud/Networking/Http/HappyEyeballsCallback.cs
+++ b/Dalamud/Networking/Http/HappyEyeballsCallback.cs
@@ -24,17 +24,17 @@ public class HappyEyeballsCallback : IDisposable
     private readonly ConcurrentDictionary<DnsEndPoint, AddressFamily> addressFamilyCache = new();
 
     private readonly AddressFamily? forcedAddressFamily;
-    private readonly int ipv4WaitMillis;
+    private readonly int ipv6GracePeriod;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="HappyEyeballsCallback"/> class.
     /// </summary>
     /// <param name="forcedAddressFamily">Optional override to force a specific AddressFamily.</param>
-    /// <param name="ipv4WaitMillis">Time to wait before initiating the IPv4 request.</param>
-    public HappyEyeballsCallback(AddressFamily? forcedAddressFamily = null, int ipv4WaitMillis = 100)
+    /// <param name="ipv6GracePeriod">Grace period for IPv6 connectivity before starting IPv4 attempt.</param>
+    public HappyEyeballsCallback(AddressFamily? forcedAddressFamily = null, int ipv6GracePeriod = 100)
     {
         this.forcedAddressFamily = forcedAddressFamily;
-        this.ipv4WaitMillis = ipv4WaitMillis;
+        this.ipv6GracePeriod = ipv6GracePeriod;
     }
 
     /// <inheritdoc/>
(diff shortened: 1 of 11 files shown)


Dalamud/Networking/Http/HappyHttpClient.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// </summary>
+    public HttpClient SharedHttpClient { get; }
+
+    /// <summary>
+    /// Gets a <see cref="SocketsHttpHandler"/> meant to be shared across any custom <see cref="HttpClient"/>s that
+    /// need to be made in other parts of the application.
+    ///
+    /// This handler comes pre-loaded with a common <see cref="HappyEyeballsCallback"/> for your convenience.
+    /// </summary>
+    public SocketsHttpHandler SharedSocketsHandler { get; }
```
KazWolfe, line 47: Unsure if exposing `SharedSocketsHandler` is a good idea at all. Stray `HttpClient`s not created properly can dispose this and lead to all sorts of problems.

Consider exposing the `HappyEyeballsCallback` instance instead for consumption, as that should only really be GC'd. Or, optionally, `Dispose()`d alongside the service itself, where `Dispose()` clears the cache.

This should *definitely* not be a `public` attribute of this service.

Dalamud/Networking/Http/HappyHttpClient.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Net.Http;
+using Dalamud.IoC;
+
+namespace Dalamud.Networking.Http;
+
+/// <summary>
+/// A service to help build and manage HttpClients with some semblance of Happy Eyeballs (RFC 8305 - IPv4 fallback)
+/// awareness.
+/// </summary>
+[PluginInterface]
```
KazWolfe, line 12: Don't expose this to plugins; they should manage their own `HttpClient`s.

(TBD: If we expose the `HappyEyeballsCallback`, do we want plugins to be able to use that instead?)

Dalamud/Plugin/Internal/Types/PluginRepository.cs
```diff
@@ -24,7 +25,7 @@ internal class PluginRepository
 
     private static readonly ModuleLog Log = new("PLUGINR");
 
-    private static readonly HttpClient HttpClient = new()
+    private static readonly HttpClient HttpClient = new(Service<HappyHttpClient>.Get().SharedSocketsHandler)
```
KazWolfe, line 28: Create a new `SocketsHttpHandler` here for this and just link to the callback.

KazWolfe:
Potential future work:

* Fetch all DNS requests and race them, more in line with how Happy Eyeballs _actually_ works. This will also help handle annoying cases where DNS returns a bad/invalid host.
* Figure out a way to "expire" records from the address cache. Unsure if `WeakReference` works on enums?

Dalamud/Networking/Http/HappyEyeballsCallback.cs
```diff
@@ -0,0 +1,118 @@
...
+        }
+
+        // The loser can still fail, but we'll wait for it.
+        // If it succeeds, cache the result. If not, just throw the exception up the chain.
+        var loser = victor == tryV6 ? tryV4 : tryV6;
+        var loserStream = loser.GetAwaiter().GetResult();
+
+        this.addressFamilyCache[context.DnsEndPoint] = loserStream.Socket.AddressFamily;
+
+        return loserStream;
```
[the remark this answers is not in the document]
goaaats: Task.WhenAny()

Dalamud/Networking/Http/HappyHttpClient.cs
goaaats: Yeah, let's just expose the callback in utility namespace somewhere so people can do it if they want to, for now.

KazWolfe:
## Testing Strategy

Testing this particular change is a bit annoying as `HttpClient` will keep connections alive. Therefore, if a (successful) IPv6 connection is made and IPv6 connectivity is subsequently dropped, all future connections (at least until `HttpClient` decides to close the connection) will experience issues.

With that out of the way, testing this seems easiest with [`clumsy`](https://github.com/jagt/clumsy) and setting a high lag time (e.g. 5 seconds) for IPv6 connections. Without this PR, opening the plugin installer should take approximately the delay time (or double) while everything is fetched. _With_ this PR, the connection should still go through reasonably quickly. My testing strategy based on this has been as follows:

1. Launch the game with IPv6 compatibility and no `clumsy` rules.
2. Open the Plugin Installer and time the request. Close the installer.
3. Enable `clumsy` and either re-enable the game or wait ??? for the keepalive to expire.
4. Open the Plugin Installer and time the requests. This should roughly match the same timing as from Step 2.

Some more advanced simulation is possible as well. For example, traffic to Cloudflare's IPv6 network can be completely blocked (see [their IP list](https://www.cloudflare.com/ips/)) using `clumsy`, which would cause Kamori to be unable to load without this PR.

Dalamud/Networking/Http/HappyEyeballsCallback.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            var race = AsyncUtils.FirstSuccessfulTask(new List<Task<NetworkStream>>
+            {
+                tryConnectIPv6,
+                this.AttemptConnection(AddressFamily.InterNetwork, context, linkedToken.Token),
+            });
+
+            // If our connections all fail, this will explode with an exception.
+            stream = race.GetAwaiter().GetResult();
+        }
```
KazWolfe, line 86: Feeling messy again :/

goaaats (approved):

PR #1189, opened 2023-04-27
Happy Eyeballs Rev 2: The DNS-ening

KazWolfe:
This pull request aims to fix some of the issues seen in Happy Eyeballs Revision 1, as well as generally make everyone's experiences better.

Firstly, this pull request switches the connection strategy to leverage DNS lookups instead of a blind IPv6-first strategy. This better allows responding to cases where users do not have IPv6 connectivity at all, as well as better handling cases where a machine is deprioritizing IPv6 for some reason. This also allows for _all_ DNS records to be tried sequentially, in case any one particular returned record is invalid.

Secondly, this pull request leverages some quirks of task control flow (and exceptions) to suppress unnecessary exceptions generally caused by a cancelled/ignored `Task` getting GC'd. Instead, HTTP-related exceptions will only be thrown if a connection could not be achieved.

Lastly, this pull request removes state tracking and caching from the Happy Eyeballs callback. This is mostly due to the fact that `HttpClient` already keeps connections alive for long enough, but also should help resolve problems in rare edge cases where users lose IPv6 connectivity midway through a game session. Ideally, some caching would be added back in at some point, but this is non-critical.

This pull request also adds a number of minor changes:
- Fix a bug where `HappyHttpClient` was not using its own shared callback class.
- Fix a semantic issue where the wrong variable was being addressed.
- Decreases connection backoff from 100ms to 75ms.
- Add a new `ZipperMerge` utility to merge two or more `IEnumerable`s together.
- Add a new `CancellableDelay` utility to save on having to try-catch `TaskCanceledException`s on site.

### DIFF
--- a/Dalamud/Networking/Http/HappyEyeballsCallback.cs
+++ b/Dalamud/Networking/Http/HappyEyeballsCallback.cs
@@ -83,12 +83,7 @@ public class HappyEyeballsCallback : IDisposable
         // If we're here, it means we have a successful connection. A failure to connect would have caused the above
         // line to explode, so we're safe to clean everything up.
         linkedToken.Cancel();
-        tasks.ForEach(task =>
-        {
-            // Suppress any other exceptions that come down the pipe. We've already thrown an exception we cared
-            // about above.
-            task.ContinueWith(inner => { _ = inner.Exception; });
-        });
+        tasks.ForEach(task => { task.ContinueWith(this.CleanupConnectionTask); });
 
         return stream;
     }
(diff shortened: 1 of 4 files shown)


KazWolfe:
I still need to test this quite a bit more, but so far everything _seems_ fine. Test strategy is the same as it was in #1187 for the most part.

KazWolfe:
Did a bit more testing tonight, nothing major seems wrong. Still can probably use some other tire kickers and/or some time in `stg` though.

goaaats (approved):

PR #1195, opened 2023-05-05
feat: load the FileDialog drive list asynchronously

### LINKED ISSUE #1130: File picker hangs game for several minutes on open
I've mentioned this in Discord before, but this is becoming increasingly frustrating, so I figured I'd make an actual issue.

Unfortunately, I can't really offer reproduction steps beyond "open the file picker". Happens in Penumbra, for example, but I've confirmed it in various plugins. Open, the game hangs for several minutes, and eventually it comes back and works fine until you need to open it again.

Otter mentioned it may be due to unresponsive network locations and `Directory.Exists` when populating the sidebar, but all I have is one samba share. If that *is* the cause, surely it's better to have that hang when you click on the location versus every single time the damn window opens. It's definitely not feasible to ask users to change their network configuration for Dalamud, either.

This showed up roughly around 6.3.

karashiiro:
Should fix #1130.

This change makes FileDialog load the drive list asynchronously, so that slower network-mapped drives don't cause enumeration of the drive list to lock up the whole game. I wrote this so that the drive list can be enumerated while it's still loading, so it uses ImmutableArray boxed in IReadOnlyList to avoid exposing implementation details. `DriveListLoader.Drives` is only accessed once each frame, so this should have a negligible memory footprint.

If desired, this can be expanded to make use of `DriveListLoader.Loading` to communicate to the user that stuff is happening. Additionally, this supports reloading the drive list safely, if we ever want to do that.

### DIFF
--- a/Dalamud/Interface/ImGuiFileDialog/DriveListLoader.cs
+++ b/Dalamud/Interface/ImGuiFileDialog/DriveListLoader.cs
@@ -8,7 +8,7 @@ namespace Dalamud.Interface.ImGuiFileDialog;
 /// <summary>
 /// A drive list loader. Thread-safety guaranteed.
 /// </summary>
-public class DriveListLoader
+internal class DriveListLoader
 {
     private bool initialized;
 
--- a/Dalamud/Interface/ImGuiFileDialog/DriveListLoader.cs
+++ b/Dalamud/Interface/ImGuiFileDialog/DriveListLoader.cs
@@ -1,0 +1,71 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Dalamud.Interface.ImGuiFileDialog;
+
+/// <summary>
+/// A drive list loader. Thread-safety guaranteed.
+/// </summary>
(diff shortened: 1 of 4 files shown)


Dalamud/Interface/ImGuiFileDialog/DriveListLoader.cs
```diff
@@ -0,0 +1,71 @@
...
+using System.Collections.Immutable;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Dalamud.Interface.ImGuiFileDialog;
+
+/// <summary>
+/// A drive list loader. Thread-safety guaranteed.
+/// </summary>
+public class DriveListLoader
```
goaaats, line 11: Why is this public? I feel like this should be an implementation detail. Otherwise, looks good.

goaaats (changes requested):

Dalamud/Interface/ImGuiFileDialog/DriveListLoader.cs
karashiiro: Updated

goaaats (approved):

PR #1196, opened 2023-05-10
fix: only check FileDialog location existence during navigation

karashiiro:
Even when `GetDrives` is async, the FileDialog code still checks for drive existence on the UI thread, causing the same blocking issues. This changes FileDialog to only check for location existence when actually navigating to a location.


KazWolfe (approved):
LGTM and tests ok; though this still will hang the game on trying to access an offline disk. Not sure if we can reasonably avoid that, however.

karashiiro:
updated branch

karashiiro:
Any updates?

PR #1208, opened 2023-05-23
add 24pt font scale

avafloww:
Super minor, but it always bugged me that there's not an easy way to set this to 2*default (2*12=24) for 200% scaling on 4K.


KazWolfe:
yeah sure ok

KazWolfe (approved):

PR #1217, opened 2023-05-25
Drag & Drop

Ottermandias:
- Adds external WinAPI based drag & drop support and helper functions to create ImGui support from that.
- Accessible via UiBuilder
- Does not introduce C# dependencies, but uses pinvoke on ole32.dll and shell32.dll.
- I can not test how it behaves on non-windows systems. We could make Enabling it dependent on platform or something, if it causes issues, maybe?

### DIFF
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -10,7 +10,6 @@ using System.Threading.Tasks;
 using Dalamud.Configuration.Internal;
 using Dalamud.Game;
 using Dalamud.Game.Gui.Internal;
-using Dalamud.Interface.DragDrop;
 using Dalamud.Interface.Internal;
 using Dalamud.Plugin.Internal;
 using Dalamud.Utility;
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -8,6 +8,7 @@ using System.Threading.Tasks;
 using Dalamud.Configuration.Internal;
 using Dalamud.Game;
 using Dalamud.Game.Gui.Internal;
+using Dalamud.Interface.DragDrop;
 using Dalamud.Interface.Internal;
 using Dalamud.Plugin.Internal;
 using Dalamud.Utility;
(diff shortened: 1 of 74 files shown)


Dalamud/Dalamud.cs
```diff
@@ -135,6 +136,9 @@ public void DisposePlugins()
         // will not receive any windows messages
         Service<DalamudIME>.GetNullable()?.Dispose();
 
+        // this must be done before unloading interface manager, since it relies on the window handle members.
+        Service<DragDropManager>.GetNullable()?.Dispose();
```
goaaats, line 140: Is there a reason you're not depending on InterfaceManager in DragDropManager instead? That should satisfy the condition in your comment. You can also use the InherentDependencyAttribute.

Dalamud/Interface/UiBuilder.cs
```diff
@@ -100,6 +102,11 @@ internal UiBuilder(string namespaceName)
     /// </summary>
     public event Action HideUi;
 
+    /// <summary>
+    /// Gets the manager for external, WinAPI-based drag and drop functionality.
+    /// </summary>
+    public IDragDropManager DragDropManager => this.dragDropManager;
```
goaaats, line 108: It's probably safer for DragDropManager to be a plugin service instead, since the DI will recognize that and handle its lifecycle properly. See e.g. GameConfig or PartyList.

goaaats (changes requested):
Looks great otherwise. The DI stuff needs to be documented a little more, so no worries, no way you could have know that. The only thing I'm wondering about is if having the window be a drag & drop target at all times changes some input behaviour, but I reckon we won't know until we try.

Ottermandias:
Update should be fine. Tested it quickly with Penumbra and it seemed to work right. Don't know how it handles on Wine though. If this causes problems we could just make the "Enable()" call dependent on system or whatever. (could also make it an option for users, of course).

goaaats (approved):

PR #1225, opened 2023-05-28
Use vtable replacement as swapchain hook method

Soreepeong:
This at least does not crash when discord screenshare is attached. Need testing with other graphics injector programs.


Soreepeong:
Apparently this does not work with reshade atm

PR #1235, opened 2023-06-01
Interface support for plugin DI

goaaats:
To allow for better discoverability of services and easier transition to scoped services

### DIFF
--- a/Dalamud/Dalamud.csproj
+++ b/Dalamud/Dalamud.csproj
@@ -8,7 +8,7 @@
     </PropertyGroup>
 
     <PropertyGroup Label="Feature">
-        <DalamudVersion>7.6.1.0</DalamudVersion>
+        <DalamudVersion>7.6.2.0</DalamudVersion>
         <Description>XIV Launcher addon framework</Description>
         <AssemblyVersion>$(DalamudVersion)</AssemblyVersion>
         <Version>$(DalamudVersion)</Version>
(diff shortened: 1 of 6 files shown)


KazWolfe (approved):

avafloww (commented):
I'm in favour of these changes generally, but have some thoughts on the naming convention for the services.

Although I generally am a fan of the `I<interface name>` naming convention we have for interfaces, I don't particularly think it should be relevant for users of services to know if their service is an interface or a class, and it feels a bit ugly in the case of services imho. My opinion on how we should handle this:
- Push this change out to api-next / v9; although it would be nice to merge the changes in the current API version, I personally don't see it as strictly necessary, or something that should guide our changes
- Generally interfaces keep the `I` prefix, with the exception of services, which are unprefixed and have private implementations
- To mitigate the removal of the `I` prefix from service types, we could consider centralizing service types somewhere else (though I'm not personally sure if this is necessary or wanted)

I'm OK with being in the minority here on naming conventions if enough folks vote against my proposal, but would like to get other thoughts on it, whether it's "services should strip the `I` prefix", "services should keep the `I` prefix", or even "I don't care" - keeping in mind that core service implementations are likely to go private in a future API version, and that referring to the interfaces would be the only supported way to refer to these services.

KazWolfe:
Something to note here: Internal uses of the `Service<T>.Get()` call will not accept `IAetheryteList` as a valid `T`: 

> The type 'Dalamud.Plugin.Services.IAetheryteList' must be convertible to 'Dalamud.IServiceType' in order to use it as parameter 'T' in the generic class 'Dalamud.Service<T>'

This can be seen easily enough here: https://github.com/goatcorp/Dalamud/blob/5a8fff515fc93becd1f96b598da25ada88b3a65f/Dalamud/Interface/Internal/Windows/DataWindow.cs#L1675

Because this is an internal, I'm not _particularly_ concerned, but it seems like the interface should be usable here.

Issue #1236, opened 2023-06-03
Functions may be partially unhooked during Dalamud unload, leading to crashes/instability


nebel:
While investigating some crash dumps which ocurred at game shutdown, I noticed what seemed like something strange: a native game function was running with what appeared to be unsual, mangled instructions. The issue was inconsistent, but after some extensive debugging it seemed like instructions hooked by Dalamud were only being partially restored when Dalamud unloaded.

This eventually led me here: [HookManager.cs#L136](https://github.com/goatcorp/Dalamud/blob/5a8fff515fc93becd1f96b598da25ada88b3a65f/Dalamud/Hooking/Internal/HookManager.cs#L136)

When we try to remove the hook here, we currently do so by restoring the "original" bytes from the start of the hooked function until we encounter a byte which is unchanged. The problem, however, is that there is a small but not insignificant likelihood that any given byte in the "original" bytes may be the same as the corresponding hook byte in the same position. Because we stop as soon as a single byte is the same, we risk truncating our "restored" bytes.

Consider the following function which we want to hook:

```asm
(opcodes)
	48 83 79 28 00        - cmp qword ptr [rcx+28],00 { 0 }
	0F85 B5000000         - jne ffxiv_dx11.exe+1239050
	C3                    - ret
(bytes)
	48 83 79 28 00 0F 85 B5 00 00 00 C3
```

One possible valid hook for this function, and one which does *not* reproduce this bug is as follows:

```asm
(opcodes)
	FF 24 25 000B0800     - jmp qword ptr [00080B00]
	90                    - nop 
	90                    - nop 
	90                    - nop 
	90                    - nop 
	C3                    - ret 
(bytes)
	FF 24 25 00 0B 08 00 90 90 90 90 C3
```
	
Note that if we compare the byte sequence between this hook and the original function, every byte is different. However another equally valid hook which just so happens to jump to a different memory address is the following:

```asm
(opcodes)
	FF 24 25 D0000F00     - jmp qword ptr [000F00D0]
	90                    - nop 
	90                    - nop 
	90                    - nop 
	90                    - nop 
	C3                    - ret 
(bytes)
	FF 24 25 D0 00 0F 00 90 90 90 90 C3
```
	
The fifth byte in this hook is 00, the same as the original fifth byte. In this case, HookManager will generate a `snippet` of length 4, which when restored over the start of the hook, results in the following:

```asm
(opcodes)
[...]

PR #1237, opened 2023-06-03
Improve unhooking of hooked functions

nebel:
This PR is presented as an attempted solution to issue #1236 reported by me.

In most aspects of these changes I've tried to take a cautious approach. `Unhooker.TrimAfterHook` is called immediately after hooking (rather than at unhook time) with the idea that accidentally capturing additional hooks in adjacent data are less likely if we grab the snapshot as soon as possible.

However, there are a couple things which I can certainly see changing:

1. Currently the pre-trimmed original byte snapshot stores 50 bytes, just as it did before. However this may be excessive, as the longest hook I've seen in practice is 14 bytes. Trimming to 30 or even 20 may be reasonable, but I don't know how long the "longest hook" may actually be in theory. Any ideas?
2. The "naive" unhook logic is kept around for hooks which cannot be trimmed. Right now this only applies to `AsmHook`, of which there is only one in Dalamud itself. However this might be being too cautious, and it may be reasonable just to restore the entire original byte snapshot for untrimmed hooks (especially if we were to reduce this snapshot size as mentioned above).

Please let me know if you have an opinions on the above.

There are certainly a number of different approaches we might take here, and others ways to structure the code itself. I just went with a way that seemed to make sense to me, but I'm happy to hear feedback on any aspect of this PR in addition to the points above.

Thanks.

P.S. The **unhooker-dbg** branch in my repo (https://github.com/nebel/Dalamud/tree/unhooker-dbg) contains essentially the same changes as this PR but with additional debug logging, which was what I used to log the "Mismatch analysis" information presented in the linked issue. Feel free to use this branch for local testing if you're interested in seeing this information for yourself.

### DIFF
--- a/Dalamud/Hooking/AsmHook.cs
+++ b/Dalamud/Hooking/AsmHook.cs
@@ -35,12 +35,8 @@ public sealed class AsmHook : IDisposable, IDalamudHook
     {
         address = HookManager.FollowJmp(address);
 
-        var hasOtherHooks = HookManager.Originals.ContainsKey(address);
-        if (!hasOtherHooks)
-        {
-            MemoryHelper.ReadRaw(address, 0x32, out var original);
-            HookManager.Originals[address] = original;
-        }
+        // We cannot call TrimAfterHook here because the hook is activated by the caller.
+        HookManager.RegisterUnhooker(address);
 
         this.address = address;
         this.hookImpl = ReloadedHooks.Instance.CreateAsmHook(assembly, address.ToInt64(), (Reloaded.Hooks.Definitions.Enums.AsmHookBehaviour)asmHookBehaviour);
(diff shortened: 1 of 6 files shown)


Soreepeong:
I'm confused on why were 0x32 bytes getting stored in the first place, since even the 64bit absolute jump hook shouldn't occupy more than 0x10 bytes. That sounds like it may span over two functions, but since it has been working so far we probably can just ignore this.

FunctionPointerVariableHook though: we should always store/restore exactly 8 bytes. I probably copied that part from other file without giving much thoughts when I was writing that file in particular. It's most likely to be overwriting single entry in vtable, and it probably is safer to only touch exactly 8 bytes rather than possibly undershooting or overshooting it.

`RegisterUnhooker` and ctor for `Unhooker` should have `minBytes` and `maxBytes` for that purposes.

nebel:
Thanks for the feedback!

I went ahead and added `minBytes` and `maxBytes` to the `Unhooker` ctor as required arguments and then to `RegisterUnhooker` as optionals defaulting to 0 and 0x32 respectively. `Unhooker` will then read up to `maxBytes` when storing the originals, but always make sure that at least `minBytes` bytes are kept when trimming or restoring bytes. Then in `FunctionPointerVariableHook` we'll set both to 8.

Let me know if these defaults still make sense to you or if it would be better to require them at each call site. As mentioned I never saw a hook longer than 14 bytes so 0x10 may be fine, but as you say the 0x32 is working so far.

Soreepeong:
Rather than using default values, there should be overloads, as from what I remember, those default values are used as constant values on compiling the code that uses the function. Other than that, looks fine to me.

(Also I just checked out Reloaded.Hooks source code and there are reason why 0x10 bytes might not be enough; ignore that remark!)

nebel:
Thanks! I created overloads for `RegisterUnhooker` so that specifying `minBytes`/`maxBytes` is optional.

By the way, it may not be important in the end but did you have any thoughts about continuing to use the "naive" unhooking method for `AsmHook`? Currently we'll just stop after the first unchanged byte since we can't call `TrimAfterHook` in `AsmHook` itself immediately after the hook.

If this is not ideal we do have options (e.g. store the `Unhooker` the in `AsmHook` instance and call `TrimAfterHook` after it's been enabled) but not sure how important this is since it's only a single hook and might never experience the partial restore problem.

goaaats (approved):

PR #1243, opened 2023-06-08
Per-character configurations

goaaats:


KazWolfe:
Preemptively marking this for API v9 until we can validate that this won't break any existing plugins. Per goat, `PluginConfigurations` _should_ be private anyways so this likely won't have any real world impact, but still needs verification.

Two (non-blocking) questions on this PR otherwise:

1. Are we leaving plugin config folders alone here? I don't see any reason we _need_ to provide per-character folders at a Dalamud level, but at least worth asking to get a solid decision in one direction.
2. How difficult would it be (and, subsequently, how much benefit would there be) to create a config directory per character, e.g. `pluginConfigs/{content_id}/MyPlugin.json`? This feels more extensible (allows implementation of Q1 above later if we so choose), but it may also be excessive and provide little real benefit.

rreminy:
I feel we can consider having both:

- **Global level configs**
These would be the current existing configs stored under `/pluginConfigs`, or could consider moving them *(assuming its not too difficult)* to `/pluginConfigs/Global`.

- **Character level configs**
These could be the `/pluginConfigs_CHAR_{content_id}`, or could consider having them inside `/pluginConfigs` as well.

This may feel outside of the scope of keeping them private however this could be useful for some plugins that could make use of global and pre-character configurations separately, either together or on their own.

Existing variants of the methods and property would be global while per-chatacter would have `Character` on their names to differentiate them.

This also could save the trouble of having to configure some of the plugins separately for each character

KazWolfe:
@rreminy That is basically what we're planning right now.

Plugins can load the plugin-wide configuration just as they already did - `DalamudPluginInterface#GetPluginConfig`. There are now, however, two additional methods:

* `DalamudPluginInterface#GetPluginConfigForCharacter<T>()`, which returns a `T : ICharacterConfiguration` for the currently logged in character.
* `DalamudPluginInterface#GetPluginConfigForContentId<T>(ulong contentId)`, which returns a `T : ICharacterConfiguration` for the specifically listed ContentId.

Plugins may mix-and-match their behavior accordingly to load one or both of these depending on their needs. Prior to this change, if a plugin wished to use per-character configurations, they would need to manage it themselves in their config directory (see: DailyDuty).

---

My comment in the earlier PR is referring specifically to the `PluginConfigurations` class, which is _technically_ part of our public API surface to plugins, but realistically should not be used by any plugin developer (and that class should likely be made `internal`). There should be no change to how plugins load/don't load their plugin-wide configs (or, in the case of plugins that already do per-character work, how those configs are currently loaded) as a result of this change.

rreminy:
I see, alright.

I did not realize this was a PR at the time I posted my comment *(just saw this on my email)*, however this does looks fine.

KazWolfe:
T/N: Smoke passes, doesn't seem to cause any issues with plugin ABIs unless they're using that class somehow. This should be safe to bring into api-current, even if it is a (strict) breakage of the public API.

Either now or as part of API 9, we may want to adjust visibility of that class?

Blooym:
Are there still plans to introduce this?

goaaats:
I would like to, yes, but I think it should go together with a general rework of how we do config and be tied into the VFS that we added and UI that can manage configurations. Sadly, I think this is pretty low prio for me at the moment. If you want this now you should do it inside your existing config, IMO.

PR #1251, opened 2023-06-18
chore: export docfx metadata with build

avafloww:
Adds a build step that runs in parallel with the main build process to export docfx metadata and upload it as a separate workflow artifact.
This will allow us to automatically generate API documentation for the new dev docs site (WIP, will push something up once @goaaats creates the repo for it).

Also changes `fetch-depth` from 0 to 1 to speed up the build process a little bit. The checkout action treats `0` as "fetch all history from every linked repo", which is the opposite of what we want in CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          fetch-depth: 0
+          fetch-depth: 1
       - name: Setup MSBuild
         uses: microsoft/setup-msbuild@v1.0.2
       - name: Define VERSION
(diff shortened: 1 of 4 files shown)


.github/workflows/main.yml
```diff
@@ -105,3 +105,32 @@ jobs:
         git commit -m "[CI] Update staging for ${DVER} on ${GH_BRANCH}" || true
         
         git push origin main || true
+
+  # docfx metadata generation can often take several minutes, so we run it in parallel to the main build
+  generate-docfx-metadata:
+    name: Generate docfx metadata
+    runs-on: windows-2022
```
KazWolfe, line 112: Do we need to run this task on a windows runner (other than the fact that we already use it)? They're notably slower IME than their Linux alternatives.

.github/workflows/main.yml
```diff
@@ -13,7 +13,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: recursive
-          fetch-depth: 0
+          fetch-depth: 1
```
KazWolfe, line 16: This build's checkout phase took 2m13s vs [36s](https://github.com/goatcorp/Dalamud/actions/runs/5259586413/jobs/9505352763) for a prior PR. Fluke?

avafloww: Seems so - this only took 26s on my fork.

avafloww:
Closing this and taking a more isolated route to building docs, where the docs repo does most of the build process, instead of relying on metadata generated by the Dalamud build process.

PR #1252, opened 2023-06-18
Set Dalamud.Boot to cpp20

KazWolfe:
Attempts to fix the build errors more permanently by setting `Dalamud.Boot` to use C++ language version 20. Additionally removes the SDL disable for development builds.


avafloww (approved):
LGTM if it works!

PR #1257, opened 2023-06-21
/xldata window refactor

MidoriKami:
As the `/xldata` window file was getting ***quite*** large, I thought it would be a good idea to split it apart, both to make it simpler to add new modules, and to lookup how a particular module works without having to see a *bunch* of unrelated code.

The functionality of each module was not changed, only some small style fixes, and standards improvements here and there.


MidoriKami:
image

Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
```diff
@@ -0,0 +1,194 @@
...
+    /// <param name="dataKind">Data kind name, can be lower and/or without spaces.</param>
+    public void SetDataKind(string dataKind)
+    {
+        if (string.IsNullOrEmpty(dataKind))
+            return;
+
+        dataKind = dataKind switch
+        {
+            "ai" => "Addon Inspector",
+            "at" => "Object Table", // Actor Table
```
avafloww, line 100: this seems a little bit fragile; can we match to the appropriate enum variants here instead?

or, thinking about my previous comment - maybe we can do some runtime initialization with finding classes in the `Dalamud.Interface.Internal.Windows.Data` namespace with an attribute that contains the window name & any shortcuts? i.e.:
```csharp
[DataKind("DTR Bar", new string[]{"dtr"})]
internal class DtrBarWidget : IDataWindowWidget {
```
would love your thoughts on that idea - I'm not sure if it's entirely sane but it seems like it could be a bit nicer?

Dalamud/Utility/Util.cs
```diff
@@ -36,6 +37,8 @@ public static class Util
 
     private static ulong moduleStartAddr;
     private static ulong moduleEndAddr;
+    
+    public static int copyButtonIndex;
```
avafloww, line 41: this shouldn't be public api - can we make it private and maybe have an `internal static` method to reset it on `DataWindow.Draw()`?

Dalamud/Utility/Util.cs
```diff
@@ -672,6 +675,40 @@ internal static string GetRandomName()
         return names.ElementAt(rng.Next(0, names.Count() - 1)).Singular.RawString;
     }
 
+    /// <summary>
+    /// Print formatted GameObject Information to ImGui
+    /// </summary>
+    /// <param name="actor">Game Object to Display.</param>
+    /// <param name="tag">Display Tag.</param>
+    /// <param name="resolveGameData">If the GameObjects data should be resolved.</param>
+    internal static void PrintGameObject(GameObject actor, string tag, bool resolveGameData)
```
avafloww, line 684: non-actionable note/random thought: imo this is probably better in `ImGuiHelpers`, but I'd leave it here for now because there's several methods like this here already. ideally we should clean up this `Util` class because it's a bit overloaded, but because it exposes other public apis, I don't think we can do much without a major version/api bump

avafloww (commented):
nice work, thanks for taking this on! just had a few small comments here, mostly suggestions/ideas (that may or may not be reasonable) with a few small fixes.

Dalamud/Utility/Util.cs
MidoriKami: Removed with next pr update.

Solution: Have the buttons be unique based on the contents of what would be copied to clipboard.
If there are duplicate buttons, they will still trigger the copy correctly.

Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
MidoriKami: My thought was to expand the IDataWindowWidget interface to include an alias field instead, and to iterate over the known aliases to find the correct tab to be on.
```cs
string[] aliases = { "ai" };
```

Dalamud/Utility/Util.cs
MidoriKami: I do think this particular draw is pretty specific to the data window, which is why I marked it internal instead of public.
Personally I wouldn't draw gameobject data like this for my own debugging, I would want to draw data more specific to what I cared about.

Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
avafloww: fair yeah, what I suggested is probably a bit of overkill for something internal like this.

if you want to make the alias change, I'd be okay with it - I just don't really want to be matching to strings in case we decide to rename  `Object Table` to `Baguette Table` or something in the future 😄

Dalamud/Interface/Internal/Windows/Data/DataKindEnum.cs
```diff
@@ -0,0 +1,158 @@
+﻿// ReSharper disable InconsistentNaming // Naming is suppressed so we can replace '_' with ' '
```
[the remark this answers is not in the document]
MidoriKami: For my own plugins I use a `EnumLabel` attribute for setting the names.
My version also has resx localization support, which wouldn't be relevant here.

https://github.com/MidoriKami/KamiLib/blob/master/AutomaticUserInterface/Labeling/EnumLabel.cs

Then I added a extension method "GetLabel" to get the string for each enum:
https://github.com/MidoriKami/KamiLib/blob/master/Utilities/EnumExtensions.cs

You can see it in action here:
https://github.com/MidoriKami/DailyDuty/blob/master/DailyDuty/Models/Enums/ModuleName.cs

Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
KazWolfe: Agree command handling is a bit messy but also a non-issue I'd feel. We can deal with this if/when it comes up.

avafloww (dismissed):
(nothing new, just marking my review appropriately re: the change in aliases)

Dalamud/Interface/Internal/Windows/Data/TaskSchedulerWidget.cs
```diff
@@ -0,0 +1,256 @@
...
+                                                                });
+                                                        }
+                                                    });
+                                            }
+                                        });
+                                }
+                            });
+                    }
+                });
+        }
```
KazWolfe, line 169: I know this isn't your code. Just... *pls*.

KazWolfe (changes requested):
Is it feasible to move your widgets to their own namespace?

```
Windows
|-- Data
    |-- Widgets
    |   |-- ThisWidget
    |   |-- ThatWidget
    |-- IDataWindowWidget
    |-- DataWindow
```

Feels like it'd be less of an explosion of files, but can honestly go either way.

Dalamud/Interface/Internal/Windows/Data/IDataWindowWidget.cs
```diff
@@ -0,0 +1,27 @@
...
+    
+    /// <summary>
+    /// Gets or sets a value indicating whether this data window module is ready.
+    /// </summary>
+    bool Ready { get; protected set; }
+
+    /// <summary>
+    /// Loads the necessary data for this data window module.
+    /// </summary>
+    void Load();
```
KazWolfe, line 21: Can you explain your justification for a `Load` call instead of using ctors? As far as I can tell from the PR, `Load()` will be called synchronously anyways (and will block `DataWindow` construction/force reloads until everything is loaded anyways). Or, in other words, under what conditions would the "not loaded" text actually be used?

Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
MidoriKami: This can easily be addressed in the future if needed due to the implementation of the Interface.

I'll resolve this for now.

Dalamud/Interface/Internal/Windows/Data/AddonWidget.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// <inheritdoc/>
+    public void Load()
+    {
+        this.Ready = true;
+    }
+
+    /// <inheritdoc/>
+    public void Draw()
+    {
+        var gameGui = Service<GameGui>.Get();
```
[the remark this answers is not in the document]
MidoriKami: Ah that's a good point, I can modify each of these widgets to grab the services in load instead.

Would you like me to do this?

Dalamud/Interface/Internal/Windows/Data/TaskSchedulerWidget.cs
MidoriKami: I don't know what you are asking me to address with this comment.

Dalamud/Interface/Internal/Windows/Data/IDataWindowWidget.cs
MidoriKami: There is a "Force Reload" button in the UI that calls "Load" again on each of the modules.
The current version of the data window uses load to get the server opcodes.
I simply shifted this behavior into the widgets themselves.

Load could be handy to address one of your other comments regarding service getters.

https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Interface/Internal/Windows/DataWindow.cs#L1848-L1857

Dalamud/Interface/Internal/Windows/Data/TaskSchedulerWidget.cs
KazWolfe: Nothing, just flagging it in general. I'll resolve this but I do want to draw maintainer attention to the fact that this exists.

MidoriKami:
> Is it feasible to move your widgets to their own namespace?
> 
> ```
> Windows
> |-- Data
>     |-- Widgets
>     |   |-- ThisWidget
>     |   |-- ThatWidget
>     |-- IDataWindowWidget
>     |-- DataWindow
> ```
> 
> Feels like it'd be less of an explosion of files, but can honestly go either way.

I followed the style of what was in the PluginInstaller folder, I think this is a reasonable change.
I will do this now.

Dalamud/Interface/Internal/Windows/Data/IDataWindowWidget.cs
KazWolfe: Right, I suppose my question is more why use `Load()` over a constructor and just re-instantiate the objects when Force Reload is pressed.

Again, I'm not convinced you need to make this change, just that I want decisions like this on record lest we come back six months later and ask "why did we do it this way" without the appropriate context.

Dalamud/Interface/Internal/Windows/Data/AddonWidget.cs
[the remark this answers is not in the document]
KazWolfe: I *think* so? But I want to wait for goat to confirm how services should be used first; there are some questions around our IoC stuff in general right now that need to be answered.

Dalamud/Interface/Internal/Windows/Data/IDataWindowWidget.cs
MidoriKami: I think there's value to being able to define a constructor that can setup some things that are more complex, and then having load be the things that might need to be reloaded.

I use this design pattern in DailyDuty, for example, I will parse lumina for specific data in the constructor, and then in the Load method, which is triggered when the user logs in, is when I load the configuration files.

Having the flexibility is worth the awkwardness in my opinion.

Dalamud/Interface/Internal/Windows/Data/AddonWidget.cs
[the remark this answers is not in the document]
KazWolfe: Per goat, this shouldn't matter. Should be fine as-is.

MidoriKami:
Is there anything further I need to do for this pr?

KazWolfe (approved):

PR #1259, opened 2023-06-24
Add functions to get High Resolution icons

MidoriKami:
Also performed some minor style consistency changes.

These changes were made in a non-abi breaking way, and todo's were added noting that functions should be removed in API9

### DIFF
--- a/Dalamud/Data/DataManager.cs
+++ b/Dalamud/Data/DataManager.cs
@@ -30,6 +30,7 @@ namespace Dalamud.Data;
 public sealed class DataManager : IDisposable, IServiceType
 {
     private const string IconFileFormat = "ui/icon/{0:D3}000/{1}{2:D6}.tex";
+    private const string HighResolutionIconFileFormat = "ui/icon/{0:D3}000/{1}{2:D6}_hr1.tex";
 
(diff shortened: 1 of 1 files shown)


KazWolfe (approved):

KazWolfe:
Something to note here: There may be causes where a `_hr1` variant of an icon is not available. It's not worth dealing in this PR (mostly because fallback code gets [annoyingly complicated](https://github.com/KazWolfe/XIVDeck/blob/203abb50688eaa4c0af51b0ff0cb0ef8ce336517/FFXIVPlugin/Game/Managers/IconManager.cs#L46)), but this may cause annoyances for someone using the new APIs. Caveat emptor!

MidoriKami:
> Something to note here: There may be causes where a `_hr1` variant of an icon is not available. It's not worth dealing in this PR (mostly because fallback code gets [annoyingly complicated](https://github.com/KazWolfe/XIVDeck/blob/203abb50688eaa4c0af51b0ff0cb0ef8ce336517/FFXIVPlugin/Game/Managers/IconManager.cs#L46)), but this may cause annoyances for someone using the new APIs. Caveat emptor!

All textures returned are nullable, devs should be checking for null anyways.

PR #1261, opened 2023-06-24
Add IBuddyList

MidoriKami:
Something of note is I added `[MemberNotNullWhen(...)]` attributes to the boolean checks, this enables some very nice syntactical sugar for the dev.

image

### DIFF
--- a/Dalamud/Game/ClientState/Buddy/BuddyList.cs
+++ b/Dalamud/Game/ClientState/Buddy/BuddyList.cs
@@ -60,11 +60,13 @@ public sealed partial class BuddyList : IServiceType, IBuddyList
     /// <summary>
     /// Gets a value indicating whether the local player's companion is present.
     /// </summary>
+    [Obsolete("Use CompanionBuddy != null", false)]
     public bool CompanionBuddyPresent => this.CompanionBuddy != null;
 
     /// <summary>
     /// Gets a value indicating whether the local player's pet is present.
     /// </summary>
+    [Obsolete("Use PetBuddy != null", false)]
     public bool PetBuddyPresent => this.PetBuddy != null;
 
     /// <summary>
--- a/Dalamud/Game/ClientState/Buddy/BuddyList.cs
+++ b/Dalamud/Game/ClientState/Buddy/BuddyList.cs
@@ -5,6 +5,7 @@ using System.Runtime.InteropServices;
 
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
+using Dalamud.Plugin.Services;
 using Serilog;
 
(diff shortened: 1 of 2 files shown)


Dalamud/Plugin/Services/IBuddyList.cs
```diff
@@ -0,0 +1,73 @@
...
+/// It does not include the local player.
+/// </summary>
+public interface IBuddyList : IReadOnlyCollection<BuddyMember>
+{
+    /// <summary>
+    /// Gets the amount of battle buddies the local player has.
+    /// </summary>
+    public int Length { get; }
+    
+    /// <summary>
```
KazWolfe, line 19: Unsure if either of these should be part of the public API, given that they're just checking the relevant `Buddy` for `null` already.

Dalamud/Plugin/Services/IBuddyList.cs
```diff
@@ -0,0 +1,73 @@
...
+    /// </summary>
+    [MemberNotNullWhen(true, "CompanionBuddy")]
+    public bool CompanionBuddyPresent { get; }
+    
+    /// <summary>
+    /// Gets a value indicating whether the local player's pet is present.
+    /// </summary>
+    [MemberNotNullWhen(true, "PetBuddy")]
+    public bool PetBuddyPresent { get; }
+
```
KazWolfe, line 30: Unsure if either of these should be part of the public API, given that they're just checking the relevant `Buddy` for null already.

MidoriKami: That's fair, in normal use I would more likely just check for null like this:
image

KazWolfe: Hm. Unless there's a way for these objects to be not-null but still not be "present" (previously spawned but no longer around?), let's remove these from the interface and mark them deprecated in the base service.

Dalamud/Game/ClientState/Buddy/BuddyList.cs
```diff
@@ -16,7 +17,10 @@ namespace Dalamud.Game.ClientState.Buddy;
 [PluginInterface]
 [InterfaceVersion("1.0")]
 [ServiceManager.BlockingEarlyLoadedService]
-public sealed partial class BuddyList : IServiceType
+#pragma warning disable SA1015
```
karashiiro, line 20: This is fixed in https://github.com/DotNetAnalyzers/StyleCopAnalyzers/issues/3641, we should pull in the latest prerelease version of the analyzer. Not the responsibility of this PR, though.

avafloww (approved):

PR #1262, opened 2023-06-24
Add ICondition (v9)

MidoriKami:
Includes breaking changes.

### DIFF
--- a/Dalamud/Game/ClientState/Conditions/Condition.cs
+++ b/Dalamud/Game/ClientState/Conditions/Condition.cs
@@ -2,6 +2,7 @@ using System;
 
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
+using Dalamud.Plugin.Services;
 using Serilog;
 
 namespace Dalamud.Game.ClientState.Conditions;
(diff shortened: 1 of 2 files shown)


Dalamud/Game/ClientState/Conditions/Condition.cs
```diff
@@ -12,13 +13,16 @@ namespace Dalamud.Game.ClientState.Conditions;
...
+#pragma warning disable SA1015
+[ResolveVia<ICondition>]
+#pragma warning restore SA1015
+public sealed partial class Condition : IServiceType, ICondition
 {
     /// <summary>
-    /// The current max number of conditions. You can get this just by looking at the condition sheet and how many rows it has.
+    /// Gets the current max number of conditions. You can get this just by looking at the condition sheet and how many rows it has.
     /// </summary>
     public const int MaxConditionEntries = 104;
```
avafloww, line 24: can you remove this field, since it's being renamed to `MaxEntries` and this PR is targeting v9 anyways?

avafloww (changes requested):

Dalamud/Game/ClientState/Conditions/Condition.cs
MidoriKami: Ah I forgot to make it Internal, it is used elsewhere in dalamud, specifically in the data window.

avafloww: Would it make more sense to just have the data window use the `MaxEntries` from the service instance instead of a separate constant? Looks like it already retrieves a `Service<Condition>` so that might be a bit cleaner imo.

MidoriKami: Dalamud/Interface/Internal/Windows/Data/Widgets/ConditionWidget.cs
```cs
    /// <inheritdoc/>
    public void Draw()
    {
        var condition = Service<Condition>.Get();

#if DEBUG
        ImGui.Text($"ptr: 0x{condition.Address.ToInt64():X}");
#endif

        ImGui.Text("Current Conditions:");
        ImGui.Separator();

        var didAny = false;

        for (var i = 0; i < Condition.MaxConditionEntries; i++)
        {
            var typedCondition = (ConditionFlag)i;
            var cond = condition[typedCondition];

            if (!cond) continue;

            didAny = true;

            ImGui.Text($"ID: {i} Enum: {typedCondition}");
        }
```

If `Service<Condition>.Get();` returned the Interface I could remove that constant field entirely.

MidoriKami: Oh, and the constant field is still required at least privately, because that field is used to define the size of the array in Condition.cs

avafloww: ah ok, I see. thanks for the explanation!

avafloww (approved):

PR #1266, opened 2023-06-24
Add IKeyState

MidoriKami:
There's an alternative PR #1267 that includes a breaking change by changing the return value of `GetValidVirtualKeys` to `IEnumerable<VirtualKey>` instead as a more generic interface implementation.


avafloww:
Closing in favour of #1267 - thanks!

PR #1269, opened 2023-06-25
Add IsTargetable to GameObject

MidoriKami:
I very often end up having to cast a Dalamud GameObject to a ClientStructs GameObject just to check if it is targetable.

This should fix that.


MidoriKami:
Interestingly the ide didn't complain about not using `this` for `Struct->GetIsTargetable()`

KazWolfe (approved):

PR #1272, opened 2023-06-25
Add ITargetManager

MidoriKami:
It's my understanding that the terminology of "actor" to mean "gameobject" is no longer desireable.

I have adjusted the comments to refer to GameObject's instead of Actor's


Dalamud/Plugin/Services/ITargetManager.cs
```diff
@@ -0,0 +1,124 @@
...
+
+    /// <summary>
+    /// Clears the previous target.
+    /// </summary>
+    public void ClearPreviousTarget();
+
+    /// <summary>
+    /// Clears the soft target.
+    /// </summary>
+    public void ClearSoftTarget();
```
KazWolfe, line 123: Given all these are basically explicitly defined setters, are they explicitly needed? Can we coalesce the API down to just using the property setter?

(not a question that needs to be answered by the PR author proper - just generally floating it.)

KazWolfe (approved):

Dalamud/Plugin/Services/ITargetManager.cs
MidoriKami: There is merit to at least some of the functions being visible, I don't know how else you would allow this interaction:
image

KazWolfe: I _think_ setting it to just `null` would work?

MidoriKami: Ah! Yup, exactly right you are. In that case, I'm very much in favor of removing the duplicates from the interface, and leaving them in the implementation.

MidoriKami:
I was having some troubles with adding IFlyText to v9 instead of master, and that seems to have exploded my git.

PR #1278, opened 2023-06-25
Add IFlyTextGui (v9)

MidoriKami:
Includes breaking changes, by moving delegate into interface.


KazWolfe (changes requested):
Ditto - this service is simple enough that I feel it makes sense to scope it now.

KazWolfe (approved):

PR #1286, opened 2023-06-25
Add IFramework (v9)

MidoriKami:
Of note, I changed the Framework destroy stuff to be private, I can't think of any valid uses a plugin might have for those.


KazWolfe (approved):

KazWolfe:
Merging as-is so we have the interface defined.

We should consider moving Framework over to a scoped service thanks to the `Framework.Update` calls, but this may be good to target alongside `Framework.PreUpdate`/`Framework.PostUpdate` rather than as part of this PR.

PR #1296, opened 2023-06-28
Add a simple network monitor to xldata.

Ottermandias:


Dalamud/Interface/Internal/Windows/Data/Widgets/NetworkMonitorWidget.cs
```diff
@@ -0,0 +1,172 @@
...
+            catch
+            {
+                this.trackedOpCodes = null;
+            }
+        }
+    }
+
+    private void OnNetworkMessage(nint dataPtr, ushort opCode, uint sourceActorId, uint targetActorId, NetworkMessageDirection direction)
+    {
+        if (this.trackedOpCodes == null || this.trackedOpCodes.IsMatch(this.OpCodeToString(opCode)))
```
MidoriKami, line 149: Can we get an option to invert this match?
Its simple enough to write a regex to show 790 or 709 `(790|709)` but it becomes a lot more complex to *hide* 790 and 709.
Maybe there's a simple way to write the regex for this.
Often I don't know what opcode I am looking for and just want to trigger the event I need, and then narrow down the opcodes that I know I don't want.

wolfcomp: Negative Lookahead is a thing in Regex with `(?!<clause>)` this meaning `^(?!(790|709))` should hide 709 and 790

MidoriKami: > Negative Lookahead is a thing in Regex with `(?!<clause>)` this meaning `^(?!(790|709))` should hide 709 and 790

image

That does not work.

wolfcomp: proof

C# console says that it works the same way that regex101 shows it also matching the correct match line

Dalamud/Interface/Internal/Windows/Data/Widgets/NetworkMonitorWidget.cs
```diff
@@ -0,0 +1,172 @@
...
+    /// <remarks> Add known packet-name -> packet struct size associations here to copy the byte data for such packets. </remarks>>
+    private int GetSizeFromName(string name)
+        => name switch
+        {
+            _ => 0,
+        };
+
+    /// <remarks> The filter should find opCodes by number (decimal and hex) and name, if existing. </remarks>
+    private string OpCodeToString(ushort opCode)
+        => this.opCodeDict.TryGetValue(opCode, out var pair) ? $"{pair.Item1}\0{opCode}\0{opCode:X}" : $"{opCode}\0{opCode:X}";
```
wolfcomp, line 171: Should probably change the order of `pair.Item1` and `opCode` to mage Negative Lookahead with numbers easier for filtering.

goaaats (approved):

PR #1300, opened 2023-07-05
GameConfig - Cache reflective name lookup

Caraxi:
Cache the results of the reflective name lookup, improving the performance of  `GameConfig.TryGet` functions by ~100x


nebel:
By the same logic as these dictionaries use `ConcurrentDictionary`, I wonder if the downstream dictionaries which are used in `GameConfigSection.TryGetIndex` (`indexMap`, `nameMap`) should also be concurrent? I think the chances of anything happening there are vanishingly small, but the same could be said of these enum lookup dictionaries.

(Actually `nameMap` seems to be unused and could possibly be removed instead if there's no future use case being considered,)

goaaats (approved):

PR #1305, opened 2023-07-06
Improve Search

MidoriKami:
Adds a FuzzyMatch utility

Modifies plugin installer search to use FuzzyMatch.

This allows users to match plugin names easier, we often see an issue where a plugin may be called PlayerTrack but the user searches for "Player Track" and no results show.


goaaats (approved):

goaaats:
Thanks! I made the FuzzyMatcher class internal, so that it doesn't become public API.

goaaats (approved):

PR #1327, opened 2023-07-29
Use ReShade Addon functionality or VTable replacement for IDXGISwapChain hooking

Soreepeong:
Tested with ReShade + Discord overlay.

If ReShade addon initialization function returns success, then it will use ReShade's `reshade_overlay` event; otherwise, it will replace the VTable of the game's swap chain object with our implementation supplied for `Present`.

Either way, `ResizeBuffers` uses vtable replacement.

- [ ] Test RTSS
- [ ] Test OBS
- [ ] Test Discord streaming
- [ ] Sort out ReShade licensing


Dalamud/Hooking/ObjectVTableHook.cs
```diff
@@ -0,0 +1,167 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace Dalamud.Hooking;
+
+/// <summary>
+/// Manages a hook that works by replacing the vtable of target object.
+/// </summary>
+public unsafe class ObjectVTableHook : IDisposable
```
Haselnussbomber, line 9: I randomly peeked into this PR and noticed something: `ObjectVTableHook` and `ObjectVTableHook<TVTableEnum>` are public. Since it creates a new VTable in its constructor, when used in plugins, you could override another plugin's VTable hooks. To prevent that, it's probably best to make them internal. 🙂

Soreepeong: I was thinking to allow plugins to use this method of hooking too, but then again if we're to allow that, it's probably better to let Dalamud manage the hook chain. Probably better to set this internal for now and do the work to make it usable from many client plugins later.

PR #1335, opened 2023-08-02
ITextureProvider, ITextureSubstitutionProvider, TextureManager services

goaaats:
Ref-counts textures and evicts when not used

### DIFF
--- a/Dalamud/Data/DataManager.cs
+++ b/Dalamud/Data/DataManager.cs
@@ -185,15 +185,17 @@ public sealed class DataManager : IDisposable, IServiceType, IDataManager
     /// </summary>
     /// <param name="iconId">The icon ID.</param>
     /// <returns>The <see cref="TexFile"/> containing the icon.</returns>
-    /// TODO(v9): remove in api9 in favor of GetIcon(uint iconId, bool highResolution)
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(uint iconId) 
         => this.GetIcon(this.Language, iconId, false);
 
     /// <inheritdoc/>
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(uint iconId, bool highResolution)
         => this.GetIcon(this.Language, iconId, highResolution);
 
     /// <inheritdoc/>
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(bool isHq, uint iconId)
     {
         var type = isHq ? "hq/" : string.Empty;
@@ -206,11 +208,12 @@ public sealed class DataManager : IDisposable, IServiceType, IDataManager
     /// <param name="iconLanguage">The requested language.</param>
     /// <param name="iconId">The icon ID.</param>
     /// <returns>The <see cref="TexFile"/> containing the icon.</returns>
-    /// TODO(v9): remove in api9 in favor of GetIcon(ClientLanguage iconLanguage, uint iconId, bool highResolution)
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(ClientLanguage iconLanguage, uint iconId)
         => this.GetIcon(iconLanguage, iconId, false);
 
     /// <inheritdoc/>
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(ClientLanguage iconLanguage, uint iconId, bool highResolution)
     {
         var type = iconLanguage switch
@@ -231,11 +234,12 @@ public sealed class DataManager : IDisposable, IServiceType, IDataManager
     /// <param name="type">The type of the icon (e.g. 'hq' to get the HQ variant of an item icon).</param>
     /// <param name="iconId">The icon ID.</param>
     /// <returns>The <see cref="TexFile"/> containing the icon.</returns>
-    /// TODO(v9): remove in api9 in favor of GetIcon(string? type, uint iconId, bool highResolution)
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(string? type, uint iconId)
         => this.GetIcon(type, iconId, false);
 
     /// <inheritdoc/>
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(string? type, uint iconId, bool highResolution)
     {
         var format = highResolution ? HighResolutionIconFileFormat : IconFileFormat;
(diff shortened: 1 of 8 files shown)


Dalamud/Interface/Internal/TextureManager.cs
```diff
@@ -0,0 +1,458 @@
...
+    /// to render the icon.
+    /// </returns>
+    public TextureManagerTextureWrap? GetIcon(uint iconId, ITextureProvider.IconFlags flags = ITextureProvider.IconFlags.ItemHighQuality, ClientLanguage? language = null, bool keepAlive = false)
+    {
+        var hiRes = flags.HasFlag(ITextureProvider.IconFlags.HiRes);
+        
+        // 1. Item
+        var path = FormatIconPath(
+            iconId,
+            flags.HasFlag(ITextureProvider.IconFlags.ItemHighQuality) ? "hq/" : string.Empty,
```
goaaats, line 85: At the moment, items will not fall back to non-hq if there is no hq variant and the hq flag is set. Do we want that?

Dalamud/Interface/Internal/TextureManager.cs
```diff
@@ -0,0 +1,458 @@
...
+        
+        if (info is { Wrap: not null })
+            return info;
+
+        if (refresh)
+        {
+            byte[]? interceptData = null;
+            this.InterceptTexDataLoad?.Invoke(path, ref interceptData);
+
+            // TODO: Do we also want to support loading from actual fs here? Doesn't seem to be a big deal, collect interest
```
goaaats, line 203: We could easily change this API to also support loading images(png) or .tex from the FS. Should that be a goal?

KazWolfe: Yeah, I think loading in tex from the filesystem would be good if it's simple. I can think of some edge cases where someone has already handled custom resolutions (e.g. loading from a specific Penumbra collection).

Dalamud/Interface/Internal/TextureManager.cs
```diff
@@ -0,0 +1,458 @@
...
+        if (refresh && info.KeepAliveCount == 0)
+            info.LastAccess = DateTime.Now;
+        
+        if (info is { Wrap: not null })
+            return info;
+
+        if (refresh)
+        {
+            byte[]? interceptData = null;
+            this.InterceptTexDataLoad?.Invoke(path, ref interceptData);
```
KazWolfe, line 201: Is there a way we can explicitly override this, e.g. if we absolutely want the vanilla textures for some reason?

Is that something we want to support?

KazWolfe: Also, there's no case where `refresh` would be false when we actually want to *get* a texture, correct?

Dalamud/Interface/Internal/TextureManager.cs
```diff
@@ -0,0 +1,458 @@
...
+    public int Height => this.height ??= this.manager.GetInfo(this.path).Wrap!.Height;
+
+    /// <summary>
+    /// Gets a value indicating whether or not this wrap has already been disposed.
+    /// If true, the handle may be invalid.
+    /// </summary>
+    internal bool IsDisposed { get; private set; }
+
+    /// <inheritdoc/>
+    public void Dispose()
```
KazWolfe, line 448: To make sure I understand: we should still be disposing our textures, but we have an in-place backstop on unload in case we forget?

Dalamud/Interface/Internal/TextureManager.cs
KazWolfe: I think no fallback for HQ request is fine. We _may_ consider a fallback for HR, but that may be pretty niche and not worth doing.

Dalamud/Interface/Internal/TextureManager.cs
goaaats: Yes, you should, but it will also be done for you if your plugin unloads, and the texture is unloaded if it hasn't been drawn for a while so impact is minimized if you don't.

Dalamud/Interface/Internal/TextureManager.cs
goaaats: > Is that something we want to support?
We probably want a mechanism for it, but it's not in this PR. Will do something about it at a later date.

Dalamud/Interface/Internal/TextureManager.cs
```diff
@@ -197,29 +246,75 @@ internal TextureInfo GetInfo(string path, bool refresh = true)
...
-                var type = typeof(TexFile);
-                type.GetProperty("Data", BindingFlags.NonPublic | BindingFlags.Instance)!.GetSetMethod()!
-                    .Invoke(file, new object[] { interceptData });
-                type.GetProperty("Reader", BindingFlags.NonPublic | BindingFlags.Instance)!.GetSetMethod()!
-                    .Invoke(file, new object[] { new LuminaBinaryReader(file.Data) });
-                file.LoadFile();
+                // We want to load this from the disk, probably, if the path has a root
+                // Not sure if this can cause issues with e.g. network drives, might have to rethink
+                // and add a flag instead if it does.
+                if (Path.IsPathRooted(path))
```
KazWolfe, line 267:
```cs
using System;
using System.IO;

var path = "\\\\10.0.0.1\\files\\mods\\mymod.tex";
var rooted = Path.IsPathRooted(path);

Console.WriteLine($"Path is: {path}. Rooted: {rooted}");
// Path is: \\10.0.0.1\files\mods\mymod.tex. Rooted: False
```

That said, if you're using UNC, you get what you deserve.

Dalamud/Data/DataManager.cs
```diff
@@ -185,15 +185,17 @@ public bool FileExists(string path)
     /// </summary>
     /// <param name="iconId">The icon ID.</param>
     /// <returns>The <see cref="TexFile"/> containing the icon.</returns>
-    /// TODO(v9): remove in api9 in favor of GetIcon(uint iconId, bool highResolution)
+    [Obsolete("Use ITextureProvider instead")]
     public TexFile? GetIcon(uint iconId) 
         => this.GetIcon(this.Language, iconId, false);
```
KazWolfe, line 190: Things that return `TexFile` are technically being removed without a replacement, no? Pedantics, but might be worth calling that out explicitly or mentioning that plugins that need raw Tex files are now on their own (though this could be in the dev patch notes as well).

KazWolfe (dismissed):

Dalamud/Data/DataManager.cs
goaaats: Added a function to allow path lookups for icons

KazWolfe (approved):

Dalamud/Interface/Internal/TextureManager.cs
```diff
@@ -480,6 +499,10 @@ public TextureManagerPluginScoped(TextureManager textureManager)
         return wrap;
     }
 
+    /// <inheritdoc/>
+    public string? GetIconPath(uint iconId, ITextureProvider.IconFlags flags = ITextureProvider.IconFlags.HiRes, ClientLanguage? language = null)
```
wolfcomp, line 503: This function does the exact same as the one above no?

goaaats: No. This one just exists as a convenience function, if you need to do something else with an icon(e.g. load a tex file to send somewhere).

PR #1340, opened 2023-08-06
IHookProvider service, no static hook creation, fix SignatureHelper

goaaats:


Dalamud/Utility/Signatures/SignatureHelper.cs
```diff
@@ -24,7 +26,8 @@ public static class SignatureHelper
     /// </summary>
     /// <param name="self">The object to initialise.</param>
     /// <param name="log">If warnings should be logged using <see cref="PluginLog"/>.</param>
-    public static void Initialise(object self, bool log = true)
+    /// <returns>Collection of created IDalamudHooks.</returns>
+    internal static IEnumerable<IDalamudHook> Initialise(object self, bool log = true)
```
KazWolfe, line 30: Consider renaming to `Initialize` to match US EN as used in the rest of the project.

KazWolfe, line 30: This method is currently also used to initialize actual signatures as well, e.g.:

```cs
public class GameFunctions {
    private delegate byte IsQuestCompletedDelegate(ushort questId);

    [Signature("E8 ?? ?? ?? ?? 41 88 84 2C")]
    private readonly IsQuestCompletedDelegate? _isQuestCompleted = null;

    public GameFunctions() {
        SignatureHelper.Initialise(this);
    }
}
```

Given these aren't hooks (and, as such, don't need to be tracked), how are we going to be handling these cases? A precursory read suggests that they'll be created anyways (and just won't be returned), but documentation does not particularly support this point - see other review comments.

Dalamud/Plugin/Services/IHookProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        /// </summary>
+        MinHook,
+    }
+    
+    /// <summary>
+    /// Initialize <see cref="Hook{T}"/> members decorated with the <see cref="SignatureAttribute"/>.
+    /// Errors for fallible signatures will be logged.
+    /// </summary>
+    /// <param name="self">The object to initialise.</param>
+    public void InitializeFromAttributes(object self);
```
KazWolfe, line 40: Will this method also be responsible for initializing and setting plain delegates? If so, consider documenting this as doing more than just hooks.

I'm also not particularly a fan of delegate management being handled by `IHookProvider` semantically, but I suppose it makes enough sense (and likely isn't worth the effort to make into a separate system).

KazWolfe (changes requested):

Dalamud/Plugin/Services/IHookProvider.cs
Haselnussbomber: I believe it should absolutely only handle hooks and not the other stuff `SignatureHelper` does. It would just result in another inconsistent API, which will require another refactor in the future.

For delegates, we could introduce a new `MemoryHelper` method:

```cs
public static T GetDelegateForSignature<T>(string signature) where T : Delegate
{
    var scanner = Service<SigScanner>.Get();
    var address = scanner.ScanText(signature);
    return Marshal.GetDelegateForFunctionPointer<T>(address);
}
```

Then, plugins can use it like this:

```cs
using Dalamud.Memory;

public class GameFunctions
{
    private delegate byte IsQuestCompletedDelegate(ushort questId);
    private readonly IsQuestCompletedDelegate IsQuestCompleted = MemoryHelper.GetDelegateForSignature<IsQuestCompletedDelegate>("E8 ?? ?? ?? ?? 41 88 84 2C");
}
```

It's an easy to understand alternative.  
Though the line might get a little long, it doesn't need an attribute and there is no need to call `SignatureHelper.Initialise(this)` the constructor.
It also throws, just like `Hook<T>.FromSignature()`, if the signature is not found, making it undeniably clear that the signature is outdated.

Besides delegates, `SignatureHelper` can also fill a fields **initial** value. But do we *really* need to support this? It's easy enough to read it in the constructor, to create a property with a getter, or to use Lazy:

```cs
public Lazy<nint> SomeAddress => new(() => MemoryHelper.Read<nint>(Service.SigScanner.GetStaticAddressFromSig(sig)));
```

🤷‍♂️

Dalamud/Hooking/Internal/HookProviderPluginScoped.cs
```diff
@@ -0,0 +1,99 @@
...
+    public Hook<T> FromAddress<T>(IntPtr procAddress, T detour, IHookProvider.HookBackend backend = IHookProvider.HookBackend.Automatic) where T : Delegate
+    {
+        var hook = Hook<T>.FromAddress(procAddress, detour, backend == IHookProvider.HookBackend.MinHook);
+        this.trackedHooks.Add(hook);
+        return hook;
+    }
+
+    /// <inheritdoc/>
+    public Hook<T> FromSignature<T>(string signature, T detour, IHookProvider.HookBackend backend = IHookProvider.HookBackend.Automatic) where T : Delegate
+        => this.FromAddress(this.scanner.ScanText(signature), detour);
```
Haselnussbomber, line 83: The `backend` parameter is not passed here.

Dalamud/Plugin/Services/IHookProvider.cs
goaaats: I generally agree, but I think that crippling an API on purpose for correctness' sake is a little unfortunate. Not 100% sure what to do here. Looking through codesearch, hooks, delegates and address modes are used quite frequently in plugins. I would honestly tend to just leaving it like this.

Maybe the service should be renamed to IGameInteropProvider, and docs should be improved, so that it's clearer that it's not just hooks?

KazWolfe: Docs plus a better name would be nice. Maybe `INativeInteropService` or similar? These methods can be used for more than just game hooks, after all.

Haselnussbomber: I think `INativeInteropService` is fine. It just sounds a bit like it should be the home of the functions in `MemoryHelper`. Not sure if that's a good thing. 🤔

Anyhow, if you use a broader name, I’d suggest that the method names should reflect their usage then, i.e. renaming `FromSignature` to `HookFromSignature` and so on, so that they don't conflict with other things that might be created from a signature in the future, like `DelegateFromSignature`. 👀

goaaats: I addressed both of this, it's a bit more verbose now, but it does feel better overall IMO.

KazWolfe: @Haselnussbomber w/r/t `DelegateFromSignature`: one can already do that from SigScanner pretty reliably. We can add to this later, but it's not like we're removing any functionality; I think we're good.

KazWolfe:
LGTM, new API surface documented here for quick reference:

Plugins wishing to create hooks should now use `IGameInteropProvider` as their service, and call one of `HookFrom{Whatever}` based on their use case.

The `[Signature]` attribute still works as expected, but the initialization call is now going to be `IGameInteropProvider#InitializeFromAttributes(this)`. This will initialize *both* hooks and defined delegates.

Dalamud/Utility/Signatures/SignatureHelper.cs
```diff
@@ -24,7 +26,8 @@ public static class SignatureHelper
     /// </summary>
     /// <param name="self">The object to initialise.</param>
```
KazWolfe, line 27: Spelling to catch

Dalamud/Plugin/Services/IGameInteropProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+        
+        /// <summary>
+        /// Use MinHook.
+        /// You should never have to use this without talking to us first.
+        /// </summary>
+        MinHook,
+    }
+    
+    /// <summary>
+    /// Initialize <see cref="Hook{T}"/> members decorated with the <see cref="SignatureAttribute"/>.
```
KazWolfe, line 36: Explicitly specify that this also will initialize any delegates that are defined using the `[Signature]` attribute as well. At present, wording can be interpreted that this only does hooks.

KazWolfe (changes requested):

Dalamud/Plugin/Services/IGameInteropProvider.cs
```diff
@@ -0,0 +1,97 @@
...
+    /// Errors for fallible signatures will be logged.
+    /// </summary>
+    /// <param name="self">The object to initialize.</param>
+    public void InitializeFromAttributes(object self);
+    
+    /// <summary>
+    /// Creates a hook by rewriting import table address.
```
KazWolfe, line 43: Stray copy-paste from FromImport below?

KazWolfe (changes requested):

Dalamud/Plugin/Services/IGameInteropProvider.cs
Haselnussbomber: goat asked me what I would put here. I'm bad at this, but probably something along the lines of:
> Creates a hook by replacing the original address with an address pointing to a newly created jump to the detour.

Just thought I'd mention it here, since it kinda got burried in Discord. 🤷‍♂️🙂

PR #1341, opened 2023-08-08
Plugin Log Service

KazWolfe:
- Revert PluginLog to how it used to be.
- Create new IPluginLog with API surface
- Create ScopedPluginLogService handling the core


goaaats (approved):

KazWolfe:
Rebased and cleaned, API surface should be good to go.

For now, I don't want to mark `PluginLog` as obsolete because it'll generate far too much noise for developers. We'll take care of that in API 9, however.

Notably, `PluginLog<T>` has been removed as well - this will be re-done by exposing `Logger` at some point in the future after we have proper formatting in place for use with enriched log messages.

PR #1344, opened 2023-08-09
[master] Update ClientStructs

bleatbot:


KazWolfe:
Breaking changes to:
- AgentIKD (though this should be safe, it's new and no plugins use it)
- AgentLobby (ulong -> ushort), but this should not affect semantics.

KazWolfe (approved):
IKD changes ok. ulong was new, so ok.

PR #1348, opened 2023-08-16
Add support for HTTP2 connections

KazWolfe:
This pull request aims to allow the shared HTTP client to upgrade connections to HTTP2, depending on the remote server (and the system the code is running on). This is basically a direct port of what MS docs [tell us to do](https://learn.microsoft.com/en-us/dotnet/core/extensions/httpclient-http3#httpclient-settings).

n.b. this appears to work at a surface level and the very quick tests I threw at it, but given .NET HTTP's fragility, we might want this to stay in staging for a while.


goaaats:
Need to test in wine. We explicitly disable HTTP3 in Dalamud.Boot since it doesn't support QUIC APIs.

PR #1351, opened 2023-08-20
[v9] Make Wine Checks More Resilient

KazWolfe:
This pull request aims to improve our wine check resilience to better handle some edge cases:

* Remove checks looking for `HKCU:\Software\Wine` and `HKLM:\Software\Wine`, as certain programs will create these registry paths (even on Windows!) to improve Linux compatibility.
* Add checks for the `wine_server_call` and `wine_unix_to_nt_file_name` exports, which are *not* hidden by the `HideWineExports` option.
    * The other listed Wine exports will all be hidden for licensing reasons for Linux users.
* Rename `Util.IsLinux()` to `Util.IsWine()` to prevent confusion with macOS detection.

Generally speaking, all *supported* users will already be covered by the `XL_WINEONLINUX` environment variable already. However, retaining export checks does allow us a bit more flexibility in the future.

### DIFF
--- a/Dalamud.Boot/dllmain.cpp
+++ b/Dalamud.Boot/dllmain.cpp
@@ -133,8 +133,8 @@ DWORD WINAPI InitializeImpl(LPVOID lpParam, HANDLE hMainThreadContinue) {
     // ============================== VEH ======================================== //
 
     logging::I("Initializing VEH...");
-    if (utils::is_running_on_linux()) {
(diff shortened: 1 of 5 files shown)


KazWolfe:
The `Util.GetHostPlatform()` check is optional and can be pulled from this PR if we decide we don't want it.

goaaats (approved):

PR #1361, opened 2023-08-27
Add AddonLifecycle Service

MidoriKami:
The purpose of this service is to provide an easy entryway to accessing addon data without needing each plugin to hook individual addon setup and finalize calls.

This service hooks the main game system responsible for actually performing the setup and finalize for addons. Each addon is setup with the addon name in string form, making it easy for devs to receive events for specific addons.

Example usage:
```cs
    public override void AddonPreSetup(IAddonLifecycle.AddonArgs addonInfo)
    {
        if (addonInfo.AddonName != "WeeklyPuzzle") return;

        Data.FauxHollowsCompletions += 1;
        DataChanged = true;
    }
```

This approach was originally discovered by @Caraxi , and has been in use by SimpleTweaks and DailyDuty for multiple months now.

### DIFF
--- a/Dalamud/Game/AddonLifecycle/AddonLifecycle.cs
+++ b/Dalamud/Game/AddonLifecycle/AddonLifecycle.cs
@@ -8,7 +8,7 @@ using Dalamud.Logging;
 using Dalamud.Plugin.Services;
 using FFXIVClientStructs.FFXIV.Component.GUI;
 
-namespace Dalamud.Game.AddonLifetime;
+namespace Dalamud.Game.AddonLifecycle;
 
 /// <summary>
 /// This class provides events for in-game addon lifecycles.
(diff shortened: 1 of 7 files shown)


KazWolfe (changes requested):
Hi! Thank you for your pull request. I've left a couple requests for changes to look at before merging this.

Dalamud/Game/AddonLifecycle/AddonLifecycle.cs
```diff
@@ -0,0 +1,168 @@
...
+
+    private nint OnAddonSetup(AtkUnitBase* addon)
+    {
+        try
+        {
+            this.AddonPreSetup?.Invoke(new IAddonLifecycle.AddonArgs { Addon = (nint)addon });
+        }
+        catch (Exception e)
+        {
+            PluginLog.Error(e, "[AddonLifecycle] Exception in OnAddonSetup pre-setup invoke.");
```
KazWolfe, line 74: Please use `ModuleLog` here. e.g. at the top of the class:

```cs
private static readonly ModuleLog Log = new("AddonLifecycle");
```

and then call `Log.Thing()`.

Applies to all PluginLog uses.

KazWolfe (changes requested):
Looks good and seems to test well, just the logging change and we should be good to go!

KazWolfe (approved):

PR #1374, opened 2023-09-09
Korean support

lee-jinhwan:
Add a Korean options to support Korean FFXIV.

### DIFF
--- a/Dalamud.Injector/EntryPoint.cs
+++ b/Dalamud.Injector/EntryPoint.cs
@@ -349,6 +349,10 @@ namespace Dalamud.Injector
             {
                 clientLanguage = ClientLanguage.French;
             }
+            else if (languageStr[0..(len = Math.Min(languageStr.Length, (key = "korean").Length))] == key[0..len])
+            {
+                clientLanguage = ClientLanguage.Korean;
+            }
             else if (int.TryParse(languageStr, out var languageInt) && Enum.IsDefined((ClientLanguage)languageInt))
             {
                 clientLanguage = (ClientLanguage)languageInt;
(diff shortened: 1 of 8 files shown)


KazWolfe:
Hi,

First off, thank you for your contribution! Unfortunately, I'm not particularly sure we can accept this pull request due to material differences in the global and Korean FINAL FANTASY XIV clients.

Put simply, Dalamud tracks the global client, and makes certain assumptions that certain features are present. This is a deep dependency, seeing as we depend on [FFXIVClientStructs](https://github.com/aers/FFXIVClientStructs), which also documents and targets the global client. Certain subsystems referenced in ClientStructs or Dalamud's API surfaces may not necessarily exist on the Korean client, meaning those systems may break unexpectedly when run against the Korean client. Neither Dalamud nor ClientStructs are meant to be run on older versions of the game - any such attempts would often require going back and compiling the specific version of Dalamud and CS for the game version at that time.

Because of this, any users attempting to run modern versions of Dalamud against the Korean client will have a less-than-stellar experience, with it possibly not even working reliably at all depending on what changes have taken place in more recent versions of the global game client. Similarly, since the Dalamud team is unable to receive a copy of the Korean client (much less actually log in and play using that client), our ability to test the Korean client for compatibility is effectively nonexistent.

We *can* definitely merge this pull request, but I'm not sure we would be able to say with any degree of confidence that we'd be able to support the Korean client, either now or moving forward (at least, not without significant changes to much larger portions of both Dalamud and ClientStructs). We've definitely [merged pull requests](https://github.com/goatcorp/Dalamud/pull/597) in the past to make certain things easier for forks targeting non-global Dalamud, but this alone is not enough to actually add support.

It would help if we could know a bit more about what you're hoping to achieve, and how you'd like to proceed given this information.

lee-jinhwan:
@KazWolfe 
Hi,
First of all, thank you for your answer.

I understand your concern very well, and I also understand very well that I need to use a specific version. 

I'm planning to follow goatcorp for the master branch, but manage the version for Korean servers in a separate repository branch. However, it would be easier to manage if it was added to the master code, rather than adding Korean-specific code each time.

lee-jinhwan:
Additionally,

https://github.com/lee-jinhwan/Dalamud/commits/7.5.1.0
If the "Korean Support" commit from the above commit was included in the goatcorp master branch, I wouldn't have to add it every time.

In that case, I would only need to add the Korean world and injector commits needed for the Korean version from the specific version tag, and then I can release it.

Soreepeong:
I don't think Dalamud for global release should contain references for unsupported languages, as it may wrongly imply to plugin developers that the language enum values that will never be referenced from upstream Dalamud should be handled, adding extra complexity.

By the way, have you considered rebasing from/merging a specific Dalamud commit of corresponding game version instead? It should generally offer clearer list of things to do on updates, whereas making manual edits or manually cherry picking previous edits would be more likely to cause mistakes.

PR #1383, opened 2023-09-11
Add GameConfigPluginScoped

MidoriKami:


Dalamud/Game/Config/GameConfig.cs
```diff
@@ -192,3 +188,178 @@ private unsafe nint OnConfigChanged(ConfigBase* configBase, ConfigEntry* configE
...
+    internal GameConfigPluginScoped()
+    {
+        this.gameConfigService.Changed += this.ConfigChangedForward;
+    }
+    
+    /// <inheritdoc/>
+    public event EventHandler<ConfigChangeEvent>? Changed;
+    
+    /// <inheritdoc/>
+    public GameConfigSection System => this.gameConfigService.System;
```
KazWolfe, line 218: Individual `GameConfigSection`s have their own public `Changed` event. Unsure how to fix this, but it does cause issues with scoping. Will defer to Goat/Cara.

KazWolfe (changes requested):

Dalamud/Game/Config/GameConfig.cs
MidoriKami: One avenue might be to make the GameConfigSection.Changed event internal, and add those events to the interface then have pluginscoped act as a proxy for invoking the event internally.

MidoriKami: Prototype of potential solution: https://github.com/MidoriKami/Dalamud/commit/4a2d173925ef5c1481ceb691f99030a01cd782d7

KazWolfe (approved):

PR #1384, opened 2023-09-11
Add ClientStatePluginScoped

MidoriKami:
Also redefined event delegates, old delegates had unused parameters, new delegates provide what's actually used.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -10,6 +10,7 @@ using Dalamud.IoC;
 using Dalamud.IoC.Internal;
 using Dalamud.Logging.Internal;
 using Dalamud.Plugin.Services;
+using Dalamud.Utility;
 using FFXIVClientStructs.FFXIV.Client.Game;
 using Lumina.Excel.GeneratedSheets;
 
@@ -154,7 +155,7 @@ internal sealed class ClientState : IDisposable, IServiceType, IClientState
             Log.Debug("Is login");
             this.lastConditionNone = false;
             this.IsLoggedIn = true;
-            this.Login?.Invoke();
+            this.Login?.InvokeSafely();
             gameGui.ResetUiHideState();
 
             this.lifecycle.ResetLogout();
@@ -165,7 +166,7 @@ internal sealed class ClientState : IDisposable, IServiceType, IClientState
             Log.Debug("Is logout");
             this.lastConditionNone = true;
             this.IsLoggedIn = false;
-            this.Logout?.Invoke();
+            this.Logout?.InvokeSafely();
             gameGui.ResetUiHideState();
 
             this.lifecycle.SetLogout();
@@ -180,11 +181,11 @@ internal sealed class ClientState : IDisposable, IServiceType, IClientState
 
             if (this.IsPvP)
             {
-                this.EnterPvP?.Invoke();
+                this.EnterPvP?.InvokeSafely();
             }
             else
             {
-                this.LeavePvP?.Invoke();
+                this.LeavePvP?.InvokeSafely();
             }
         }
     }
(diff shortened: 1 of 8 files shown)


Dalamud/Game/ClientState/ClientState.cs
```diff
@@ -178,12 +177,102 @@ private void FrameworkOnOnUpdateEvent(Framework framework1)
 
             if (this.IsPvP)
             {
-                this.EnterPvP?.InvokeSafely();
+                this.EnterPvP?.Invoke();
```
KazWolfe, line 180: Can you please clarify why you're switching these to `Invoke` over `InvokeSafely`?

Dalamud/Game/ClientState/ClientState.cs
```diff
@@ -64,22 +63,22 @@ private ClientState(SigScanner sigScanner, DalamudStartInfo startInfo, GameLifec
     private delegate IntPtr SetupTerritoryTypeDelegate(IntPtr manager, ushort terriType);
 
     /// <inheritdoc/>
-    public event EventHandler<ushort> TerritoryChanged;
+    public event Action<ushort>? TerritoryChanged;
 
     /// <inheritdoc/>
-    public event EventHandler Login;
+    public event Action? Login;
```
KazWolfe, line 69: Can you please clarify why you're switching this to an `Action`?

KazWolfe (commented):
Looks good to me - couple requests for clarification just so we have a paper trail.

Dalamud/Game/ClientState/ClientState.cs
MidoriKami: While adding interfaces, I have been declaring events in the public interface as non-nullable, and changing the internal events to nullable. The IDE doesn't complain about this and provides nice intellisense both internally and externally.

InvokeSafely was used on the non-nullable objects to check for null and then invoke, this is not necessary anymore as the internal events are properly nullables now. Rather than using a dalamud wrapper for invoke, I felt it was best to use the language features directly.

Dalamud/Game/ClientState/ClientState.cs
MidoriKami: EventHandler provides the function signature of `public void EventHandler(object sender, EventArgs args)`

Prior to this proposed PR that delegate was invoked using `this.Login?.InvokeSafely(this, null);` which is very redundant.
The consumer of this API already has the ability to use `ClientState` as you must have a pluginscoped instance to even subscribe to the event. Additionally passing null to args makes it a meaningless parameter.

This logic was applied to all the other provided events in this API aswell. These events will not provide the necessary information without including unused parameters.

Dalamud/Game/ClientState/ClientState.cs
KazWolfe: I will note that [`InvokeSafely`](https://github.com/goatcorp/Dalamud/blob/130a57f850660992c12ca426cc0510f0b7f253f8/Dalamud/Utility/EventHandlerExtensions.cs#L58) also wraps all actions in a [generic try-catch](https://github.com/goatcorp/Dalamud/blob/130a57f850660992c12ca426cc0510f0b7f253f8/Dalamud/Utility/EventHandlerExtensions.cs#L86) to resolve any uncaught exceptions. Is this no longer needed/desired for our API surface?

MidoriKami: Good point, do we want to InvokeSafely in the ScopedService instead? This way exceptions would be caught per-plugin, instead of bubbling up to the core service?

KazWolfe: Nah, keep them in the core service. If we have an issue or an uncaught exception in Dalamud itself somehow, it'd be good to handle those in the same way.

I also don't see a particularly compelling reason to catch exceptions per-plugin, as there's not much that we would be able to do at the plugin-scope level (and the core stacktrace will already point at the faulting plugin).

KazWolfe (approved):

PR #1401, opened 2023-09-18
[AddonLifecycle] Fix incorrect delegate signature

MidoriKami:
An incorrect return value caused login screen corruption.

Weirdly this didn't effect trial accounts 🤔


KazWolfe (approved):
approved - same changes made locally during troubleshooting

PR #1409, opened 2023-09-20
Delegate nullification

MidoriKami:
Also as per goats tacit approval, marked Util.FileWriteAllTextSafe public


NotNite:
For context of those not in Discord: We figured out that scoped services don't protect plugin classes from memory leaks, so we have to null the events on Dispose. This is a somewhat common footgun for events in C# garbage collection.

I personally believe that everyone should be cleaning up in the first place, but this acts as a safeguard. This is documented on [MSFT learn](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/events/how-to-subscribe-to-and-unsubscribe-from-events#unsubscribing) and Kami was able to test with dotMemory.

goaaats (approved):

PR #1410, opened 2023-09-20
Catch exceptions in Window.Draw

Haselnussbomber:
When an Exception is thrown inside the `Draw()` method of a plugins window, it would prevent the execution of the remaining code in `Window.DrawInternal()`. I added a try/catch block around the `Draw()` call to prevent this.

The most important reason for this is to ensure that the `PostDraw()` method is always called, even if the `Draw()` method fails.
The `PostDraw()` method might be important for popping styles and colors that were pushed in the `PreDraw()` method to change the windows appearance.

### DIFF
--- a/Dalamud/Interface/Windowing/Window.cs
+++ b/Dalamud/Interface/Windowing/Window.cs
@@ -1,9 +1,11 @@
+using System;
 using System.Numerics;
 
 using Dalamud.Configuration.Internal;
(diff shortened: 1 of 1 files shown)


MidoriKami:
Instead of Serilog, use ModuleLog
```cs
    private static readonly ModuleLog Log = new("WindowSystem");
```

goaaats (approved):

PR #1413, opened 2023-09-21
Use WorkingPluginId to identify plugins in profiles

goaaats:
Solves a lot of weirdness regarding devPlugins in profiles and in general, since the introduction of plugins.
Makes profile management more maintainable in general, as we can now for real uniquely identify installs of plugins, even if there are two with the same InternalName.

We should never use InternalName to uniquely identify plugins anymore, and the code that has to deal with installing plugins should be able to handle multiple plugins with the same InternalName at some point(TODO).
There are also some blank spots UX-wise that I will try to sort out at some point, noted by TODOs.
Probably best to leave for after v9.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstaller/ProfileManagerWidget.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstaller/ProfileManagerWidget.cs
@@ -406,7 +406,7 @@ internal class ProfileManagerWidget
             foreach (var plugin in profile.Plugins.ToArray())
             {
                 didAny = true;
-                var pmPlugin = pm.InstalledPlugins.FirstOrDefault(x => x.Manifest.InternalName == plugin.InternalName);
+                var pmPlugin = pm.InstalledPlugins.FirstOrDefault(x => x.Manifest.WorkingPluginId == plugin.WorkingPluginId);
                 var btnOffset = 2;
 
                 if (pmPlugin != null)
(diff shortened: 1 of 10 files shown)


KazWolfe:
> We should never use InternalName to uniquely identify plugins anymore, and the code that has to deal with installing plugins should be able to handle multiple plugins with the same InternalName at some point(TODO).

This is a bit confusing/concerning to me. Based on my understanding, a `WorkingPluginId` is generated on the host computer (?), so is also unique to each specific install.

Does this mean that we're going to change how our IPC or other API surfaces work? I want my plugins to be able to detect if other plugins are enabled for IPC or other purposes - does this change that?

goaaats:
> This is a bit confusing/concerning to me. Based on my understanding, a WorkingPluginId is generated on the host computer (?), so is also unique to each specific install.
Yes, that is how it works. When you click "Install", a plugin receives a unique ID that will never change, even throughout updates.

> Does this mean that we're going to change how our IPC or other API surfaces work? I want my plugins to be able to detect if other plugins are enabled for IPC or other purposes - does this change that?
No, that functionality should all still stay the same. What I mean by that is that all of the code around browsing and managing plugins should be able to cope with plugins from two repos/the same repo having the same internal names, and two plugins with the same internal name being installed. From an API perspective, towards other plugins, the internal name should still be a "contract" about what this plugin is. We just need unique IDs for some Dalamud features, so that we don't get confused about which plugin we are referring to.

Dalamud/Plugin/Internal/PluginManager.cs
```diff
@@ -1290,13 +1290,30 @@ private async Task<LocalPlugin> LoadPluginAsync(FileInfo dllFile, LocalPluginMan
...
+            plugin = new LocalPlugin(dllFile, manifest);
+        }
+        
+        // Perform a migration from InternalName to GUIDs. The plugin should definitely have a GUID here.
+        // This will also happen if you are installing a plugin with the installer, and that's intended!
+        // It means that, if you have a profile which has unsatisfied plugins, installing a matching plugin will
+        // enter it into the profiles it can match.
+        if (plugin.Manifest.WorkingPluginId == Guid.Empty)
+            throw new Exception("Plugin should have a WorkingPluginId at this point");
+        this.profileManager.MigrateProfilesToGuidsForPlugin(plugin.Manifest.InternalName, plugin.Manifest.WorkingPluginId);
```
KazWolfe, line 1307: Add a newline here pls - formatting is annoying me

Dalamud/Interface/Internal/Windows/PluginInstaller/ProfileManagerWidget.cs
```diff
@@ -437,26 +437,33 @@ private void DrawEdit()
...
+                    var firstAvailableInstalled = pm.InstalledPlugins.FirstOrDefault(x => x.InternalName == plugin.InternalName);
+                    var installable =
                         pm.AvailablePlugins.FirstOrDefault(
                             x => x.InternalName == plugin.InternalName && !x.SourceRepo.IsThirdParty);
-                    if (available != null)
+                    
+                    if (firstAvailableInstalled != null)
+                    {
+                        // TODO
+                        ImGui.Text("GOAT WAS TOO LAZY TO IMPLEMENT THIS");
```
KazWolfe, line 449: *bap*

KazWolfe (approved):

PR #1419, opened 2023-09-22
Remove PluginLog

MidoriKami:
I wanted to get this in sooner rather than later as some plugin devs are already beginning the process of updating, and it would be good for them to switch their logging to scoped log sooner rather than later.


goaaats:
I don't think we wanted to do this in v9? For @KazWolfe

wolfcomp:
I would say mark it all as obsolete, so people have time to convert over to the new DI way.

MidoriKami:
The plugin scoped log is already available https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Logging/ScopedPluginLogService.cs

Caraxi:
Haven't we already talked about removing things before they've been obsolete for a while? It makes it much more difficult to update if there is nothing to tell you what it needs to be changed to

wolfcomp:
This makes it sound much like It's ok to just merge this then https://github.com/aers/FFXIVClientStructs/pull/556 which i know people asked for being still there

MidoriKami:
> Haven't we already talked about removing things before they've been obsolete for a while? It makes it much more difficult to update if there is nothing to tell you what it needs to be changed to

Good point, I'll restore it, but mark it as error.

goaaats:
Yes, the new stuff is there, the old stuff will not be removed this cycle. That was the original intention on my end, but Kaz took ownership so I wasn't sure what the status is since I basically handed it over entirely.

MidoriKami:
I'll un-involve myself on this topic then.

PR #1425, opened 2023-09-23
AddonEventManager Add Safety Check

MidoriKami:
I realized a common pitfall with attaching events is trying to attach to a component node, these nodes do not allow events to be attached directly, you have to attach the events to the contained AtkResNode itself.

Without this safety check, adding an event to a component node causes an immediate crash to desktop.

image

### DIFF
--- a/Dalamud/Game/Addon/Events/PluginEventController.cs
+++ b/Dalamud/Game/Addon/Events/PluginEventController.cs
@@ -52,7 +52,7 @@ internal unsafe class PluginEventController : IDisposable
 
         if ((int)node->Type >= 1000)
         {
-            Log.Error("Wrong node type. Attempted to attach an event to a component node. No event was attached.");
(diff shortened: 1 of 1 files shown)


goaaats:
Would it be better to throw here? Can this happen by accident?

MidoriKami:
It might be better to throw, throwing would more clearly indicate the line in the plugin code that is causing the problem.

> Can this happen by accident?

It is pretty easy to do accidentally, what can commonly happen is a dev wants to add an event to a button, buttons are component nodes, you can't attach the event to the button itself, you have to attach it to the contained atkresnode.

Using AtkUnitBase's `AtkComponentButton* GetButtonNodeById(uint nodeId);` for example, returns the address of the component, not the AtkResNode

image

MidoriKami:
Sorry there is an error in my last comment, you must attach to the contained `AtkResNode`, you can not attach to the `OwnerNode`

image

MidoriKami:
After getting a clarification from Aers the NodeType check would not help this case, the user simply has to pass a `Node` and not a `Component`

PR #1430, opened 2023-09-24
Fix window system scoping

Infiziert90:
Changes how the Namespace is applied to both WindowSystem and Window, changing it from PushID to a combined string which is used in `Begin()`

The resulting Window ID:  `WindowName##WindowSystem##Window`
Original: `PushID WindowSystem -> PushID Window -> Begin(WindowName)`

Any additional # will be removed from the start of both namespace strings to prevent "###", which would overwrite WindowName "###"


MidoriKami:
https://github.com/Infiziert90/Dalamud/blob/Namespace-Fix/Dalamud/Interface/Internal/Windows/ChangelogWindow.cs#L46

This line either needs to be removed, or the namespace added to the window name itself.

PR #1452, opened 2023-10-03
Implement dalamud-platform launch argument

marzent:
Identical to https://github.com/goatcorp/Dalamud/pull/1433, but mergeable with the current master branch.


KazWolfe:
Can you please rebase this branch?

KazWolfe (approved):
One day I will learn why we're so obsessed with launch args instead of env vars. `XL_PLATFORM` would have been _much_ simpler...

PR #1472, opened 2023-10-06
fix: testing API level being ignored

Aireil:
This needs a `TestingDalamudApiLevel` field added to the plugin master, with the exact same behavior as `TestingAssemblyVersion`.
This _should_ work even without that field in the plugin master, it will just use the normal assembly.


KazWolfe:
Please rebase this against the `apiX` branch. Otherwise, LGTM.

goaaats (approved):

goaaats (approved):

PR #1473, opened 2023-10-06
AddonLifecycle Add AddonReceiveEvent

MidoriKami:
Adds events for receiving addon events without users needing to hook them.

Also fixes a bunch of copy-paste errors in documentation.

### DIFF
--- a/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonDrawArgs.cs
+++ b/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonDrawArgs.cs
@@ -1,7 +1,7 @@
 ﻿namespace Dalamud.Game.Addon.Lifecycle.AddonArgTypes;
 
 /// <summary>
-/// Addon argument data for Finalize events.
+/// Addon argument data for Draw events.
 /// </summary>
 public class AddonDrawArgs : AddonArgs
 {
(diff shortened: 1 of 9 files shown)


KazWolfe:
Logic seems fine, but how does this work with other hooks? (That is, can we make any sort of guarantee that `Before`/`After` will always run, regardless of other hooks?)

MidoriKami:
> Logic seems fine, but how does this work with other hooks? (That is, can we make any sort of guarantee that `Before`/`After` will always run, regardless of other hooks?)

I don't understand your question.

MidoriKami:
After getting some clarification via conversation with KazWofle, the concern can be summarized as:
Since this system is hooking the ReceiveEvent from the addons in much the same way other plugin developers can, are there any conflicts or issues that can arise from this approach?

There's a couple possible conflicts:
#### What happens if someone hooks an addons ReceiveEvent function, but doesn't call `.Original`
Due to hook order being indeterminate someone else could hook the ReceiveEvent function and prevent this service's ReceiveEvent from being called.

#### What happens if someone's hook manages to modify the values before our event?
Our "Before" and "After" calls aren't guaranteed to be before and after if someone else hooks the ReceiveEvent function themselves.

#### Response
These concerns are not specific to this service, and due to the nature of hooks, it's always possible that conflicts like this will occur. I believe it is out of scope for this service to be overly concerned with these possible conflicts.

There is sufficient exception handling in place to where it is unlikely a plugin developer will be able to induce an error in the service, the result will be that other users of the services won't receive calls to their event delegate. This is no different than if the service didn't exist and multiple people hooked the same ReceiveEvent function.

I assert that if you need complex behaviors with addons then you should hook the methods you need directly and interact with the game the way you need to, while this service provides a basic easy to user interface that satisfies many common use cases.

goaaats (approved):

PR #1474, opened 2023-10-06
Show warning before deleting plugin data

foophoof:
These two buttons are fairly easy to click accidentally (especially the one right next to the "enable testing" button), so it would be good to show a confirmation modal before actually deleting the data.

Also, I hadn't realised that "reset plugin data" meant "delete plugin configuration", in the context of reloading the plugin I thought it meant resetting the in-memory data, so the button didn't do what I expected either, an additional reminder would've helped. Not sure how many would get confused by this, though.

This is my first time contributing here and also my first time writing any C# in a _long_ time, so I'm not 100% confident in this code, but I did test it locally and it seemed to work for me. I mostly just duplicated some of the code from the other modals.


goaaats (approved):

goaaats:
Looks good, and sorry for the trouble!

PR #1477, opened 2023-10-08
Fix crash on regex error

MidoriKami:
**Also removes obsolete source context lookup**


MidoriKami:
The diff for the toolbar function is misleading, logic is unchanged there, simply use imraii for the table instead of doing it manually.

KazWolfe (approved):

PR #1483, opened 2023-10-10
AddonLifecycle Extra Safety

MidoriKami:
This PR adds a sig for AtkEventListener, so that we can get the address of the base AtkEventListener.ReceiveEvent function.
This is used during the OnSetup of an addon to check to see if that addon implements their own ReceiveEvent function.
If they do not implement their own ReceiveEvent function, then we **should not** hook the AtkEventListener.ReceiveEvent base function.

### DIFF
--- a/Dalamud/Game/Addon/Lifecycle/AddonLifecycle.cs
+++ b/Dalamud/Game/Addon/Lifecycle/AddonLifecycle.cs
@@ -49,7 +49,8 @@ internal unsafe class AddonLifecycle : IDisposable, IServiceType
         this.address = new AddonLifecycleAddressResolver();
         this.address.Setup(sigScanner);
 
-        this.disallowedReceiveEventAddress = (nint)((AtkEventListener*)this.address.AtkEventListener)->vfunc[2];
+        // We want value of the function pointer at vFunc[2]
+        this.disallowedReceiveEventAddress = ((nint*)this.address.AtkEventListener)![2];
         
         this.framework.Update += this.OnFrameworkUpdate;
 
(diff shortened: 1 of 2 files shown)


Dalamud/Game/Addon/Lifecycle/AddonLifecycle.cs
```diff
@@ -169,13 +174,19 @@ private void OnAddonSetup(AtkUnitBase* addon, uint valueCount, AtkValue* values)
...
         {
             // Hook the addon's ReceiveEvent function here, but only enable the hook if we have an active listener.
+            // Disallows hooking the core internal event handler.
             var addonName = MemoryHelper.ReadStringNullTerminated((nint)addon->Name);
-            var receiveEventHook = Hook<AddonReceiveEventDelegate>.FromAddress((nint)addon->VTable->ReceiveEvent, this.OnReceiveEvent);
-            this.receiveEventHooks.TryAdd(addonName, receiveEventHook);
-
-            if (this.eventListeners.Any(listener => (listener.EventType is AddonEvent.PostReceiveEvent or AddonEvent.PreReceiveEvent) && listener.AddonName == addonName))
+            var receiveEventAddress = (nint)addon->VTable->ReceiveEvent;
+            Log.Debug($"{receiveEventAddress:X} - {this.disallowedReceiveEventAddress:X}");
```
KazWolfe, line 180: Remove this log entry, or set to verbose with more descriptive text.

Dalamud/Game/Addon/Lifecycle/AddonLifecycle.cs
```diff
@@ -197,7 +208,14 @@ private void OnAddonSetup(AtkUnitBase* addon, uint valueCount, AtkValue* values)
...
         }
 
-        addon->OnSetup(valueCount, values);
+        try
+        {
+            addon->OnSetup(valueCount, values);
+        }
+        catch (Exception e)
+        {
+            Log.Error(e, "Caught exception when calling original AddonSetup. This may be a bug in the game or another plugin hooking this method.");
```
KazWolfe, line 217: (Not a review - just idle commentary)

Does it make sense to handle exceptions calling original? If the game is breaking things, we should let it crash imo. I'm unsure if it's additionally worth bailing out problems caused by other plugin devs - they'd crash the game if this system didn't exist, or if they manage to get their hooks in first somehow.

KazWolfe (changes requested):

Dalamud/Game/Addon/Lifecycle/AddonLifecycle.cs
MidoriKami: Oops, nice catch.

KazWolfe (approved):

PR #1486, opened 2023-10-12
Add appropriate HTTP headers to repository requests

lojewalo:
HTTP requests to repositories should specify they're looking for JSON and who's asking, so this just adds `Accept: application/json` and `User-Agent: Dalamud/<git hash shown in /xldev>` headers to the requests.

### DIFF
--- a/Dalamud/Plugin/Internal/Types/PluginRepository.cs
+++ b/Dalamud/Plugin/Internal/Types/PluginRepository.cs
@@ -47,7 +47,7 @@ internal class PluginRepository
             },
             UserAgent =
             {
-                new ProductInfoHeaderValue("Dalamud", typeof(Dalamud).Assembly.GetName().Version?.ToString()),
+                new ProductInfoHeaderValue("Dalamud", Util.AssemblyVersion),
             },
         },
     };
--- a/Dalamud/Plugin/Internal/Types/PluginRepository.cs
+++ b/Dalamud/Plugin/Internal/Types/PluginRepository.cs
@@ -47,7 +47,7 @@ internal class PluginRepository
             },
             UserAgent =
             {
-                new ProductInfoHeaderValue("Dalamud", $"{Util.GetGitHash()}[{Util.GetGitCommitCount()}]"),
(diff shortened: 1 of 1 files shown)


Dalamud/Plugin/Internal/Types/PluginRepository.cs
```diff
@@ -37,10 +37,18 @@ internal class PluginRepository
...
+            {
+                new MediaTypeWithQualityHeaderValue("application/json"),
+            },
             CacheControl = new CacheControlHeaderValue
             {
                 NoCache = true,
             },
+            UserAgent =
+            {
+                new ProductInfoHeaderValue("Dalamud", $"{Util.GetGitHash()}[{Util.GetGitCommitCount()}]"),
```
avafloww, line 50: Could you change this to use the version number (i.e. `Dalamud/9.0.0.2`) instead? imo we'd prefer that; it feels more likely to be useful as a UA, and is more convention-following/standards-abiding than a hash.

avafloww (changes requested):
just one small change but otherwise this feels OK to merge!

Dalamud/Plugin/Internal/Types/PluginRepository.cs
lojewalo: I didn't know where to find that, but I'll search around!

lojewalo:
```cs
ImGui.MenuItem(dalamud.StartInfo.GameVersion?.ToString() ?? "Unknown version", false);
```
Looks like that's not accessible from a static context, though.

avafloww: > ```cs
> ImGui.MenuItem(dalamud.StartInfo.GameVersion?.ToString() ?? "Unknown version", false);
> ```
> 
> Looks like that's not accessible from a static context, though. Unless you just mean use reflection for AssemblyVersion?

er, yeah, sorry for the ambiguity - the Dalamud AssemblyVersion is what we're looking for here! iirc `Util.AssemblyVersion` exists for this

KazWolfe (approved):
Cool, looks good, thank you!

avafloww (approved):

PR #1490, opened 2023-10-16
add support for area instance to SeString.CreateMapLink()

im-scared:
`SeString.CreateMapLink()` does not currently support area instance in the name of the link. this change adds support for `instance`, so that map link text can match what is shown in game with fields like `<flag>` and `<pos>` as follows:

image

this is only a change to the text of the link, as instance doesn't affect the link behavior itself (as far as i know).

### DIFF
--- a/Dalamud/Game/Text/SeStringHandling/SeString.cs
+++ b/Dalamud/Game/Text/SeStringHandling/SeString.cs
@@ -254,10 +254,10 @@ public class SeString
     /// <param name="rawX">The raw x-coordinate for this link.</param>
     /// <param name="rawY">The raw y-coordinate for this link..</param>
     /// <returns>An SeString containing all of the payloads necessary to display a map link in the chat log.</returns>
-    public static SeString CreateMapLink(uint territoryId, uint mapId, int rawX, int rawY)
+    public static SeString CreateMapLink(uint territoryId, uint mapId, int rawX, int rawY, int? instance = null)
     {
         var mapPayload = new MapLinkPayload(territoryId, mapId, rawX, rawY);
-        var nameString = $"{mapPayload.PlaceName} {mapPayload.CoordinateString}";
+        var nameString = GetMapLinkNameString(mapPayload.PlaceName, mapPayload.CoordinateString, instance);
 
         var payloads = new List<Payload>(new Payload[]
         {
@@ -280,10 +280,10 @@ public class SeString
     /// <param name="yCoord">The human-readable y-coordinate for this link.</param>
     /// <param name="fudgeFactor">An optional offset to account for rounding and truncation errors; it is best to leave this untouched in most cases.</param>
     /// <returns>An SeString containing all of the payloads necessary to display a map link in the chat log.</returns>
-    public static SeString CreateMapLink(uint territoryId, uint mapId, float xCoord, float yCoord, float fudgeFactor = 0.05f)
+    public static SeString CreateMapLink(uint territoryId, uint mapId, float xCoord, float yCoord, float fudgeFactor = 0.05f, int? instance = null)
     {
         var mapPayload = new MapLinkPayload(territoryId, mapId, xCoord, yCoord, fudgeFactor);
-        var nameString = $"{mapPayload.PlaceName} {mapPayload.CoordinateString}";
+        var nameString = GetMapLinkNameString(mapPayload.PlaceName, mapPayload.CoordinateString, instance);
 
         var payloads = new List<Payload>(new Payload[]
         {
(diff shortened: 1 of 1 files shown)


Dalamud/Game/Text/SeStringHandling/SeString.cs
```diff
@@ -254,10 +254,10 @@ public static SeString CreateItemLink(Item item, bool isHq, string? displayNameO
     /// <param name="rawX">The raw x-coordinate for this link.</param>
     /// <param name="rawY">The raw y-coordinate for this link..</param>
     /// <returns>An SeString containing all of the payloads necessary to display a map link in the chat log.</returns>
-    public static SeString CreateMapLink(uint territoryId, uint mapId, int rawX, int rawY)
+    public static SeString CreateMapLink(uint territoryId, uint mapId, int rawX, int rawY, int? instance = null)
```
KazWolfe, line 257: This is a breaking API change. Please remove the default param and instead create an overload without the optional param:

```csharp
public static SeString CreateMapLink(uint territoryId, uint mapId, int rawX, int rawY) => 
    CreateMapLink(territoryId, mapId, rawX, rawY, null);

public static SeString CreateMapLink(uint territoryId, uint mapId, int rawX, int rawY, int? instance) {
   ...
```

Applies to all changed public methods.

KazWolfe (changes requested):

KazWolfe:
As an aside for musings: do we know how the game generates these links? Would it somehow be possible to call that more directly? (What would the tradeoffs of this be?)

Dalamud/Game/Text/SeStringHandling/SeString.cs
im-scared: ahh yes will do! just curious (because i don't know enough about c# yet) is this breaking because when compiled, old plugins will still be trying to find the function without the extra param, despite the default?

KazWolfe: Yep, that’s the easy answer. We had a problem specifically related to this crop up today with an earlier PR already.

im-scared:
> As an aside for musings: do we know how the game generates these links? Would it somehow be possible to call that more directly? (What would the tradeoffs of this be?)

i am sprout. know nothing. but would love to find out there's a more direct solution ^w^

Dalamud/Game/Text/SeStringHandling/SeString.cs
im-scared: oh actually, since we're doing overloads now instead of defaults, would it be better to place instance after `mapId` and `placeName` respectively, so it aligns with where it would appear in the link text?

im-scared: i did rearrange the function arguments, but if you'd prefer i add `instance` on the end i don't mind putting it back

im-scared:
ah shoot i should have taken a screenshot, but i built and tested this and was able to generate a map link with instance in the chatlog! ^w^

Dalamud/Game/Text/SeStringHandling/SeString.cs
```diff
@@ -306,7 +332,20 @@ public static SeString CreateMapLink(uint territoryId, uint mapId, float xCoord,
...
+    /// Creates an SeString representing an entire Payload chain that can be used to link a map position in the chat log, matching a specified zone name.
+    /// Returns null if no corresponding PlaceName was found.
+    /// </summary>
+    /// <param name="placeName">The name of the location for this link.  This should be exactly the name as seen in a displayed map link in-game for the same zone.</param>
+    /// <param name="instance">An optional area instance number to be included in this link.</param>
+    /// <param name="xCoord">The human-readable x-coordinate for this link.</param>
+    /// <param name="yCoord">The human-readable y-coordinate for this link.</param>
+    /// <param name="fudgeFactor">An optional offset to account for rounding and truncation errors; it is best to leave this untouched in most cases.</param>
+    /// <returns>An SeString containing all of the payloads necessary to display a map link in the chat log.</returns>
+    public static SeString? CreateMapLink(string placeName, int? instance, float xCoord, float yCoord, float fudgeFactor = 0.05f)
```
KazWolfe, line 348: Please rename anything using `instance` to `CreateMapInstanceLink` or `CreateInstancedMapLink`. Keeping the current order and nullability is fine I think, but this is more explicit and likely easier to understand at a glance.

For example, `CreateMapLink("Bozja", currentInstance, mod.PosX, mob.PosY);` will resolve to `placeName, xCoord, yCoord, fudgeFactor` despite very clear variables saying that's... not what it does. IDEs will yell at you for this, but relying on IDEs to catch mistakes like this isn't particularly fun imo.

We'll look at merging them together properly in API X.

im-scared: done and tested! ^w^

```csharp
            this.chatGui.Print(SeString.CreateMapLink(817, 495, 10.6f, 18.2f));
            this.chatGui.Print(SeString.CreateMapLink(817, 495, 43, 136));
            this.chatGui.Print(SeString.CreateMapLink("The Rak'tika Greatwood", 10.6f, 18.2f));
            this.chatGui.Print(SeString.CreateMapLinkWithInstance(817, 495, 6, 10.6f, 18.2f));
            this.chatGui.Print(SeString.CreateMapLinkWithInstance(817, 495, 1, 43, 136));
            this.chatGui.Print(SeString.CreateMapLinkWithInstance("The Rak'tika Greatwood", 3, 10.6f, 18.2f));
```

image

KazWolfe (approved):

PR #1494, opened 2023-10-20
Resize collision node instead of root node in DtrBar

nebel:
This PR changes the way the DtrBar is resized to allow mouse click and hover events.

Currently, we allow such events by expanding the size of the root node to cover all additional Dalamud-created nodes. This works, but we've since discovered some rare edge case issues related to how the game's HUD layout system. Essentially, if the bar is anchored to a certain side of the screen and that side is also the bar's growth-direction, it can result in the addon having an unstable location since it ends up anchoring to a potentially moving point on the root node.

With this change, we instead opt to resize the primary collision node and leave the root node intact. This means the root node's size is never changed and so it behaves identically to the unmodified version. This should solve any positioning issues. One downside to this approach is we can no longer rely on using the game's default collision size to determine the size of the native nodes and are forced to calculate it ourselves, but fortunately the logic to do so is not so complex. On the other hand we can now avoid a forced update request as we no longer have to rely on the game's right-justify logic for fixing node alignment.

### DIFF
--- a/Dalamud/Game/Gui/Dtr/DtrBar.cs
+++ b/Dalamud/Game/Gui/Dtr/DtrBar.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 
 using Dalamud.Configuration.Internal;
-using Dalamud.Game.Addon;
 using Dalamud.Game.Addon.Events;
 using Dalamud.Game.Addon.Lifecycle;
 using Dalamud.Game.Addon.Lifecycle.AddonArgTypes;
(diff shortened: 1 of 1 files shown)


MidoriKami:
👍 I endorse this PR.

nebel and I chatted about this fix a bit, some interesting tidbits where this is a rare case where events are being handled exclusively through the collision node, and not through the AtkUnitBase. This means there was actually no need to resize the addon itself. *This doesn't apply to other addons though*, it seems this is a bit of a rare case where we can just manipulate the collision node directly.

goaaats (approved):

PR #1495, opened 2023-10-20
Bump Luminas

lmcintyre:
Adds a completely new namespace for generated Lumina sheets, GeneratedSheets2. These are generated from the project https://github.com/xivdev/EXDSchema for the current game version, [2023.09.28.0000.0000](https://github.com/xivdev/EXDSchema/tree/main/2023.09.28.0000.0000).
Will write up some dev docs for the API soon.
This impl is non-breaking but non-final.


lmcintyre:
cringe

goaaats:
Needs to be bumped in all projects. We should have a separate targets file that declares shared dependencies.

goaaats (approved):

PR #1499, opened 2023-10-22
Bump Luminas

lmcintyre:
Adds a completely new namespace for generated Lumina sheets, GeneratedSheets2. These are generated from the project https://github.com/xivdev/EXDSchema for the current game version, [2023.09.28.0000.0000](https://github.com/xivdev/EXDSchema/tree/main/2023.09.28.0000.0000).
Will write up some dev docs for the API soon.
This impl is non-breaking but non-final.

This utilizes Lumina 3.15.2 which fixes the previously breaking behavior upon attempting to get a row which does not exist.

### DIFF
--- a/Dalamud.CorePlugin/Dalamud.CorePlugin.csproj
+++ b/Dalamud.CorePlugin/Dalamud.CorePlugin.csproj
@@ -27,8 +27,8 @@
     </ItemGroup>
 
     <ItemGroup>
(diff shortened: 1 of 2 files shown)


goaaats (approved):

lmcintyre:
Just tested against almost every plugin that uses Lumina in the main repository (I had 60 installed.) No issues encountered regarding Lumina behavior. Ready for merge.

PR #1504, opened 2023-10-27
Use GameObject Method

MidoriKami:
Uses the GetObjectId method instead, which correctly looks up the ObjectId for objects that would otherwise have an id of 0xE0000000 in the ObjectTable.

Easy example, without this change all gathering nodes in the island sanctuary show an objectId of 0xE0000000

With this change they show meaningful objectId's as seen below:
image

### DIFF
--- a/Dalamud/Game/ClientState/Objects/Types/GameObject.cs
+++ b/Dalamud/Game/ClientState/Objects/Types/GameObject.cs
@@ -101,7 +101,7 @@ public unsafe partial class GameObject
     /// <summary>
     /// Gets the object ID of this <see cref="GameObject" />.
     /// </summary>
-    public uint ObjectId => this.Struct->ObjectID;
+    public uint ObjectId => this.Struct->GetObjectID().ObjectID;
 
     /// <summary>
     /// Gets the data ID for linking to other respective game data.


goaaats:
We should make this change, but I don't know if it's good to change behavior like this. I would guess that some plugins rely on this being 0xE0000000.

MidoriKami:
My thought was to try it in staging for a few days and see, in my very brief testing it didn't seem like anything broke in any obvious way.

Understandable if you want to wait for an API bump instead. It's worth noting, even with an API bump communicating this subtle change may be difficult.

KazWolfe:
Yeah, I'm going to defer this until the next API bump. While I do agree with the change in general, new findings around `ObjectId` will probably require that we rethink how we expose these things.

MidoriKami:
Then it might be best to close the PR, and instead at api-next return a dalamud equivalent of [GameObjectId](https://github.com/aers/FFXIVClientStructs/blob/e2355970e236ffe09dfc7af674d1e29fb6b93786/FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs#L13)
While providing helpers to get the objectId the current way

KazWolfe:
Yeah fair enough. I'll convert this to an Issue.

If we have a need for `GameObjectId` or equivalent, I think it'll be fine to bring it in early, just that we'll deprecate it come API 10 when we redo things.

KazWolfe:
See #1512.

PR #1507, opened 2023-10-29
Fix crash from BuildLookupTable access violation

Soreepeong:
ImGui resolves ` ` with `FindGlyph`, which uses `FallbackGlyph`, which is resolved after resolving ` `.
On the first call of `BuildLookupTable`, called from `BuildFonts`, `FallbackGlyph` is set to `null`, making `FindGlyph` return `null`.
On our secondary calls of `BuildLookupTable`, `FallbackGlyph` is set to some value that is not null, making ImGui attempt to treat whatever was there as a ` `.
This may have caused random glyphs to be sized randomly, if not an access violation exception.

### DIFF
--- a/Dalamud/Interface/GameFonts/GameFontManager.cs
+++ b/Dalamud/Interface/GameFonts/GameFontManager.cs
@@ -169,7 +169,7 @@ internal class GameFontManager : IServiceType
         }
 
         if (rebuildLookupTable && fontPtr.Glyphs.Size > 0)
-            fontPtr.BuildLookupTable();
+            fontPtr.BuildLookupTableNonstandard();
     }
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


goaaats:
Is that something we can also fix on the ImGui-side?

Soreepeong:
We can change `FindGlyph` before `FallbackGlyph = ...` line in `BuildLookupTable` to `FindGlyphNoFallback`, but I think `BuildLookupTable` is supposed to be an imgui internal function that external use cases probably aren't considered.

goaaats:
Oh, I thought it was something we added, makes sense. Thanks for looking into this, much appreciated!

goaaats (approved):

PR #1510, opened 2023-10-30
Write configs asynchronously, first part of async API rewrite

goaaats:
Should fix some microstuttering complaints we have been getting. Reads aren't async yet, they definitely should be, but they should take the same code path as before the VFS in 99.9% of cases so it doesn't matter at the moment.

### DIFF
--- a/Dalamud/Configuration/Internal/DalamudConfiguration.cs
+++ b/Dalamud/Configuration/Internal/DalamudConfiguration.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 
 using Dalamud.Game.Text;
 using Dalamud.Interface.Internal.Windows.PluginInstaller;
@@ -38,6 +39,8 @@ internal sealed class DalamudConfiguration : IServiceType, IDisposable
 
     [JsonIgnore]
     private bool isSaveQueued;
+
+    private Task? writeTask;
 
     /// <summary>
     /// Delegate for the <see cref="DalamudConfiguration.DalamudConfigurationSaved"/> event that occurs when the dalamud configuration is saved.
@@ -488,6 +491,9 @@ internal sealed class DalamudConfiguration : IServiceType, IDisposable
     {
         // Make sure that we save, if a save is queued while we are shutting down
         this.Update();
+        
+        // Wait for the write to finish
+        this.writeTask?.Wait();
     }
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


Dalamud/Storage/ReliableFileStorage.cs
```diff
@@ -91,8 +92,9 @@ public bool Exists(string path, Guid containerId = default)
     /// <param name="path">Path to write to.</param>
     /// <param name="contents">The contents of the file.</param>
     /// <param name="containerId">Container to write to.</param>
-    public void WriteAllText(string path, string? contents, Guid containerId = default)
```
Infiziert90, line 94: Isn't that here a breaking change, as the class is public and could have been used since it was released with Api 9?

KazWolfe: This isn't a plugin-exposed service. The entire interface should probably be `internal`, but eh.

Dalamud/Plugin/DalamudPluginInterface.cs
```diff
@@ -338,12 +338,30 @@ public ICallGateSubscriber<TRet> GetIpcSubscriber<TRet>(string name)
...
+    /// <param name="currentConfig">The current configuration.</param>
+    /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
+    public async Task SavePluginConfigAsync(IPluginConfiguration? currentConfig)
+    {
+        if (currentConfig == null)
+            return;
+
+#pragma warning disable CS0618 // Type or member is obsolete
+        await this.configs.SaveAsync(currentConfig, this.plugin.InternalName, this.plugin.Manifest.WorkingPluginId);
+#pragma warning restore CS0618 // Type or member is obsolete
```
KazWolfe, line 364: These pragmas are unnecessary, `SaveAsync` is not obsolete.

Dalamud/Configuration/Internal/DalamudConfiguration.cs
```diff
@@ -508,7 +514,8 @@ private void Save()
     {
         ThreadSafety.AssertMainThread();
 
-        Service<ReliableFileStorage>.Get().WriteAllText(
+        this.writeTask?.Wait();
+        this.writeTask = Service<ReliableFileStorage>.Get().WriteAllTextAsync(
             this.configPath, JsonConvert.SerializeObject(this, SerializerSettings));
         this.DalamudConfigurationSaved?.Invoke(this);
```
KazWolfe, line 520: We'll be invoking `DalamudConfigurationSaved` before the save actually completes. This *seems* fine, given that the only use I can see is updating `DalamudPluginInterface#GeneralChatType`, but it does feel like a behavioral change that should at least be noted.

KazWolfe (changes requested):

goaaats:
> To prevent double writes, any queued io operations should have a cancellation token before committing and care must be taken before creating a new write task to make sure that only one queued write exists at the same time. An example of that would be an older write operation overwriting a newer one if its thread had to be stalled for any reason

Are you referring to something specific, or are you saying that the system should handle this? 

I'm waiting for these tasks in the two spots I'm using this now before kicking off another, so this shouldn't be an issue, error handling isn't great but I wanted to get this in for the patch. I'll go over it again when this is going to be a public API.

goaaats:
Superseded

PR #1519, opened 2023-11-04
Use system fonts as Dalamud font

Soreepeong:
* DirectWrite has been used to query installed fonts.
* Font names will be localized, according to Dalamud language settings. CurrentCultureIgnoreCase is used to sort the font list.
* Depending on language choice (SC, TC, and K in particular) it will try to load default Windows fonts with corresponding glyph ranges, if it has not been loaded yet.
* If SC or TC is being used, Chinese character range will be dropped when loading AXIS fonts, so that they can be fully replaced by user-provided fonts.
    * Noto Sans K can be removed from assets.
* Some helper classes and methods are added.
* Fixed some fonts having blurred texts when kerning is enabled, by rounding kerning distance into integers.
* Reset button always uses default monospace font.
* Some fonts such as Wingdings cannot be loaded; this is a stb_truetype limitation.

@Bluefissure Which Chinese fonts should be attempted if Chinese characters are set to be always displayable? Currently using Yahei/UI and JhengHei/UI for SC and TC respectively; adding non-Windows default fonts in the list of fonts to try works too, such as Source Han Sans SC/TC.

## Screenshots
image
image
image
image
image
image
image


Bluefissure:
I think Yahei is good, it's also the default font of Windows 10/11.

Soreepeong:
open again when its done

PR #1523, opened 2023-11-08
Fix fly text atk arrays + offsets

Aireil:


Bluefissure:
The new indexes are the same as I found in [my little toy](https://github.com/Bluefissure/MultiHit/blob/master/MultiHit/Plugin.cs#L288-L290) and I tested yesterday with the change of num array size 10, it's working well.
I may test this build with new index & offset later today.

goaaats (approved):

PR #1527, opened 2023-11-11
Fix GameFontStyle typo, update size recommendations

Soreepeong:
8

Also, changes for `GetRecommendedFamilyAndSize` is done to accommodate the use of very slightly smaller font, like using "12.8px" font when 13px is requested.


Soreepeong:
Merging this commit into better one.

PR #1534, opened 2023-11-21
Update .editorconfig for Resharper/Rider to conform with StyleCop rules

Soreepeong:
More probably exist but these are the ones I've found irritating so far.


MidoriKami:
Does this also fix the ordering with "required" properties?

Soreepeong:
StyleCop is like a US cop; not just any cop, but a US cop

goaaats (approved):

PR #1541, opened 2023-11-23
Add injector option to not apply any exception handlers

Soreepeong:
So that you do not end up with some extra unnecessary stack trace when you're debugging with one attached.

### DIFF
--- a/Dalamud.Boot/dllmain.cpp
+++ b/Dalamud.Boot/dllmain.cpp
@@ -134,7 +134,7 @@ DWORD WINAPI InitializeImpl(LPVOID lpParam, HANDLE hMainThreadContinue) {
 
     logging::I("Initializing VEH...");
     if (g_startInfo.NoExceptionHandlers) {
-        logging::I("=> Exception handlers are disabled from DalamudStartInfo.");
+        logging::W("=> Exception handlers are disabled from DalamudStartInfo.");
     } else if (utils::is_running_on_wine()) {
         logging::I("=> VEH was disabled, running on wine");
     } else if (g_startInfo.BootVehEnabled) {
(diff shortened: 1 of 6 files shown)


Dalamud.Boot/dllmain.cpp
```diff
@@ -133,7 +133,9 @@ DWORD WINAPI InitializeImpl(LPVOID lpParam, HANDLE hMainThreadContinue) {
     // ============================== VEH ======================================== //
 
     logging::I("Initializing VEH...");
-    if (utils::is_running_on_wine()) {
+    if (g_startInfo.NoExceptionHandlers) {
+        logging::I("=> Exception handlers are disabled from DalamudStartInfo.");
```
KazWolfe, line 137: Please mark this as a warning - it's an exceptional case that we want extremely visible in logs.

Dalamud.Common/DalamudStartInfo.cs
```diff
@@ -17,38 +17,6 @@ public DalamudStartInfo()
         // ignored
     }
 
-    /// <summary>
-    /// Initializes a new instance of the <see cref="DalamudStartInfo"/> class.
-    /// </summary>
-    /// <param name="other">Object to copy values from.</param>
-    public DalamudStartInfo(DalamudStartInfo other)
```
KazWolfe, line 24: No longer used anywhere? (non-blocking, just asking)

KazWolfe (changes requested):

Dalamud.Common/DalamudStartInfo.cs
Soreepeong: C# records have an implicit constructor doing the same; even without that, `new DalamudStartInfo(new DalamudStartInfo())` compiles.

KazWolfe (approved):

PR #1542, opened 2023-11-23
Rewrite ImGuiScene using TerraFX + Add DX12 backend

Soreepeong:
image

### Effective
* Frequency of heap allocation has been hugely reduced.
* Memory leak in ImGuiScene has been addressed.
* Number of supported image file types has been increased.

### Code
* Integrated `deps/ImGuiScene` into Dalamud in a namespace `Dalamud.ImGuiScene`, after a rewrite to avoid allocations and use TerraFX instead of SharpDX.
* Changed image loader from stb_image to WindowsCodecs.
* Duct taping into InterfaceManager to avoid modifying ImGuiScene submodule can stop.
* `Dx12Win32Scene` has been added.
    * You can use `Dx12OnDx11Win32Scene` to test it out.
* Incorporates: #1526
    * `IImGuiScene.SetTexturePipeline`, `GetTexturePipeline`, and `ReleaseTexturePipeline` has been added. `CreateTexturePipeline` is on the concrete class for the interface.

### Possible future
* Remove `PInvoke.User32`, `PInvoke.Kernel32`, and `NativeFunctions`, and replace them with `TerraFX.Interop.Windows.Windows`.

### Obsoleting
* `SharpDX.Direct3D11` is pending for deletion; it has been `[Obsolete]`d in `UiBuilder`.
    * Expose `DevicePtr` instead, so that consumer plugins can wrap the native `ID3D11Device*` into their wrapper library of choice.
* `SharpDX.Mathematics` is pending for deletion; it has been `[Obsolete]`d in `VectorExtensions`.
    * That's one less autocomplete candidate when you're trying to write `Vector3` and stuff..

## Notes
ImGui Native Backend cannot be used as-is. ImGui's default DX11 implementation does not push and pop HS/DS/CS in `ID3D11DeviceContext`, and Win32 implementation deals with IME, which probably will conflict with the game's handling of IME.


goaaats:
Curious about the note regarding the native DX11 backend, I tried this a while ago and it seemed to be working fine. Was I missing something?

Soreepeong:
image

IME completion window would not show, and when multi-viewport is enabled, outside-the-game windows would get considered as not a valid target for IME input; only type English to those.

Soreepeong:
DX11 part works fine; it's restructuring of what has been working so far after all. DX12 part, on the other hand, needs more work. On a moderate load of loading/unloading textures, it causes AVE accessing nullptr during Create(Whatever)Resource calls.

Dalamud.Boot/dllmain.cpp
```diff
@@ -5,12 +5,35 @@
...
+//
+// #include "hooks.h"
 
 HMODULE g_hModule;
 HINSTANCE g_hGameInstance = GetModuleHandleW(nullptr);
 
 DWORD WINAPI InitializeImpl(LPVOID lpParam, HANDLE hMainThreadContinue) {
     g_startInfo.from_envvars();
+    //
+    // static std::optional<hooks::import_hook<decltype(D3D11CreateDevice)>> s_hookD3D11CreateDevice;
```
goaaats, line 19: What is this for? If it's valuable to leave around for someone that might need it in the future, could you leave a comment explaining it?

Soreepeong: That actually shouldn't be left there, as that's there only for testing. If DirectX debug mode is desired, I think it'd be better to implement that on C# side too?

Dalamud/ImGuiScene/Helpers/ImGuiViewportHelpers.cs
```diff
@@ -0,0 +1,164 @@
...
+    /// </summary>
+    /// <param name="viewport">An instance of <see cref="ImGuiViewportPtr"/>.</param>
+    public delegate void ChangedViewportDelegate(ImGuiViewportPtr viewport);
+
+    /// <summary>
+    /// Disables ImGui from disabling alpha for Viewport window backgrounds.
+    /// </summary>
+    public static unsafe void EnableViewportWindowBackgroundAlpha()
+    {
+        var offset = 0x00007FFB6ADA632C - 0x00007FFB6AD60000;
```
goaaats, line 146: This is a little scary, we should ideally fix it in our imgui fork. Maybe leave a todo so that we can do that when we rebuild it some day?

Soreepeong: Ideally we would remove this line:
https://github.com/goatcorp/gc-imgui/blob/e4920dc10cbf32c0cd9b4a48a4a71c8814242460/imgui.cpp#L6126

goaaats:
@Soreepeong Did you do any tests on how this interacts with reshade while working on it? Same as before?

Soreepeong:
This makes reshade unaware of windows created from using viewports.

PR #1546, opened 2023-11-24
Fix incorrect ImGui code

Sirius902:
Adds a missing `ImGui.EndTabBar()` call to end the tab bar here. https://github.com/goatcorp/Dalamud/blob/40e90a39c8250528eeb94191a026a821c8a1493c/Dalamud/Interface/Internal/Windows/Settings/SettingsWindow.cs#L94

Note: I also just upstreamed some fixes for incorrect ImGui code that will crash without clear stacks with 42af3c80f2a259becb214754d67b5452b97d35cf.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/Settings/SettingsWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/Settings/SettingsWindow.cs
@@ -155,6 +155,8 @@ internal class SettingsWindow : Window
                     ImGui.EndTabItem();
                 }
             }
+
+            ImGui.EndTabBar();
         }
 
         ImGui.SetCursorPos(windowSize - ImGuiHelpers.ScaledVector2(70));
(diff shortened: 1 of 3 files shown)


goaaats (dismissed):

Dalamud/Interface/Internal/Windows/Settings/SettingsWindow.cs
```diff
@@ -157,6 +157,8 @@ public override void Draw()
             }
         }
 
+        ImGui.EndTabBar();
```
wolfcomp, line 160: Should be moved 2 lines up due to ending a conditional check, where this can break.

Sirius902: Oh good catch, I believe I moved it to the correct location now.

goaaats (approved):

PR #1547, opened 2023-11-25
Add `IDalamudConfigReader` and early Analytics system

KazWolfe:
Add a new service, `IDalamudConfigReader`. This service is designed to expose certain internal Dalamud configurations or states to plugins in a simple manner. Config entries are exposed to plugins as required, and generally in cases where there's no other way to get them (e.g. ImGui styles).

In addition, leverage this service to provide two new config options: 

* `PluginAnalyticsConsent`, which controls whether plugins are allowed to collect analytics on players:
    * `OptedOut`: The user has requested that no analytics be collected on them.
    * `Ask` (default): Plugins must ask the user via their own UI before collecting analytics.
    * `OptedIn`: The user has automatically consented to analytics collection from all plugins.
 * `PluginAnalyticsId`, which is an opaque string that uniquely identifies a specific Dalamud installation. Every plugin is exposed a different Analytics ID derived from the plugin's InternalName and the user's global Analytics ID (as a GUID). If a user has opted out of analytics, this value will return a consistent value across all installs of that plugin.

This PR is a preview/draft, and comments/suggestions are welcome as this API is scoped out.

### DIFF
--- a/Dalamud/Configuration/DalamudConfigReader.cs
+++ b/Dalamud/Configuration/DalamudConfigReader.cs
@@ -1,8 +1,10 @@
 ﻿using Dalamud.Configuration.Internal;
+using Dalamud.Configuration.Internal.Types;
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
 using Dalamud.Plugin.Internal.Types;
 using Dalamud.Plugin.Services;
+using Dalamud.Utility;
 
 namespace Dalamud.Configuration;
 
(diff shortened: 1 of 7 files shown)


Soreepeong:
The internal `DalamudConfiguration` should be marked to implement`IDalamudConfigReader`, so that it is easier to know when removing an option would be a breaking change for pluguns and typos can be caught quicker.

KazWolfe:
From the discussion in Discord:

I don't necessarily want to bind our internal configs to public API (in case we want to remove something from the config) or limit this service to *just* configs - this may be an ideal place to expose other (read-only) states in the future.

Similarly, I don't necessarily want to expose the full set of internal configs to plugins. While, yes, plugins can reflect anything they want, I don't necessarily think it's a good idea to provide an easy interface to access the true analytics ID or other configuration options.

KazWolfe:
We've determined that it's better for services to expose their own config fields where appropriate, so IDalmaudConfigReader is unnecessary.

The Analytics API has been stale for a long time with no seeming desire from the ecosystem to progress, implying this isn't really something we want to manage.

PR #1550, opened 2023-11-28
Send Dalamud user-agent when downloading plugins

lojewalo:
The sequel to https://github.com/goatcorp/Dalamud/pull/1486.

### DIFF
--- a/Dalamud/Networking/Http/HappyHttpClient.cs
+++ b/Dalamud/Networking/Http/HappyHttpClient.cs
@@ -1,6 +1,9 @@
 ﻿using System;
 using System.Net;
 using System.Net.Http;
+using System.Net.Http.Headers;
+
+using Dalamud.Utility;
 
 namespace Dalamud.Networking.Http;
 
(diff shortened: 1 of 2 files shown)


KazWolfe (approved):
Nice, you got to the thing I was going to mention before I even had to say it.

goaaats (changes requested):
This is now the same as https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Plugin/Internal/Types/PluginRepository.cs#L50, is that one still required? Can you remove it if it's not?

lojewalo:
No, they're separate `HttpClient` instances.

KazWolfe:
> This is now the same as https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Plugin/Internal/Types/PluginRepository.cs#L50, is that one still required? Can you remove it if it's not?

Yeah, it is because we need to set additional default settings there (specifically `Timeout`) that we don't necessarily want on our global shared HTTP client. We can rewrite that class to use the global client and specify our timeouts via cancellation tokens if we really wanted to, I suppose.

goaaats:
That probably should have been the approach then, it's better to use one client per application in either case. I don't think it should be hardcoded in two separate spots.

PR #1551, opened 2023-11-28
Remove second HttpClient from PluginRepository

KazWolfe:
A naive attempt to fix the problem brought up in #1550's reviews.

```
----- Request Start ----->
Method:          GET
Path:            /_test

Headers:
Accept:          application/json
Accept-Encoding: gzip, deflate, br
Cache-Control:   no-cache
Host:            localhost:8080
User-Agent:      Dalamud/9.0.0.13
<----- Request End -----
```


Dalamud/Plugin/Internal/Types/PluginRepository.cs
```diff
@@ -204,4 +208,17 @@ private bool IsValidManifest(RemotePluginManifest manifest)
 
         return true;
     }
+
+    private async Task<HttpResponseMessage> GetPluginMaster(string url, int timeout = HttpRequestTimeoutSeconds)
+    {
+        var httpClient = Service<HappyHttpClient>.Get().SharedHttpClient;
```
goaaats, line 214: This class has happyHttpClient as a member now, should be fine using that!

goaaats (commented):
Is there a reason you're not removing `this.httpClient` while you're at it?

goaaats:
Oh oops, I think I just messed up the merge, never mind

goaaats (approved):

PR #1553, opened 2023-11-29
Use the substitution provider for textures in the ULD Wrapper.

Ottermandias:


Dalamud/Interface/UldWrapper.cs
```diff
@@ -155,20 +155,27 @@ public void Dispose()
 
         // Try to load HD textures first. 
         var hrPath = texturePath.Replace(".tex", "_hr1.tex");
+        var substitution = Service<TextureManager>.Get();
+        hrPath = substitution.GetSubstitutedPath(hrPath);
         var hd = true;
-        var file = this.data.GetFile<TexFile>(hrPath);
-        if (file == null)
+        var tex = Path.IsPathRooted(hrPath)
```
goaaats, line 161: I don't like that this is necessary, we really should have an API that allows this in ITextureProvider. Maybe something to look at in the future.

PR #1554, opened 2023-12-02
IGameInventory: checks game inventory change

Soreepeong:
This new plugin-facing service would offer developers to subscribe to inventory modification events, such as:
* An item has been added to the inventory.
* An item has been removed from the inventory.
* An item has been changed in the inventory.
* An item has been moved to another slot, possibly across inventories.
* Two items have been merged to one slot, possibly across inventories.
* An item has been split into two slots, possibly across inventories.

Note that the last 3 events require two item slots to be modified, which would be reinterpreting the first 3 events. Should one not desire such reinterpretation, they can use `InventoryChangedRaw` event. Or, if they want to get all inventory events happened in a frame in a batch, they can use `InventoryChanged`.

Ideas by and worked with @MidoriKami. SortaKinda can make use of this feature, by subscribing to item change events so that it can sort accordingly, possibly with the exception of manual item move across slots in a same class of inventory.

Note: `ConsoleWindow` also has been changed to give up earlier if regexError is true; it tries to re-compile in every iteration inside LINQ, and failing miserably over and again.

On `...Explicit` variants of events: should one want to avoid having to specify type names and casts manually, they can use those events instead. In addition, the event dispatching code can get some type safety, through the implicit type checking from invoking explicit event handler variants.


Kurochi51:
What's the difference between the last 2 events? I imagine it's something along the lines of "2 singular items are merged into a stacking item" vs "one item merges into an existing stack item", but I figured I'd ask.

Soreepeong:
Split can be done by right clicking on a stackable item, and merge can be done by dragging a stack of item onto other stack of item where both are the same item.

Kurochi51:
Split? Both events mention merge 🤔

goaaats (approved):

PR #1557, opened 2023-12-03
Fix Dalamud trying to unload IServiceType and crashing

Sirius902:
Dalamud currently iterates over all types to determine which services to unload, attempting to unload `IServiceType` which crashes since it's not a concrete type. Instead, we shouldn't attempt to unload and by extension call methods with reflection on abstract types.

### DIFF
--- a/Dalamud/ServiceManager.cs
+++ b/Dalamud/ServiceManager.cs
@@ -336,7 +336,7 @@ internal static class ServiceManager
         
         foreach (var serviceType in Assembly.GetExecutingAssembly().GetTypes())
         {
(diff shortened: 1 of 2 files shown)


KazWolfe (approved):

KazWolfe:
Reverted in #1559, and not actually deployed to stg.

This apparently exposes some lower level breakage in loading plugins at startup (somehow), which may seriously negatively impact certain plugins. Out of an abundance of caution (and me being bad at QA), I think it's best to let the game crash on launch until Kizer can fix the underlying dependency issues.

Apologies!

Issue #1498, opened 2023-10-21
[Feature Request] Improvements to Japanese input


torakiti2017:
- Use GoogleJapaneseIME.
- The same with Microsoft IME.

When typing in Japanese in the plugin window, text is also typed in the chat field.
Is it possible to prevent text from being entered in the chat field when entering text in the plugin field?
1.Entering text in the plugin window
<img width="447" alt="2" src="https://github.com/goatcorp/Dalamud/assets/86459831/0f70b7fc-3e29-45b6-824b-04333e6f9a2c">

2.Text is also entered in the chat field
<img width="192" alt="3" src="https://github.com/goatcorp/Dalamud/assets/86459831/ed61d157-7803-40b9-93d4-c342bfc86ab9">

Also, when I focus on the text input field in the plugin, the chat input afterwards is in half-width English characters. Since the input type in the chat field is double-byte, is it possible to force it not to be half-width English characters?
1.Click on the input field in the plugin window
<img width="447" alt="11" src="https://github.com/goatcorp/Dalamud/assets/86459831/018253f2-3961-4ff0-9106-9bb1e80e7d3b">

2.Click on the input field of the chat field
<img width="107" alt="12" src="https://github.com/goatcorp/Dalamud/assets/86459831/3661f79c-40a2-48ba-94e9-0f6b853b8de7">

3.input "s" and "a
<img width="80" alt="13" src="https://github.com/goatcorp/Dalamud/assets/86459831/b4adc524-ff40-4415-baa5-e2f223447cd2">
Half-width characters are entered.

4.The expected behaviour is that "さ" is entered
<img width="94" alt="14" src="https://github.com/goatcorp/Dalamud/assets/86459831/22abdc73-638f-4097-972f-3a025825f3e9">

nebel:
I tested this and wanted to provide more details.

> When typing in Japanese in the plugin window, text is also typed in the chat field.

I confirmed that this bug happens with the caveat that the chat window (or other native text input field) must be active (i.e. with a visible cursor) and using kana input for it to happen. Non-IME input doesn't behave this way and will only input to imgui.

> Also, when I focus on the text input field in the plugin, the chat input afterwards is in half-width English characters.

This bug is maybe worse than it sounds. The actual behavior seems to be something like this:

1. Focus any imgui text input. It doesn't matter if we're using kana or alpha input.
[...]

PR #1564, opened 2023-12-07
Make CJK IMEs work better

Soreepeong:
| Japanese IME | Korean IME |
| ---- | ---- |
| image | image |

## Changes
* When using CJK IMEs, it will use SeIconFont to display the current input mode. Only applicable when current thread keyboard layout (=input method) is one of CJK, regardless of variants (trad/simplified/hk/etc.)
* When using Japanese IME, it will display correctly segmented candidate selector display.
* When using Korean IME, it will stop randomly dropping characters on all the random occasions you can imagine.
* IME windows will now show in other viewports too.
* IME will not stop working when changing across viewports.

Someone should test Chinese IMEs. This probably should work because Chromium had the least amount of special-casing for Chinese IME among CJK IME handling code, but better make sure.


goaaats:
@Bluefissure would you mind testing this PR with a Chinese IME?

nebel:
I tested these changes against the Japanese issues listed in #1498 and it seems to fix both of them (simultaneous input problem, and the "stuck in kana mode" problem).

There is still a small issue where, if there is an active input (i.e. blinking cursor) in a native game input, and we switch to an imgui input, then both completion windows will appear as text is typed since both inputs are kind of simultaneously active (and if there is a completion active already when focus is changed, those completion options will also carry over to the imgui input completion). This only happens one-way as switching to a native game input will clear imgui focus, but not vice-versa. However, selecting a completion will only add it to the actually active input, so I think this is a very minor issue and not worth holding back these changes.

Bluefissure:
It's working fine in Microsoft Pinyin (Chinese IME) as:
image

It also fixes https://github.com/ottercorp/Dalamud/issues/39

Soreepeong:
@Bluefissure Now that you're testing it, can you test if Latin character input mode displays its corresponding symbol "A" or "_A"? Since Japanese IME had different condition for that compared to Korean, need to check if Chinese IME works like Korean IME or Japanese IME.

Bluefissure:
> @Bluefissure Now that you're testing it, can you test if Latin character input mode displays its corresponding symbol "英"? Since Japanese IME had different condition for that compared to Korean, need to check if Chinese IME works like Korean IME or Japanese IME.

Yes it shows "英" in Latin mode of Microsoft Pinyin 
image

and "中" for Chinese mode
image

and shows nothing when I turn the IME off
image

Soreepeong:
Sounds good enough then!

Dalamud/Interface/Internal/DalamudIme.cs
```diff
@@ -0,0 +1,630 @@
...
+
+    /// <summary>
+    /// Gets a value indicating whether to display the cursor in input text. This also deals with blinking.
+    /// </summary>
+    internal static bool ShowCursorInInputText
+    {
+        get
+        {
+            if (!ImGuiHelpers.IsImGuiInitialized)
+                return true;
```
goaaats, line 82: Should this be false?

Soreepeong: It doesn't matter; it decides whether to display the blinking `|` on text input control, and those earlier if statements are there just to prevent null dereferencing, in case something accesses that property before ImGui gets initialized.

goaaats (approved):

PR #1565, opened 2023-12-08
[master] Update ClientStructs

bleatbot:


wolfcomp:
Do not merge until the next action run, as there was a slight-unseen issue when I checked one of the PRs for CS and didn't catch that some of the types were wrong. This was fixed with aers/FFXIVClientStructs#724

wolfcomp:
Might be nice to consider merging this after aers/FFXIVClientStructs#701 is merged as it will include a lot of new systems possibly desired by a few main repo plugins.

KazWolfe:
`AddonActionBarCross` was completely deleted. While it appears to be renamed to `AddonActionCross`, this *is* a breaking change to the CS API.

wolfcomp:
Only the filename was changed to match the class name.
The namespace is the exact same as is the class name.

KazWolfe (approved):

PR #1569, opened 2023-12-09
Allow plugins to bundle their own Lumina

NotNite:
Started from a discussion in `#dalamud-dev`: https://canary.discord.com/channels/581875019861328007/860813266468732938/1183117203093073940

- Plugins that bundle their own Lumina will have exceptions thrown trying to use Dalamud APIs that expose Lumina. I consider this a tradeoff.
- The only plugin that exposes a custom Lumina on the official repository (to my knowledge) is Distant Seas, which I have a fix prepared for.

I was able to load on the title screen and didn't see any exceptions or broken logic in the plugins I have installed.

### DIFF
--- a/Dalamud/Plugin/Internal/Types/LocalPlugin.cs
+++ b/Dalamud/Plugin/Internal/Types/LocalPlugin.cs
@@ -624,8 +624,6 @@ internal class LocalPlugin : IDisposable
         config.IsUnloadable = true;
         config.LoadInMemory = true;
         config.PreferSharedTypes = false;
-        config.SharedAssemblies.Add(typeof(Lumina.GameData).Assembly.GetName());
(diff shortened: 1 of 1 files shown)


KazWolfe:
> Plugins that bundle their own Lumina will have exceptions thrown trying to use Dalamud APIs that expose Lumina. I consider this a tradeoff.

This is an extremely severe trade-off as it renders `DataManager` more or less completely unusable, as well as creating a ton of possible interop issues when it comes to working with the Dalamud API in plugins.

Assuming this goes forward, please ensure that this is *very clearly* documented such that new developers to the ecosystem don't get extremely confused as to why things don't work and why half the API surface is now inaccessible to them. Keep in mind that there is nothing requiring devs use the Targets files (and, in fact, there are a few cases where the targets file is overly simplified - leading to cases where devs may be modifying `.csproj` files without realizing exactly what it is they're doing).

PR #1570, opened 2023-12-10
Fix AddonLifecycle ABI and deprecate arg class public ctors

Soreepeong:
It turned out that there are plugins that constructs their own instance of the event args. As it was for internal consumption, the public constructors have been deprecated, but `get; internal set;`s have been turned to public `get; init;` with an internal `get; set;` version of the property.


MidoriKami:
This addresses an issue where some devs were constructing AddonArgs to manually invoke their listeners.

I consider constructing your own AddonArgs be an improper use of this API, however since it is something that's been able to be done since 6.5, this functionality can't be removed right away.

It's intended to make these constructors internal in API 10

KazWolfe (approved):

PR #1575, opened 2023-12-15
Add Default SizeConstraints

MidoriKami:
I think this is a more user friendly approach to size constraints, as now it's not nullable the dev can just `SizeConstraints.Minimum = new Vector(..., ...)` without needing to `new` the whole struct. While at the same time allows folks that are already `new`ing the whole struct to maintain the same behavior.

Also includes removal of redundant default arguments on PlaySoundEffect


KazWolfe:
The following will still result in flickering and other undesirable behavior:

```cs
this.SizeConstraints = new WindowSizeConstraints
{
    MinimumSize = new Vector2(640, 480),
};
```

MidoriKami:
Ah right, that was the exact thing I was hoping to avoid.

PR #1576, opened 2023-12-16
AddonEventManager fix thread safety

MidoriKami:


Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -146,11 +156,34 @@ internal void RemovePluginEventController(string pluginId)
...
+        
+            this.newEventControllers.Clear();  
+        }
 
+        if (this.oldEventControllers.Any())
+        {
+            foreach (var toRemove in this.oldEventControllers)
+            {
+                this.pluginEventControllers.Remove(toRemove);
+                toRemove.Dispose();
```
Soreepeong, line 180: This should be wrapped in `try { ... } catch { /* ignore */ }`, or [`ExplicitDisposeIgnoreExceptions`](https://github.com/goatcorp/Dalamud/blob/fb864dd56d9dbc868496f579138f50399b7acec8/Dalamud/Utility/Util.cs#L655) extension method should be used. It was fine to not catch exception from `Dispose` before moving the it into `Framework.Update` block, since the exception would not have had an effect on handling of other event controllers.

While `List<T>.Add(...)` technically can throw too, the reason for that would be OOM, at which point there's no point trying to remedy that, so I think it's alright to leave the adding part out for exception handling. On the other hand, `IDisposable.Dispose` got more random reasons to throw, so yeah.

Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -146,11 +156,34 @@ internal void RemovePluginEventController(string pluginId)
...
+    private void OnFrameworkUpdate(IFramework framework1)
+    {
+        if (this.newEventControllers.Any())
+        {
+            foreach (var toAdd in this.newEventControllers)
+            {
+                this.pluginEventControllers.Add(toAdd);
+            }
+        
+            this.newEventControllers.Clear();
```
Soreepeong, line 172: Use [`TryTake`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1.trytake?view=net-8.0) instead; with the current construction of `foreach (bag) {...} /* A */ bag.Clear();`, if a plugin tries to add an event listener at a truly unfortunate timing of `/* A */`, its request would be ignored. 

```cs
while (this.newEventControllers.TryTake(out var toAdd))
{
    if (this.pluginEventControllers.All(x => x.PluginId != toAdd.PluginId))
        this.pluginEventControllers.Add(toAdd);
}
```

Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -146,11 +156,34 @@ internal void RemovePluginEventController(string pluginId)
...
 
+        if (this.oldEventControllers.Any())
+        {
+            foreach (var toRemove in this.oldEventControllers)
+            {
+                this.pluginEventControllers.Remove(toRemove);
+                toRemove.Dispose();
+            }
+        
+            this.oldEventControllers.Clear();
```
Soreepeong, line 183: Same over here, with the adding part.

Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -133,7 +143,7 @@ internal void AddPluginEventController(string pluginId)
         if (this.pluginEventControllers.All(entry => entry.PluginId != pluginId))
```
Soreepeong, line 143: What if `this.pluginEventControllers.Add` get called from `OnFrameworkUpdate`, while this `.All(...)` is being evaluated? Might as well blindly add to `this.newEventControllers`, and let `OnFrameworkUpdate` deal with duplicates.

Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -146,11 +156,34 @@ internal void RemovePluginEventController(string pluginId)
         if (this.pluginEventControllers.FirstOrDefault(entry => entry.PluginId == pluginId) is { } controller)
```
Soreepeong, line 156: Same here with the adding part.

Soreepeong (changes requested):
A few remaining possible causes for race conditions.

MidoriKami:
Per a conversation with Kizer in DM's, a much simpler method of ensuring thread safety was determined by simply forcing the offending code to run on the framework thread.

Additionally it was discussed that AddEvent and RemoveEvent modify several active game systems, and should not be allowed to be invoked from within a thread, while you ***might*** be able to get away with adding/removing events in a thread, it won't be reliable and prone to randomly exploding on you.

Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -87,6 +92,8 @@ public void Dispose()
     /// <returns>IAddonEventHandle used to remove the event.</returns>
     internal IAddonEventHandle? AddEvent(string pluginId, IntPtr atkUnitBase, IntPtr atkResNode, AddonEventType eventType, IAddonEventManager.AddonEventHandler eventHandler)
     {
+        if (!ThreadSafety.IsMainThread) throw new NotSupportedException("This should be done only from the main thread. Modifying active native code on non-main thread is not supported.");
```
goaaats, line 95: Nit: InvalidOperationException would be better here. Will change it really quick.

goaaats (approved):

Dalamud/Game/Addon/Events/AddonEventManager.cs
MidoriKami: Haha it was originally InvalidOperationException, when I presented my modification to the wording Kizer suggested NotSupportedException instead.

goaaats (approved):

PR #1580, opened 2023-12-19
Allow plugins to load Dalamud dependency assemblies

marzent:
This allows plugins to probe the directory of Dalamud when resolving assemblies as fallback, allowing them to load dependencies which are not loaded into Dalamud proper yet.
For managed assemblies additional probing paths are resolved last, just before failure, so this should be a fairly safe change (and would only take effect when resolution would fail anyways).
Also this disallows plugins loading their own `Dalamud.dll`.

### DIFF
--- a/Dalamud/Plugin/Internal/Loader/PluginLoader.cs
+++ b/Dalamud/Plugin/Internal/Loader/PluginLoader.cs
@@ -1,7 +1,7 @@
 // Copyright (c) Nate McMaster, Dalamud team.
 // Licensed under the Apache License, Version 2.0. See License.txt in the Loader root for license information.
 
-using System;
+using System.IO;
 using System.Reflection;
 using System.Runtime.Loader;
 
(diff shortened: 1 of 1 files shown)


marzent:
Reproducing the original issue is fairly difficult timing wise, however this change can be tested easily by moving for example `Penumbra.String.dll` into the Dalamud directory and Penumbra still working properly (and imagining Dalamud would also depend on that dll but only loading it later).

Dalamud/Plugin/Internal/Loader/PluginLoader.cs
```diff
@@ -151,6 +151,9 @@ private static AssemblyLoadContextBuilder CreateLoadContextBuilder(LoaderConfig
             builder.PreferDefaultLoadContextAssembly(assemblyName);
         }
 
+        builder.PreferDefaultLoadContextAssembly(Assembly.GetExecutingAssembly().GetName());
```
goaaats, line 154: Could you maybe add a comment explaining what these do?

marzent: Added some comments (and swapped the two lines, so it makes more sense)

KazWolfe (approved):
Panic approving because stg is somehow broken again, despite passing tests after a local clean build.

PR #1589, opened 2023-12-25
[AddonEventManager] Actually Ensure Thread Safety

MidoriKami:
Since plugins often load and unload async, this has been reporting exceptions due to events being registered and unregisterd from a non-main thread.

This PR will force the specific native calls that AddonEventManager uses, to be on the main game thread.

There is a **potential** but **probably impossible** race condition introduced here if someone tries to Register and event and then Unregister that event in the next call, as the Registration will return a handle to the event immediately, even if the event hasn't been registered yet.

It's probably impossible because the same way the Register gets queued in the Framework.RunOnFrameworkThread, the Unregister will most likely get queued after it regardless.

### DIFF
--- a/Dalamud/Game/Addon/Events/AddonEventListener.cs
+++ b/Dalamud/Game/Addon/Events/AddonEventListener.cs
@@ -67,7 +67,10 @@ internal unsafe class AddonEventListener : IDisposable
     {
         if (node is null) return;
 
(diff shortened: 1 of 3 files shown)


MidoriKami:
Most of these changes were discussed with @Soreepeong in DM's
This should make this fully thread safe

KazWolfe (approved):

PR #1592, opened 2023-12-31
fix: Make auto-update work again, the lazy way

KazWolfe:
This pull request aims to at least get auto update working again.

Auto updates were recently broken per some PM rewrites. This (alongside a case where a certain custom-repo plugin was triggering a chat message on the login screen) caused Dalamud to attempt to auto update before the update list was properly calculated, thereby not doing anything but marking the auto-update phase as complete. 

While it is possible to (and we probably *should*) fix the underlying problem with PluginManager such that an auto update triggers after all repos are loaded, this is somewhat more work and probably won't be clean. Per goat, this may be a good opportunity to also handle some cases such as informing plugins that there may be an update present.

### DIFF
--- a/Dalamud/Game/ChatHandlers.cs
+++ b/Dalamud/Game/ChatHandlers.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using System.Text.RegularExpressions;
+using System.Threading;
 using System.Threading.Tasks;
 
 using CheapLoc;
@@ -112,6 +113,7 @@ internal class ChatHandlers : IServiceType
 
     private bool hasSeenLoadingMsg;
     private bool startedAutoUpdatingPlugins;
+    private CancellationTokenSource deferredAutoUpdateCts = new();
 
     [ServiceManager.ServiceConstructor]
     private ChatHandlers(ChatGui chatGui)
@@ -173,7 +175,7 @@ internal class ChatHandlers : IServiceType
                 this.PrintWelcomeMessage();
             
             if (!this.startedAutoUpdatingPlugins)
-                this.AutoUpdatePlugins();
+                this.AutoUpdatePluginsWithRetry();
         }
 
         // For injections while logged in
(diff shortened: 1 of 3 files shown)


Dalamud/Game/ChatHandlers.cs
```diff
@@ -6,6 +6,7 @@
 
 using CheapLoc;
 using Dalamud.Configuration.Internal;
+using Dalamud.Game.ClientState.JobGauge.Enums;
```
goaaats, line 9: What's this for?

Dalamud/Game/ChatHandlers.cs
```diff
@@ -165,16 +171,25 @@ private void OnChatMessage(XivChatType type, uint senderId, ref SeString sender,
         if (clientState == null)
             return;
 
+        if (type == XivChatType.Notice)
+        {
+            if (!this.hasSeenLoadingMsg)
+                this.PrintWelcomeMessage();
+            
+            // FIXME (kazwolfe): This will only run on the first Notice received, meaning auto-updates won't happen
```
goaaats, line 179: I actually don't think we should change this. A lot of plugins are expecting to be updated when the character is logged in, and will show stuff like changelogs via ImGui or in chat.

Dalamud/Game/ChatHandlers.cs
```diff
@@ -271,17 +286,22 @@ private void AutoUpdatePlugins()
         var notifications = Service<NotificationManager>.GetNullable();
 
         if (chatGui == null || pluginManager == null || notifications == null)
+        {
+            Log.Warning("Aborting auto-update because a required service was not loaded.");
             return;
+        }
 
         if (!pluginManager.ReposReady || !pluginManager.InstalledPlugins.Any() || !pluginManager.AvailablePlugins.Any())
```
goaaats, line 294: We should retry immediately when this happens, once we saw the first notice.

goaaats (changes requested):

Dalamud/Game/ChatHandlers.cs
KazWolfe: derp.

Dalamud/Game/ChatHandlers.cs
KazWolfe: Hrm. Seems strange to defer updates like this but in the interest of compat I'll drop this for now.

Plugins imo shouldn't expect to update anywhere in particular, and I'd even argue that if we're auto-updating plugins, we shouldn't even load the old version. I'm sure this is one of those "sounds easy on paper but painful in practice" things, so it's not particularly worth digging for now.

If we *do* want auto-updates to fire after character login, we should still move this out of ChatHandlers - nothing is stopping a plugin from sending a `Notice` on the login screen (which is how we got into this mess in the first place).

Dalamud/Game/ChatHandlers.cs
KazWolfe: How would you like me to implement this?

Any future `Notice` will already attempt to re-trigger an auto update (as the `startedAutoUpdaingPlugins` bool hasn't been tripped yet). If you want the first Notice to still perform a retrigger I can make this method return a bool and retry from the caller, but that feels like we're just racing code at that point.

goaaats (approved):

goaaats (approved):

PR #1597, opened 2024-01-03
Use custom TaskScheduler for Framework.RunOnTick

Soreepeong:
Now you can `await` for something inside the async delegate passed to `RunOnTick`, without blocking the main thread.

## Changes
* `IFramework.RunOnFrameworkThread<?>(Func<Task<?>>, ...)` has been deprecated. These functions make no sense.
* `IFramework.DelayTicks` has been added. To be used like `Task.Delay`, but counting number of framework ticks instead.
* `IFramework.FrameworkThreadTaskFactory` has been added, so that one can use `.ContinueWith` to schedule the continuation task to be run on the framework thread.
* `Framework.RunOnTick` has been reimplemented using what effectively is `Task.WhenAll(Task.Delay(...), Framework.DelayTicks(...)).ContinueWith(...)`.
* `TaskSchedulerWidget` got extra test functionalities.

### DIFF
--- a/Dalamud/Game/Framework.cs
+++ b/Dalamud/Game/Framework.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


goaaats:
It would be nice to have a self-test for this, in the self-test window. Should be easy to add, no? Maybe we should also do that for the font stuff, which could also serve as examples.

Soreepeong:
Working on example right now!

Soreepeong:
image

Obviously, downloading using Framework.Update is slow; it'd make download speed proportional to game framerate limits.

goaaats (approved):

PR #1604, opened 2024-01-14
[master] Update ClientStructs

bleatbot:


KazWolfe:
Approving for merge after maintenance starts. Given that these are `void*` to concrete types on functions that likely nobody is using (0 results on GitHub Code Search), I see little potential for breakage. Alternatively, .55 will give us enough of a headache otherwise that it doesn't matter.

KazWolfe (dismissed):

KazWolfe (approved):

PR #1606, opened 2024-01-15
[master] Update ClientStructs

bleatbot:


KazWolfe:
Accepting JobHud API breakage. This has been around for two months now so we're pushing it a bit though per aers nobody is using these who isn't aware of this.

KazWolfe (approved):

PR #1609, opened 2024-01-17
Fix swapped U and V check

Infiziert90:
The check fails with parts that aren't same width/height because U and V are swapped

After this it loads the broken ULDs fine, and the old ones also still work in my tests.


KazWolfe (approved):
yeah sure

PR #1618, opened 2024-01-20
IFontAtlas API9 compat: support reading GameFontHandle.ImFont during UiBuilder.After/BuildFonts

Soreepeong:
Also added `[Api10ToDo]` internal marker attribute.

### DIFF
--- a/Dalamud/Interface/GameFonts/GameFontHandle.cs
+++ b/Dalamud/Interface/GameFonts/GameFontHandle.cs
@@ -2,6 +2,7 @@ using System.Numerics;
 
 using Dalamud.Interface.ManagedFontAtlas;
 using Dalamud.Interface.ManagedFontAtlas.Internals;
+using Dalamud.Utility;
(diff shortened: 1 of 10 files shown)


goaaats:
Could you also note the "migration step" plugins need to make to adapt to the new behavior, which we can put in patch notes? E.g. what code changes are necessary.

Soreepeong:
Will do, but would do that in an another PR, as this PR intends to fix a crashing bug.

goaaats:
I'll hold you to it

goaaats (approved):

Soreepeong:
Also can you make it ignore the check API compatibility one? I've changed `IFontHandle.Push` to return a `struct FontPopper` instead of `IDisposable`, to reduce heap allocation per every frame.

goaaats:
Yeah, all good. It's just "informational" for review, it won't block the submit. Thanks!

PR #1622, opened 2024-01-23
Lock font resources on Push and miscellaneous direct accesses

Soreepeong:
* `FontHandle`:
    * Moved common code from `DelegateFontHandle` and `GamePrebakedFontHandle`.
    * Added `LockUntilPostFrame` so that the obtained `ImFontPtr` and its accompanying resources are kept valid until everything is rendered.
    * Added more code comments to `Try/Lock`.
    * Moved font access thread checking logic from `InterfaceManager` to `LockUntilPostFrame`.
* `GameFontHandle`: Make the property `ImFont` a forwarder to `FontHandle.LockUntilPostFrame`.
* `InterfaceManager`
    * Added companion logic to `FontHandle.LockUntilPostFrame`
    * Accessing default/icon/mono fonts will forward to `FontHandle.LockUntilPostFrame`.


Soreepeong:
Each commit corresponds to a section in the root post. It probably will make much more sense to review each commit, instead of as a whole.

Dalamud/Interface/ManagedFontAtlas/Internals/FontHandle.cs
```diff
@@ -0,0 +1,295 @@
...
+    {
+        if (this.TryLock(out _) is not { } locked)
+            return default;
+
+        if (!ThreadSafety.IsMainThread && nextNonMainThreadFontAccessWarningCheck < Environment.TickCount64)
+        {
+            nextNonMainThreadFontAccessWarningCheck =
+                Environment.TickCount64 + NonMainThreadFontAccessWarningCheckInterval;
+            var stack = new StackTrace();
+            if (Service<PluginManager>.GetNullable()?.FindCallingPlugin(stack) is { } plugin)
```
goaaats, line 107: This is extremely slow in my experience, did you see it cause any lag when rendering? Not sure if anything does enough pushing and popping for it to matter

goaaats (approved):

Dalamud/Interface/ManagedFontAtlas/Internals/FontHandle.cs
goaaats: Ah, I missed that, thanks for clarifying!

PR #1624, opened 2024-01-29
Fix reset and reload not working

MidoriKami:
Fixes `Reset plugin data and reload` to use internal name instead of `Name`

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
@@ -3773,7 +3773,7 @@ internal class PluginInstallerWindow : Window, IDisposable
 
         public static string DeletePluginConfigWarningModal_Title => Loc.Localize("InstallerDeletePluginConfigWarning", "Warning###InstallerDeletePluginConfigWarning");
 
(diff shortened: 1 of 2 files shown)


Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
```diff
@@ -2556,7 +2556,7 @@ private void DrawInstalledPluginContextMenu(LocalPlugin plugin, PluginTestingOpt
 
             if (ImGui.MenuItem(Locs.PluginContext_DeletePluginConfigReload))
             {
-                this.ShowDeletePluginConfigWarningModal(plugin.Name).ContinueWith(t =>
+                this.ShowDeletePluginConfigWarningModal(plugin.Manifest.InternalName).ContinueWith(t =>
```
KazWolfe, line 2559: Doesn't this show the InternalName in the pop-up rather than the actual plugin name?

MidoriKami: Yes, the regular name seemed kinda weird because for chat2 it showed "vChat 2"
And with internal name it showed "vChatTwo"

I'm fine with reverting that line if you prefer.

KazWolfe: Please do. We don't expose the internal name anywhere else, and some plugins (Allagan Tools) have markedly different InternalNames that may confuse users.

If you want to fix the stray `v`, `PluginInstallerWindow.cs` L3710

KazWolfe (approved):

goaaats (approved):

PR #1625, opened 2024-01-29
Directly expose `CharacterData.ClassJob` from ClientStructs

Kurochi51:
I know this is exposed through the ExcelResolver, but polling that on framework gets expensive quickly, as I found out more than half the time I spend resolving lumina sheets to get the very same value that's readily available. I don't really expect this to be merged, at the end of the day the scale of this performance difference is miniscule in comparison to ImGui shenanigans, but it's worth a shot.

### DIFF
--- a/Dalamud/Game/ClientState/Objects/Types/Character.cs
+++ b/Dalamud/Game/ClientState/Objects/Types/Character.cs
@@ -67,6 +67,11 @@ public unsafe class Character : GameObject
     public byte ShieldPercentage => this.Struct->CharacterData.ShieldValue;
 
     /// <summary>
+    /// Gets the ClassJobID of this Chara.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


MidoriKami:
I don't understand what performance hit you are referring to.

To get the ClassJob ID without a lookup simply use ExcelResolver.Id
```cs
ClientState.LocalPlayer.ClassJob.Id
```

This returns the contained ClassJobID without accessing DataManager.

If you are instead doing:
```cs
ClientState.LocalPlayer.ClassJob.GameData.RowId
```
Then of course it will do DataManager accesses every frame.

Kurochi51:
Nah, the entire ClassJob field in `ClientState.LocalPlayer` is an `ExcelResolver`, so accessing it at all will cause it to fetch the lumina sheet.
yes

Kurochi51:
Nevermind, I misunderstood how this worked.

PR #1632, opened 2024-02-07
Change "Enable AntiDebug" label to "Disable Debugging Protections" to make it clearer

redstrate:
You need to enable this to allow debugging, but the label has the negative which doesn't make sense.

### DIFF
--- a/Dalamud/Interface/Internal/DalamudInterface.cs
+++ b/Dalamud/Interface/Internal/DalamudInterface.cs
@@ -667,7 +667,7 @@ internal class DalamudInterface : IDisposable, IServiceType
                     }
 
                     var antiDebug = Service<AntiDebug>.Get();
-                    if (ImGui.MenuItem("Enable AntiDebug", null, antiDebug.IsEnabled))
+                    if (ImGui.MenuItem("Disable Debugging Protections", null, antiDebug.IsEnabled))
                     {
                         var newEnabled = !antiDebug.IsEnabled;
                         if (newEnabled)


KazWolfe:
Rejecting out of hand - while this is a correct (and valid) thing to change, it has a lot of history behind it that it has basically become a form of Easter egg or in-joke for the community.

NotNite:
not sure I agree with this being closed, just because it's a very easy thing to change and "we have a running injoke of explaining a hard-to-understand label in our discord" is kind of weird

redstrate:
I'm also confused, because I recently had to enable this option and was really confused as to whether or not it was enabled. It wasn't until later and some more reading that I realized "AntiDebug" is the name of the mechanism that strips away the debug protections and you're "enabling" that.

If this is intentionally like this, maybe we should put a comment explaining that it's not meant to be changed?

KazWolfe:
Likely a good idea - I'll defer to Goat here since it was Word of Goat that I not change this in the past.

goaaats:
It's probably time to say goodbye to AntiDebug, but I feel like if we are changing the text, we might as well change it to what it actually does - "Disable debugging protections" or some such.

redstrate:
Changed to "Disable Debugging Protections" now, thx. I was thinking even "Disable anti-debugging protections" to still channel the name, but that's probably way too long for the menu

KazWolfe (approved):
T/N: Rename the service for apiX.

Still will require Goat sign-off.

PR #1633, opened 2024-02-07
Remove EnableWindowsTargeting from build.sh's run step

redstrate:
This removes the property that shouldn't be there, because it was considered a target.


KazWolfe:
T/N: partially reverting #1255. 

I assume you've tested that things still build properly on *nix hosts?

redstrate:
> T/N: partially reverting #1255.
> 
> I assume you've tested that things still build properly on *nix hosts?

Yes it builds on my machine except for the Injector, which are known not to build properly yet on *nix

KazWolfe (approved):
validated per [dotnet run docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-run#description) not taking `/p` flags in general.

PR #1636, opened 2024-02-11
Make the crash handler work on wine properly

marzent:
This includes a few fixes to make injector arg passing properly work (and passes in everything XL does).

VEH has been fully functional on wine 7+ from what I could gather (and the original Dalamud crash was into a VEH semi-stub on wine 6), however it currently chokes on `std::chrono::zoned_time`, which is only really used for debug logging and so it should be fine. 
In any case will try to get it fixed upstream.

### DIFF
--- a/Dalamud.Boot/DalamudStartInfo.cpp
+++ b/Dalamud.Boot/DalamudStartInfo.cpp
@@ -68,10 +68,25 @@ void from_json(const nlohmann::json& json, DalamudStartInfo::ClientLanguage& val
     }
 }
 
+void from_json(const nlohmann::json& json, DalamudStartInfo::LoadMethod& value) {
+    if (json.is_number_integer()) {
+        value = static_cast<DalamudStartInfo::LoadMethod>(json.get<int>());
+
+    }
+    else if (json.is_string()) {
+        const auto langstr = unicode::convert<std::string>(json.get<std::string>(), &unicode::lower);
+        if (langstr == "entrypoint")
+            value = DalamudStartInfo::LoadMethod::Entrypoint;
+        else if (langstr == "inject")
+            value = DalamudStartInfo::LoadMethod::DllInject;
+    }
+}
+
 void from_json(const nlohmann::json& json, DalamudStartInfo& config) {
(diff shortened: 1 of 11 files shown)


marzent:
Saving a tspack should now also work correctly on wine. 
There was another `std::chrono` issue (replaced by win32 calls) and moved towards `std::filestream` to save the tspack since the COM use with `IShellItem_BindToHandler` did not work as expected too (returns `REGDB_E_CLASSNOTREG`).

KazWolfe:
Tested OK on Windows.

Linux failed to even start the game prior to a commit being dropped (c183e8f). Since patched, game still boots again. Current state is an apparent bug in Wine >= 8.5 that causes a crash reading PDBs leading to no VEH screen and the game instead hanging indefinitely. Tested OK with [Wine 7.10.r3](https://github.com/goatcorp/wine-xiv-git/releases/tag/7.10.r3.g560db77d).

T/N: Crash screen on Steam Deck takes about 30 seconds to pop, though this is likely environmental.

marzent:
Yeah the pdb issue on Wine >= 8.5 is a bit annoying and not something I initially tested for.
In the meantime the last commit should fix it (as in properly run into the next crash handler), until we update our wine to fix that and then it should automagically start working.
As a side bonus, if for whatever reason the crash handler dies on windows as well, the game will no longer hang and continue with the next crash handler similarly.

KazWolfe:
Windows tested OK.

Wine 7.10 tests OK, VEH pops up.

Stock wine does not pop VEH (as expected); Wine's own exception handler pops instead. (Is this intended?)

marzent:
>  Wine's own exception handler pops instead.

This should greatly help debugging native crashes (plus the managed stack, if useful, will also be logged)

KazWolfe:
Also of note: on Windows, forcefully killing `DalamudCrashHandler.exe` after the game crashes will now cause FFXIV to quit after a few seconds. Prior behavior was to hang indefinitely.

I think this is fine, so +1ing.

KazWolfe (approved):

PR #1640, opened 2024-02-12
Make Dalamud Crash Handler work on Wine (Take 2)

marzent:
This eats up any `EXCEPTION_WINE_ASSERTION`s (or really any SEH exception, didn't bother to write a filter here and overcomplicate things) that can come up with certain wine versions when parsing managed pdb files.


marzent:
Will do a bit more testing to rule out other potential issues across different versions, so everything can be reviewed at once

PR #1642, opened 2024-02-12
Make RewriteRemoteEntryPointW report IErrorInfo, do VirtualProtectEx before WriteProcessMemory

Soreepeong:


KazWolfe:
Per #dalamud-dev, this fixes entrypoint injection on Wine so long as wait on plugins is disabled (?).

Not merging yet since maintenance and nobody should be using Wine entrypoint stuff so it should be safe.

KazWolfe (approved):

marzent:
Did a bit more testing and the entrypoint side of things does work as expected, again as long as wait on plugins is disabled.

With the main thread being blocked however, all MonoMod operations block on wine (and something else as well, but not too sure exactly what, will have to look into that more later).

In any case I think this is still good to have for the more technically correct virtual memory management and user-visible message boxes if something goes wrong with the new entrypoint rewrite (and the other issue probably lying somewhere in managed code, so not too relevant for this).

goaaats (approved):

PR #1643, opened 2024-02-13
[master] Update ClientStructs

bleatbot:


KazWolfe (approved):
idk why im even bothering, we're going down for maint in two hours

PR #1649, opened 2024-02-14
Temp fix insufficient ImDrawList texture handling

Soreepeong:
Fixes `ImDrawList` not correctly dealing with the current texture for that draw list not in tune with the global state. Currently, `ImDrawList::AddPolyLine` and `ImDrawList::AddRectFilled` are affected.

* The implementation for `AddRectFilled` is entirely replaced with the hook.
* The implementation for `AddPolyLine` is wrapped with `Push/PopTextureID`.

### Changes required for imgui to remove this temporary fix
* `imgui_draw.cpp`:1433 `ImDrawList::AddRectFilled`
  The if block needs a `PushTextureID(_Data->TexIdCommon)/PopTextureID()` block, if `_Data->TexIdCommon != _CmdHeader.TextureId`.
* `imgui_draw.cpp`:729 `ImDrawList::AddPolyLine`
  The if block always needs to call `PushTextureID` if the abovementioned condition is not met. Change `push_texture_id` to only have one condition.


Dalamud/Interface/ManagedFontAtlas/Internals/ImGuiDockNodeUpdateForceFont.cs
```diff
@@ -0,0 +1,78 @@
...
+/// It might be better to just special-case that particular function,
+/// as no other code touches ImDrawList that is irrelevant to the global shared state,
+/// with the exception of Dock... functions which are called from ImGui::NewFrame,
+/// which are guaranteed to use the global default font.
+/// </summary>
+[ServiceManager.EarlyLoadedService]
+internal class ImGuiRenderWindowDecorationsForceFont : IServiceType, IDisposable
+{
+    private const int CImGuiRenderWindowDecorationsOffset = 0x461B0;
+    private const int CImGuiWindowDockIsActiveOffset = 0x401;
```
KazWolfe, line 23: Will these offsets need changing if/when we update our cimgui instance?

Soreepeong: Yeah, along with the constants at `DalamudIme.cs`

KazWolfe:
**Steps to Reproduce Initial Bug:**

1. In `/xlsettings`, enable window docking and reset to default font (AXIS)
2. Add any other window (e.g. ImGui Demo) to Chat2's window
3. Observe that Chat2's window is transparent when viewed.
4. Set font back to NotoSans, observe Chat2 window is no longer transparent.

--

Bug no longer reproducible with patch applied, things seem good?

KazWolfe (approved):

Soreepeong:
Plugins calling `ImGui.GetForegroundDrawList(viewport)` apparently ends up using the global state that is not in tune with the draw list. Changing it to draft for now to address that too.

goaaats (approved):

PR #1652, opened 2024-02-14
Make Dalamud handle top-level SEH exceptions

marzent:


Dalamud.Boot/xivfixes.cpp
```diff
@@ -556,6 +556,37 @@ void xivfixes::clr_failfast_hijack(bool bApply)
     }
 }
 
+void xivfixes::keep_unhandled_exception_filter(bool bApply)
```
goaaats, line 559: Could you add a comment explaining what this does? I assume it's to prevent the game from setting its exception filter?

Dalamud.Boot/xivfixes.h
```diff
@@ -7,6 +7,7 @@ namespace xivfixes {
     void redirect_openprocess(bool bApply);
     void backup_userdata_save(bool bApply);
     void clr_failfast_hijack(bool bApply);
```
goaaats, line 9: If you're in here, you can probably delete this one, it was just an experiment and never worked correctly

Dalamud.Boot/xivfixes.cpp
marzent: Yeah, will add a comment

Dalamud.Boot/xivfixes.h
marzent: Should be gone now

KazWolfe (approved):
Seems to test alright on my end.

Strangely, after initial compilation it took about ~90 seconds to open the crash dialog despite the minidump being generated instantly. After launching with `--crash-handler-console`, the dialog began popping near instantly even after the option was removed. Going to blame Windows caching something here I guess?

goaaats (approved):

PR #1656, opened 2024-02-15
Implement Managed Asserts for ImGui 1.88

Soreepeong:


goaaats:
I was thinking that it might be more useful to override IM_ASSERT with a trampoline that passes the line and message into managed code, and make a checkbox to enable that, rather than actually adding our own validations.

I prototyped that and it seemed to work quite well, not sure what's the better way to go. With that approach you would also get inspections for a lot of other "invalid usage" of ImGui APIs.

Soreepeong:
If `IM_ASSERT` has not been nop'd on our current cimgui.dll, then definitely; if we need to compile it again, then we might as well add in other duct tape changes to there.

goaaats:
Obviously if we do stuff like that, it might be more worthwhile to fix the build situation for the native stuff first... But overriding those macros is technically the "supported" way to achieve that, so I'm more inclined towards that instead of messing with the context from C#. Seems like it's getting more pressing by the minute so maybe I should invest the time to work it out.

Soreepeong:
Regardless though, ImGui asserts will check for wrong usages at a point that may be difficult to trace back to its source; having scoped asserts for stacks in addition would be nice.

PR #1661, opened 2024-02-16
Crash Handler Improvements

KazWolfe:
- Remove the metric submission from the crash handler. It served virtually no purpose and didn't give us much/any useful information anyways.
- Add a new button to save a troubleshooting pack, below restart for now.
- Require the user choose a restart mode before enabling the restart button.
- Reword a few things all around.

image

---

This will require some additional QA, as I can't actually get a tspack to save from the crash screen on a fake inject. Not sure why, but it happens without these changes too...

### DIFF
--- a/DalamudCrashHandler/DalamudCrashHandler.cpp
+++ b/DalamudCrashHandler/DalamudCrashHandler.cpp
@@ -610,6 +610,7 @@ enum {
     IdRadioRestartWithoutDalamud,
 
     IdButtonRestart = 201,
+    IdButtonSaveTsPack = 202,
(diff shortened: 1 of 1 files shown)


goaaats (dismissed):

tommadness:
Is the `upload a troubleshoot pack` action the same as the `Save A Troubleshooting Pack` button?

KazWolfe:
Yes, both the hyperlink and the button to the same thing. The latter is just there for visibility purposes for users who don't read the help text.

goaaats (approved):

PR #1664, opened 2024-02-17
Fix B4G4R4A4->B8G8R8A8 channel extraction

Soreepeong:
On systems without support for B4G4R4A4 pixel format (when DirectX feature level 11_1 is missing), the conversion will take place; a copy-and-paste error made the read pointer advance twice as fast as it should have been.

Also reordered one function to make the order consistent.


Soreepeong:
| Reported | Repro |
| --- | --- |
| image | image |

Repro is done by forcing `SupportsDxgiFormat(Format.B4G4R4A4_UNorm)` to return false.

goaaats:
Do you think the check for modded textures is still necessary after fixing this?

goaaats (approved):

Soreepeong:
Not really; I'd expect the game to show corrupted fonts too if Dalamud reads the corrupted font files, unless there's a bug in Lumina sqpack reader which is unlikely. Let's rollback that

PR #1666, opened 2024-02-18
Add SetFontScaleMode(ImFontPtr, FontScaleMode)

Soreepeong:
`IgnoreGlobalScale` was advertised as "excludes the given font from global scaling", but the intent I had in mind was "excludes the given font from being scaled in any manner **during the font build process**". As the latter functionality is needed, obsoleted `IgnoreGlobalScale` and added `SetFontScaleMode`.

Effects of different values can be checked using "Game Prebaked Fonts".
image


Soreepeong:
API compat errors are additions to interfaces that are not meant to be implemented by plugins.

goaaats (approved):

PR #1670, opened 2024-02-19
[master] Update ClientStructs

bleatbot:


KazWolfe (approved):
breaking changes approved; nobody should  be using these apis at the moment

PR #1671, opened 2024-02-19
Make IsResumeGameAfterPluginLoad work with DllInject

marzent:
This contains only the part of https://github.com/goatcorp/Dalamud/pull/1668 that makes `DllInject` block correctly on startup instead of silently failing to do so.

### DIFF
--- a/Dalamud.Boot/DalamudStartInfo.cpp
+++ b/Dalamud.Boot/DalamudStartInfo.cpp
@@ -74,10 +74,10 @@ void from_json(const nlohmann::json& json, DalamudStartInfo::LoadMethod& value) 
 
     }
     else if (json.is_string()) {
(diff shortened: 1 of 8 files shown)


furfy:
Tried it together with my friend on Arch Linux, compiled with the latest commit, with Penumbra mods, and for some reason they can't catch up on few initial game startups. Pre-caching or indexing maybe?
But now everything works great!

KazWolfe:
@marzent Is this still applicable?

marzent:
In general yes, but it will need a rebase...

When this PR was originally made DllInject was also the only useable option on wine, so it would have been more "nice" back then.

Before putting work into this it would be good to know if ignoring `IsResumeGameAfterPluginLoad` with DllInject is actually seen as a big enough issue and if this PR is wanted.

PR #1673, opened 2024-02-20
IGameConfig: fix load-time race condition

### LINKED ISSUE #1672: GameConfig.cs Returns Null Reference During Plugin Load
A revival of the following as I do not appear to be able to re-open the issue.
https://github.com/goatcorp/Dalamud/issues/1529

I was unable to get a full log from the person who reported this issue to me, but this should hopefully cover the key aspects of the error.
Was unable to get this to reproduce on my own machine.
image

Soreepeong:
This PR fixes #1672.

As some public properties of `IGameConfig` are being set on the first `Framework` tick, there was a short window that those properties were null, which goes against the interface declaration.

This commit fixes that, by making those properties block for the full initialization of the class.

A possible side effect is that a plugin that is set to block the game from loading until it loads will now hang the game if it tries to access the game configuration from its constructor, instead of throwing a `NullReferenceException`. As it would mean that the plugin was buggy at the first place and it would have sometimes failed to load anyway, it might as well be a non-breaking change.

Additionally, minor cleanup for `DalamudAssetManager` is done.

### DIFF
--- a/Dalamud/Plugin/Services/IGameConfig.cs
+++ b/Dalamud/Plugin/Services/IGameConfig.cs
@@ -1,14 +1,20 @@
-﻿using System;
-using System.Diagnostics;
+﻿using System.Diagnostics;
+using System.Threading.Tasks;
 
 using Dalamud.Game.Config;
-using FFXIVClientStructs.FFXIV.Common.Configuration;
+using Dalamud.Plugin.Internal.Types;
 
 namespace Dalamud.Plugin.Services;
 
 /// <summary>
 /// This class represents the game's configuration.
 /// </summary>
+/// <remarks>
+/// Avoid accessing configuration from your plugin constructor, especially if your plugin sets
+/// <see cref="PluginManifest.LoadRequiredState"/> to <c>2</c> and <see cref="PluginManifest.LoadSync"/> to <c>true</c>.
+/// If property access from the plugin constructor is desired, do the value retrieval asynchronously via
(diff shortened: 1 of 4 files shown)


Dalamud/Plugin/Services/IGameConfig.cs
```diff
@@ -31,6 +37,16 @@ public interface IGameConfig
...
 
+    /// <summary>
+    /// Gets a task representing the initialization state of this instance of <see cref="IGameConfig"/>.
+    /// </summary>
+    /// <remarks>
+    /// Accessing <see cref="GameConfigSection"/>-typed properties such as <see cref="System"/>, directly or indirectly
+    /// via <see cref="TryGet(Game.Config.SystemConfigOption,out bool)"/>,
+    /// <see cref="Set(Game.Config.SystemConfigOption,bool)"/>, or alike will block, if this task is incomplete.
+    /// </remarks>
+    public Task InitializationTask { get; }
```
KazWolfe, line 48: What's the intended use for this Task? 

That is, under what conditions would a plugin want to read or care about this, and can we not just block service completion until GameConfig is ready? (What's the downside to doing this?)

Soreepeong: I was thinking about the case where a plugin might want to queue setting retrieval upon plugin constructor using `.ContinueWith` to that task, but `IFramework.RunOnTick` can be used instead. It's been removed.

goaaats (approved):

PR #1678, opened 2024-02-22
Limit console log lines held in memory

rreminy:
Currently log lines held in memory by the Dalamud Console window are currently unbounded.

This PR aims to set an initial limit to 10,000 lines using my newly implemented `RollingList<T>` implementation. This should prevent unbounded memory growth occurring as a result of leaving the game open for multiple days. No changes are made to the files themselves.

~~This PR adds a new NuGet dependency to aid in its implementation: `AG.Collections.RollingList` - https://www.nuget.org/packages/AG.Collections.RollingList~~ This dependency has been removed as requested

### DIFF
--- a/Dalamud/Configuration/Internal/DalamudConfiguration.cs
+++ b/Dalamud/Configuration/Internal/DalamudConfiguration.cs
@@ -216,6 +216,11 @@ internal sealed class DalamudConfiguration : IServiceType, IDisposable
     public bool LogOpenAtStartup { get; set; }
 
     /// <summary>
+    /// Gets or sets the number of lines to keep for the Dalamud Console window.
+    /// </summary>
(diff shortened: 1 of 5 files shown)


rreminy:
I added `LogLinesLimit` to `DalamudConfiguration` and made sure the `RollingList<T>`'s size is read from there. I also added a handler to update the `RollingList<T>` size whenever the Dalamud Configuration is saved.

I made sure changing the value is handled properly by using the Immediate Window in Visual Studio.

I'm now clueless as to where to expose this value for changing by developers. I was originally thinking somewhere in `/xlsettings` but I cannot find a good spot. Another idea would be the toolbar but also want to be careful not to clutter it. Ideas welcome.

rreminy:
oops, i meant to rename my branch, I'll remake this PR

PR #1681, opened 2024-02-24
Implement INotificationManager

Soreepeong:
### Features
* A notification window is optionally interactable, and now exposes a `DrawActions` event that can be used to draw interactable elements.
* A notification can have an icon based on `IDalamudTextureWrap`, `SeIconChar`, or `FontAwesomeIcon`. If not set and `Type` is `None`, then the plugin icon will be displayed.
* A notification always has a dismiss button, which can be used regardless of whether the notification is interactable.
* Interactable notifications will dismiss on click by default.
* Hovering on a notification will extend its duration by default.
* Unloading a plugin will remove registered events from a notification being displayed.

| Before CorePlugin unload | After unload | No type nor icon is set |
| --- | --- | --- |
| image | image | image |


Soreepeong:
image

Added expiry progressbar on bottom. If the notification is sticky for any reason, it will always be filled.

Soreepeong:
Added progressbar and moved close button to top along with showing the notification creation time. See https://discord.com/channels/581875019861328007/860813266468732938/1211323024251555870 for an example video.

Soreepeong:
image
image

Added examples for using icon sources.

Soreepeong:
Should be done at a later date but not now: make Plugin Installer show non-user-dismissable notifications and display progress and possibly a cancel button on there.

Dalamud/Interface/ImGuiNotification/Internal/ActiveNotification.cs
```diff
@@ -0,0 +1,355 @@
...
+        // After replacing, if the old texture is not the old texture, then dispose the old texture.
+        if (Interlocked.Exchange(ref this.iconTextureWrap, textureWrapTask) is { } wrapTaskToDispose &&
+            wrapTaskToDispose != textureWrapTask)
+        {
+            wrapTaskToDispose.ToContentDisposedTask(true);
+        }
+    }
+
+    /// <summary>Removes non-Dalamud invocation targets from events.</summary>
+    internal void RemoveNonDalamudInvocations()
```
goaaats, line 255: Maybe it would be good to leave a note here as to why this is required(=> we don't want to remove notifications when a plugin goes away, so we just make the actions do nothing?)

Dalamud/Interface/ImGuiNotification/NotificationUtilities.cs
```diff
@@ -0,0 +1,154 @@
...
+
+        if (!Service<PluginImageCache>.Get().TryGetIcon(
+                plugin,
+                plugin.Manifest,
+                plugin.IsThirdParty,
+                out var texture) || texture is null)
+        {
+            texture = plugin switch
+            {
+                { IsDev: true } => dam.GetDalamudTextureWrap(DalamudAsset.DevPluginIcon),
```
goaaats, line 136: Would it make sense to just draw the "question mark" icon instead? These are just overlays, might look a little odd.

goaaats (changes requested):

goaaats (approved):

PR #1682, opened 2024-02-25
Add IContextMenu service

WorkingRobot:
Adds a new `IContextMenu` service for plugins. It's not a drop-in replacement for the Dalamud.ContextMenu nuget package (should it be?). It adds several new features on top of it, as well (like nested submenus). Menu items can be added either through the `OnMenuOpened` event or the `AddMenuItem` method.

Tests/code snippets/images TBD.


goaaats:
I would really appreciate it if you could rework the test for context menus at https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Interface/Internal/Windows/SelfTest/AgingSteps/ContextMenuAgingStep.cs to test all of the stuff this can do, and that it returns sane data.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
```diff
@@ -0,0 +1,472 @@
...
+
+    /// <summary>
+    /// They're in duty.
+    /// </summary>
+    InDuty = 1ul << 43,
+
+    /// <summary>
+    /// They're a trail adventurer.
+    /// </summary>
+    TrailAdventurer = 1ul << 44,
```
Haselnussbomber, line 429: I fixed this typo 2 days ago in client structs. (https://github.com/aers/FFXIVClientStructs/commit/8a08fa4ce28752a2040f8bfab44d8e5055e6a1ce) 😅

```suggestion
    /// <summary>
    /// They're a trial adventurer.
    /// </summary>
    TriallAdventurer = 1ul << 44,
```

Haselnussbomber (changes requested):

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
```diff
@@ -0,0 +1,472 @@
...
+    [FieldOffset(0)]
+    internal readonly InfoProxyCommonList.CharacterData InternalData;
+
+    private const int StructSizeInBytes = 0x68;
+
+    /// <summary>
+    /// The view of the backing data, in <see cref="ulong"/>.
+    /// </summary>
+    [FieldOffset(0)]
+    private fixed ulong dataUInt64[StructSizeInBytes / 0x8];
```
Soreepeong, line 28: For `GameInventoryItem` struct, this was used to make `Equals` and `GetHashCode` fast; is that purpose also applicable here?

Dalamud/Game/Gui/ContextMenu/ContextMenuGui.cs
```diff
@@ -0,0 +1,501 @@
...
+
+using ValueType = FFXIVClientStructs.FFXIV.Component.GUI.ValueType;
+
+namespace Dalamud.Game.Gui.ContextMenu;
+
+/// <summary>
+/// This class handles interacting with the game's (right-click) context menu.
+/// </summary>
+[InterfaceVersion("1.0")]
+[ServiceManager.BlockingEarlyLoadedService]
```
Soreepeong, line 28: Use `EarlyLoadingService`. This service does not need to block the game from loading to initialize "better".

Dalamud/Game/Gui/ContextMenu/ContextMenuGui.cs
```diff
@@ -0,0 +1,501 @@
...
+        if (submenu->IsVisible)
+            submenu->FireCallbackInt(-1);
+
+        this.contextAddonOpenByAgentHook.Dispose();
+        this.contextMenuOnVf72Hook.Dispose();
+    }
+
+    /// <inheritdoc/>
+    public void AddMenuItem(ContextMenuType menuType, MenuItem item)
+    {
```
Soreepeong, line 97: Should menu modification functions be only called from the UI thread? Use `ThreadSafety.AssertMainThread()` in that case; otherwise, use thread-safe containers or locks.

Dalamud/Game/Gui/ContextMenu/ContextMenuGui.cs
```diff
@@ -0,0 +1,501 @@
...
+        if (!this.MenuItems.TryGetValue(menuType, out var items))
+            this.MenuItems[menuType] = items = new();
+        items.Add(item);
+    }
+
+    /// <inheritdoc/>
+    public bool RemoveMenuItem(ContextMenuType menuType, MenuItem item)
+    {
```
Soreepeong, line 105: See above on thread safety.

Dalamud/Game/Gui/ContextMenu/ContextMenuGui.cs
```diff
@@ -0,0 +1,501 @@
...
+
+    private AtkValue* ExpandContextMenuArray(Span<AtkValue> oldValues, int newSize)
+    {
+        // if the array has enough room, don't reallocate
+        if (oldValues.Length >= newSize)
+            return (AtkValue*)Unsafe.AsPointer(ref oldValues[0]);
+
+        var size = (sizeof(AtkValue) * newSize) + 8;
+        var newArray = (nint)IMemorySpace.GetUISpace()->Malloc((ulong)size, 0);
+        NativeMemory.Fill((void*)newArray, (nuint)size, 0);
```
Soreepeong, line 119: Probably a good idea to throw a `OutOfMemoryException` first if `newArray` is null, just in case future modification to this class results in `newSize` value that is not sane.

Dalamud/Game/Gui/ContextMenu/ContextMenuGui.cs
```diff
@@ -0,0 +1,501 @@
...
+        this.SetupGenericMenu(8, 0, 5, 6, items, ref valueCount, ref values);
+    }
+
+    private nint ContextAddonOpenByAgentDetour(RaptureAtkModule* module, byte* addonName, AtkUnitBase* addon, int valueCount, AtkValue* values, AgentInterface* agent, nint a7, ushort parentAddonId)
+    {
+        var oldValues = values;
+
+        var n = MemoryHelper.ReadStringNullTerminated((nint)addonName);
+
+        if (n.Equals("ContextMenu", StringComparison.Ordinal))
```
Soreepeong, line 285: Consider using `MemoryHelper.EqualsZeroTerminatedString` instead.

Dalamud/Game/Gui/ContextMenu/ContextMenuType.cs
```diff
@@ -0,0 +1,18 @@
...
+/// <summary>
+/// The type of context menu.
+/// Each one has a different associated <see cref="MenuTarget"/>.
+/// </summary>
+public enum ContextMenuType
+{
+    /// <summary>
+    /// The default context menu.
+    /// </summary>
+    Default,
```
kalilistic, line 12: historically was called GameObject by most implementations. I just copied this from Anna's previous version myself - not sure what the "right" answer is.

kalilistic: @Aireil probably has thoughts

Aireil: I think default does make more sense, as it can be the crafting log or action bars for example.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
KazWolfe: Remove this entire enum please. It’s sourced from Lumina, and I see no reason to duplicate effort.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
```diff
@@ -0,0 +1,472 @@
...
+using Dalamud.Memory;
+using FFXIVClientStructs.FFXIV.Client.UI.Info;
+
+namespace Dalamud.Game.Gui.ContextMenu;
+
+/// <summary>
+/// Dalamud wrapper around a ClientStructs CharacterData.
+/// </summary>
+[StructLayout(LayoutKind.Explicit, Size = StructSizeInBytes)]
+public unsafe struct CharacterData
```
KazWolfe, line 14: Is there (a) a reason this isn’t a Class, and (b) a reason why this has to be inside ContextMenu? Most other wrappers are classes to my knowledge.

Dalamud/Plugin/Services/IContextMenu.cs
```diff
@@ -0,0 +1,37 @@
+using Dalamud.Game.Gui.ContextMenu;
+using Dalamud.Game.Text;
+using Dalamud.Game.Text.SeStringHandling;
+
+namespace Dalamud.Plugin.Services;
+
+/// <summary>
+/// This class provides methods for interacting with the game's context menu.
+/// </summary>
+public interface IContextMenu
```
KazWolfe, line 10: Inconsistent naming - we use IContextMenu here but ContextMenuGui elsewhere.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
WorkingRobot: GameInventoryItem was a class, so I was just doing the same

Dalamud/Game/Gui/ContextMenu/ContextMenuAddressResolver.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    public nint RaptureAtkModuleOpenAddon { get; private set; }
+
+    /// <summary>
+    /// Scan for and setup any configured address pointers.
+    /// </summary>
+    /// <param name="sig">The signature scanner to facilitate setup.</param>
+    protected override void Setup64Bit(ISigScanner sig)
+    {
+        this.ContextAddonOpenByAgent = sig.ScanText("E8 ?? ?? ?? ?? 0F B7 C0 48 83 C4 60");
```
KazWolfe, line 34: Are these signatures not present in ClientStructs? If not, is there a compelling reason to have them here?

KazWolfe:
Something that may be worth discussion elsewhere: do we want to allow non-prefixed menu items at all? I see value in making it *extremely* clear that a specific menu item has come from a plugin, but likewise we have users who want to keep plugins but hide any and all easy to see traces that their game is modded.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
KazWolfe: I’m unsure GameInventoryItem was correct with their decision though they also use it for certain equality magic. If there’s no compelling reason to make this a struct, however, I think it would be best to convert to a class. 

I’d also suggest moving it somewhere more appropriate as this is for InfoProxy in general and I can see this being useful elsewhere later. This would also likely benefit a rename to CharacterSearchEntry or something similar.

goaaats:
It should not be possible to add menu items without a prefix. The prefix color and icon should be customizable.

Dalamud/Game/Gui/ContextMenu/ContextMenuType.cs
MidoriKami: Calling it "GameObject" lead to much confusion on my part when I originally tried using the Context Menu nuget.

I definitely think "Default" is more user friendly.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
MidoriKami: ***At the time*** GameInventoryItem needed to be a struct.

It was created with the intent of being able to wrap an address (like how Dalamud GameObject wraps an address), and to be a way to hold a copy of the data if needed.

**At the time** this was required because we were scanning the inventory memory every frame, and it was definitely a time-sensitive task. Initial version of the scanner were very slow.

Immediately prior to releasing the GameInventory service it was discovered that there's a inventory changed flag in one of the game systems, we now use this flag to determine if the inventory should be scanned.

**Currently** there's not really much need for most of the optimizations in GameInventoryItem, it could be cut down significantly and reduced back down to a standard class object wrapper. *Ideally* I'd love to just return the struct from ClientStructs itself instead of using a wrapper, but that's not practical at this time.

MidoriKami:
> It should not be possible to add menu items without a prefix. The prefix color and icon should be customizable.

Customizable by each plugin? Or customizable via dalamud settings?

Dalamud/Game/Gui/ContextMenu/ContextMenuGui.cs
WorkingRobot: Throwing an exception here will crash the game since this is always called from within a hook detour, but I suppose it's better to fail fast.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
```diff
@@ -0,0 +1,193 @@
...
+using FFXIVClientStructs.FFXIV.Client.UI.Info;
+
+using Lumina.Excel.GeneratedSheets;
+
+namespace Dalamud.Game.Gui.ContextMenu;
+
+/// <summary>
+/// Dalamud wrapper around a ClientStructs CharacterData.
+/// </summary>
+public unsafe class CharacterData
```
KazWolfe, line 16: Copying from prior thread:

> I’d also suggest moving it somewhere more appropriate as this is for InfoProxy in general and I can see this being useful elsewhere later. This would also likely benefit a rename to `CharacterSearchEntry` or something similar.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
```diff
@@ -0,0 +1,193 @@
...
+{
+    /// <summary>
+    /// Initializes a new instance of the <see cref="CharacterData"/> class.
+    /// </summary>
+    /// <param name="data">Character data to wrap.</param>
+    internal CharacterData(InfoProxyCommonList.CharacterData data)
```
KazWolfe, line 22: It looks like you're copying all relevant data into this class. That's fine, but please document that you're doing this - our other wrappers normally read from the backing pointer as-is and expose an `Address` as an escape hatch for users that would like extra information from the raw struct.

KazWolfe (changes requested):

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
WorkingRobot: Where do you think is a better place to put this? I'd move it, but there's nothing in Dalamud that reads info proxies yet outside of `Dalamud.Game.Network.Internal.NetworkHandlers`. Is `Dalamud.Game.Network.Structures` a good location?

Dalamud/Game/Gui/ContextMenu/ContextMenuType.cs
WorkingRobot: I'll keep this as default since the default context is used for everything but inventory items.

Dalamud/Game/Gui/ContextMenu/CharacterData.cs
KazWolfe: Yeah, that would work decently. Maybe `Dalamud.Game.Network.Structures.InfoProxy`? We can eventually move Market stuff into its own subdirectory as well.

Dalamud/Game/Gui/ContextMenu/ContextMenuAddressResolver.cs
WorkingRobot: They're not in CS, but I don't see a reason for them to be outside of ContextMenu. No one's going to call these directly unless they're doing some automation. If you find a different reason to hook this function outside of ContextMenu, I can add it to CS.

KazWolfe: The intent of putting them in CS is to reduce the maintenance burden during patch cycles - instead of having to run sig checks and update processes across multiple projects and systems, it's all in one place. 

Ideally we'd be working on minimizing our sig influence to structures and systems that don't exist or can't easily exist in CS in its current state.

goaaats:
> > It should not be possible to add menu items without a prefix. The prefix color and icon should be customizable.
> 
> Customizable by each plugin? Or customizable via dalamud settings?
> 
> Might be interesting to allow users to customize them like we allow them to modify the DTR bar entries.

I don't think that's necessary, no need to overcomplicate it for now. It should only be set by devs.

Caraxi:
> > It should not be possible to add menu items without a prefix. The prefix color and icon should be customizable.
> 
> Should I allow any custom `SeIconChar` as a prefix or should I limit it to only boxed letters?

If it ends up being only boxed letters please include the ST symbol from server time icon

Dalamud/Game/Gui/ContextMenu/ContextMenu.cs
```diff
@@ -0,0 +1,648 @@
...
+        Log.Verbose($"Got {args.MenuType} context menu with addon 0x{args.AddonPtr:X8} ({args.AddonName}) and agent 0x{args.AgentPtr:X8}");
+        if (args.Target is MenuTargetDefault targetDefault)
+        {
+            {
+                var b = new StringBuilder();
+                b.AppendLine($"Target: {targetDefault.TargetName}");
+                b.AppendLine($"Home World: {targetDefault.TargetHomeWorld.GameData?.Name.ToDalamudString() ?? "Unknown"} ({targetDefault.TargetHomeWorld.Id})");
+                b.AppendLine($"Content Id: 0x{targetDefault.TargetContentId:X8}");
+                b.AppendLine($"Object Id: 0x{targetDefault.TargetObjectId:X8}");
+                Log.Information(b.ToString());
```
KazWolfe, line 389: Are these intended to be `Log.Information`? If so, what is preventing logspam here?

WorkingRobot: They are not, no

KazWolfe: If so, please drop to `Verbose` or even consider removing this class entirely. How much value do these log entries have over a debugger? (n.b. that even verbose logs incur _some_ cost - this includes string interpolation.)

WorkingRobot: I've only left this in for debugging purposes. I'll remove this when I'm done testing.

KazWolfe: Sounds good. I’ll mark this as a formal requested change then.

Dalamud/Game/Gui/ContextMenu/ContextMenu.cs
```diff
@@ -0,0 +1,648 @@
...
+            }
+        }
+
+        var ret = this.contextAddonOpenByAgentHook.Original(module, addonName, addon, valueCount, values, agent, a7, parentAddonId);
+        if (values != oldValues)
+            this.FreeExpandedContextMenuArray(values, valueCount);
+        return ret;
+    }
+
+    private void LogMenuOpened(MenuOpenedArgs args)
```
KazWolfe, line 378: Debug method; please remove prior to shipping.

KazWolfe (changes requested):

goaaats:
> > It should not be possible to add menu items without a prefix. The prefix color and icon should be customizable.
> 
> Should I allow any custom `SeIconChar` as a prefix or should I limit it to only boxed letters?

I think any is probably fine for now. Maybe I'll change my mind before release, but I'll take care of that myself then. By default it should be the red D, as in the system menu, anyhow.

KazWolfe:
https://github.com/aers/FFXIVClientStructs/pull/816 merged via #1685, please rollback CS changes and update accordingly.

Dalamud/Game/Inventory/GameInventoryItem.cs
```diff
@@ -101,11 +104,17 @@ static GameInventoryItem()
     public ReadOnlySpan<ushort> Materia => new(Unsafe.AsPointer(ref Unsafe.AsRef(in this.InternalItem.Materia[0])), 5);
 
     /// <summary>
-    /// Gets the array of materia grades.
+    /// Gets the array of materia grades as bytes.
     /// </summary>
-    public ReadOnlySpan<byte> MateriaGrade =>
+    public ReadOnlySpan<byte> MateriaGradeBytes =>
         new(Unsafe.AsPointer(ref Unsafe.AsRef(in this.InternalItem.MateriaGrade[0])), 5);
```
KazWolfe, line 110: Mark this as `private` and place a `// TODO` for API 10 to fix the type. I'd like to avoid adding a new API that's just going to be removed in the very near future.

KazWolfe (changes requested):

Dalamud/Game/Inventory/GameInventoryItem.cs
Soreepeong: Prefer to use `[Api10ToDo(Api10ToDoAttribute.DeleteCompatBehavior)]` so that it's easier to track down what to change at a later date.

Dalamud/Game/Gui/ContextMenu/MenuTargetDefault.cs
```diff
@@ -0,0 +1,67 @@
...
+using FFXIVClientStructs.FFXIV.Client.UI.Agent;
+
+using Lumina.Excel.GeneratedSheets;
+
+namespace Dalamud.Game.Gui.ContextMenu;
+
+/// <summary>
+/// Target information on a default context menu.
+/// </summary>
+public sealed unsafe class MenuTargetDefault : MenuTarget
```
KazWolfe, line 15: Please add `AgentAddress` as a `nint` as an escape hatch so plugins can get extra data easily (e.g. mount information).

Dalamud/Game/Gui/ContextMenu/MenuTargetDefault.cs
```diff
@@ -0,0 +1,67 @@
...
+
+    /// <summary>
+    /// Gets the content id of the target.
+    /// </summary>
+    public ulong TargetContentId => this.Context->TargetContentId;
+
+    /// <summary>
+    /// Gets the home world id of the target.
+    /// </summary>
+    public ExcelResolver<World> TargetHomeWorld => new((uint)this.Context->TargetHomeWorldId);
```
KazWolfe, line 49: Can `World` be null for certain object types? If so, can you make this nullable and handle the case of `TargetHomeWorld` being invalid (zero?) by returning a `null`?

Dalamud/Utility/EventHandlerExtensions.cs
```diff
@@ -99,6 +100,23 @@ public static void InvokeSafely(this IFramework.OnUpdateDelegate? updateDelegate
         }
     }
 
+    /// <summary>
+    /// Replacement for Invoke() on OnMenuOpenedDelegate to catch exceptions that stop event propagation in case
+    /// of a thrown Exception inside of an invocation.
+    /// </summary>
+    /// <param name="openedDelegate">The OnMenuOpenedDelegate in question.</param>
+    /// <param name="argument">Templated argument for Action.</param>
+    public static void InvokeSafely(this IContextMenu.OnMenuOpenedDelegate? openedDelegate, MenuOpenedArgs argument)
```
KazWolfe, line 109: (non-blocking) this is going to become a headache at some point... need to think of a better strategy for this.

KazWolfe (changes requested):
Still need to QA this, but looks good so far. Some minor changes/nits for various things but should be easy.

For QAing, should I be able to add items to native submenus? (If so, QA verify that they get icons - only fully custom submenus where the parent has an icon should not need icons).

Dalamud/Game/Gui/ContextMenu/MenuTargetDefault.cs
WorkingRobot: I've seen it be 0 or -2,147... The resolver handles it and returns a null when the plugin calls `.GameData`.

Dalamud/Game/Gui/ContextMenu/MenuTargetDefault.cs
WorkingRobot: The AgentAddress is exposed via the `AgentPtr` property in `MenuArgs`.

Dalamud/Game/Gui/ContextMenu/MenuTargetDefault.cs
KazWolfe: Hrm. I'm unsure which would provide a better API experience then, seeing as most people tend to assume that real data resolves properly. At the very least, please add a docstring noting that this can be invalid (and therefore _actually_ return null).

Mostly trying to prevent surprised devs more than anything.

wolfcomp: If it can return null, the nullability definition should be used which is just `?` at the end

WorkingRobot: ExcelResolver's `GameData` has always had the nullable suffix.

goaaats (approved):
Leaving it to Kaz to merge, looks fine to me, although I didn't look at the entire thing. Trusting the judgement of the masses on API surface.

KazWolfe (approved):
Approving as first-pass. Going to try to allow API surface changes (within reason) through API 10.

I wish we could mark whole APIs as experimental...

PR #1685, opened 2024-02-25
[master] Update ClientStructs

bleatbot:


wolfcomp:
"fails" due to change in `MemberFunction` visibility going from `public` to `private` ref
img

KazWolfe:
API changes accepted though I will also echo questions as to whether CS should be doing this sort of check. Merge held for a decision on https://github.com/aers/FFXIVClientStructs/pull/816.

KazWolfe (approved):

PR #1689, opened 2024-02-28
fix duty pop message

pohky:
fix for unexpected payload strings in the duty pop chat message introduced by a recent change in lumina


pohky:
fixed in #1697

PR #1690, opened 2024-03-04
boot: implement --unhandled-exception=stalldebug

Soreepeong:
Using this option will stall the crashed thread until the debugger attaches, so that the newly attached debugger can be handed over the exception, enabling breaking at correct stack location.

`--no-exception-handlers` injector option controlled whether Dalamud Crash Handler would intercept exceptions before. Those are now either `default` or `none` option for `--unhandled-exception`.


KazWolfe:
Are you okay with rebasing this to API 10?

Soreepeong:
Rebased to apiX.

KazWolfe (approved):

PR #1696, opened 2024-03-05
Add IInternal/PublicDisposableService

Soreepeong:
Plugins are exposed interfaces that are not inherited from `IDisposable`, but services implementing plugin interfaces often implement `IDisposable`. Some plugins may try to call `IDisposable.Dispose` on everything provided, and it also is possible to use `using` clause too eagerly while working on Dalamud itself, such as writing `using var smth = await Service<SomeService>.GetAsync();`. Such behaviors often lead to a difficult-to-debug errors, and making those services either not an `IDisposable` or making `IDisposable.Dispose` do nothing if the object has been loaded would prevent such errors. As `ServiceManager` must be the only class dealing with construction and disposal of services, `IInternalDisposableService` has been added to limit who can dispose the object. `IPublicDisposableService` also has been added to classes that can be constructed and accessed directly by plugins; for those, `Dispose` will be ignored if the instance is a service instance, and only `DisposeService` will respond.

In addition, `DalamudPluginInterface` and `UiBuilder` also have been changed so that their `IDisposable.Dispose` no longer respond, and instead, internal functions have been added to only allow disposal from Dalamud.


haroldiedema (approved):
Very cool addition! 👍  (Una)

Dalamud/Dalamud.cs
```diff
@@ -187,27 +186,6 @@ public void WaitForUnload()
         this.unloadSignal.WaitOne();
     }
 
-    /// <summary>
-    /// Dispose subsystems related to plugin handling.
-    /// </summary>
-    public void DisposePlugins()
-    {
-        // this must be done before unloading interface manager, in order to do rebuild
```
goaaats, line 195: Are these dependencies still upheld? Who is responsible for it now?

Soreepeong: This function hasn't been called for a while, so it's probably safe to remove this at this point.
* IME: WndProc hook manager now exists separately, and it'll deal with clwaning up WndProc hooks.
* IM: On its own Dispose, it'll clean its rendering hooks from the Framework thread if it's active, so there's no chance PresentDetour getting called after that. If the game is exiting, Present calls are just over, so it doesn't matter anyway.

Soreepeong: Though, it probably is for the better to move that comment to IM.DisposeService. Will do that later.

Dalamud/Interface/Internal/InterfaceManager.cs
```diff
@@ -235,25 +234,42 @@ public IntPtr GameWindowHandle
     /// <summary>
     /// Dispose of managed and unmanaged resources.
     /// </summary>
-    public void Dispose()
+    void IInternalDisposableService.DisposeService()
     {
+        // Unload hooks from the framework thread if possible.
+        // The functions being unhooked are mostly called from the main thread, so unhooking here would avoid any
+        // chance of unhooking a function that currently is being called.
         if (Service<Framework>.GetNullable() is { } framework)
```
goaaats, line 242: Could this ever not be the case? If so, does that mean something is off with unload order? It seems to me that we would always want to avoid calling ClearHooks() off the framework thread, but please correct me if I'm missing something.

Dalamud/Storage/ReliableFileStorage.cs
```diff
@@ -22,12 +22,14 @@ namespace Dalamud.Storage;
 /// This is not an early-loaded service, as it is needed before they are initialized.
 /// </remarks>
 [ServiceManager.ProvidedService]
-public class ReliableFileStorage : IServiceType, IDisposable
+public class ReliableFileStorage : IPublicDisposableService
```
goaaats, line 25: I actually think that this is public by mistake, sorry... It should probably be internal. Maybe it's good to keep IPublicDisposableService around, since we might need it in the future - though realistically, we should probably try to keep as much complexity as possible out of Service<T> until we need it.

goaaats (changes requested):

Dalamud/Interface/Internal/InterfaceManager.cs
Soreepeong: `RunOnFrameworkThread` will do nothing if it's called after `Framework.Destroy`, so it doesn't matter. Otherwise, when initiated from /xldev "Unload Dalamud", framework should still be running, so that function actually does things. Anyway, added more comments on that.

Dalamud/Storage/ReliableFileStorage.cs
Soreepeong: Marked it obsolete, and added to TODO that we should make it internal on api 10.

goaaats (approved):

goaaats (approved):

PR #1697, opened 2024-03-05
Fix duty pop chat message italics

WorkingRobot:
Fixes the following:
image

Everything else should work the same.


KazWolfe:
dupe #1689

We're looking at whether we should yell at the Lumina people.

Soreepeong:
This probably should go in regardless of Lumina though, as the game consistently applies italics on some duty names, and it would be nice if Dalamud also did that.

KazWolfe:
Noted, point in your favor 😛.

We still need to decide if we want to yell at the Lumina people but I'll merge this as-is since I also prefer people use italics where appropriate.

KazWolfe (approved):

PR #1698, opened 2024-03-06
Accommodate nested AddonLifecycle event calls

Soreepeong:
The game is free to call event handlers of another addon from one addon, but the previous code was written under the assumption that only one function may be called at a time. This changes the recycled addon args into pooled args.

### DIFF
--- a/Dalamud/Game/Addon/AddonLifecyclePooledArgs.cs
+++ b/Dalamud/Game/Addon/AddonLifecyclePooledArgs.cs
@@ -1,0 +1,107 @@
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+using Dalamud.Game.Addon.Lifecycle.AddonArgTypes;
+
+namespace Dalamud.Game.Addon;
+
+/// <summary>Argument pool for Addon Lifecycle services.</summary>
+[ServiceManager.EarlyLoadedService]
+internal sealed class AddonLifecyclePooledArgs : IServiceType
+{
+    private readonly AddonSetupArgs?[] addonSetupArgPool = new AddonSetupArgs?[64];
+    private readonly AddonFinalizeArgs?[] addonFinalizeArgPool = new AddonFinalizeArgs?[64];
+    private readonly AddonDrawArgs?[] addonDrawArgPool = new AddonDrawArgs?[64];
+    private readonly AddonUpdateArgs?[] addonUpdateArgPool = new AddonUpdateArgs?[64];
+    private readonly AddonRefreshArgs?[] addonRefreshArgPool = new AddonRefreshArgs?[64];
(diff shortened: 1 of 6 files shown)


Aireil:
Hi Team,

I would like to further emphasize the importance of merging this pull request, "Accommodate nested AddonLifecycle event calls #1698". 

The issue at hand is a critical one that has been affecting the Always Yes tweak in SimpleTweaks for the past three months. The root cause of this problem lies in an assumption made in the previous code that only one function may be called at a time. However, the game is free to call event handlers of another addon from one addon, as Kizer rightly pointed out. 

This pull request addresses this issue by changing the recycled addon args into pooled args. This change is crucial as it rectifies the incorrect data that Dalamud was providing to the plugin, which was causing the plugin to break.

Moreover, there's another significant aspect that needs to be highlighted. The addon name was previously cached, which should not have been the case because the addon address is reused. This is a critical detail because caching the addon name can lead to incorrect data being associated with the addon, especially when the addon address is reused. By accommodating nested AddonLifecycle event calls, we can ensure that the correct data is associated with the correct addon, thereby preventing such issues from occurring in the future.

Given the importance of this fix and its potential to significantly improve the user experience, I strongly believe that it would be beneficial to expedite the review and merge of this pull request. This will not only resolve the current issues but also prevent similar problems from arising in the future.

Thank you for considering this request. I look forward to seeing the positive impact that the merge of this pull request will have on the user experience.

Best,
KalGPT

Dalamud/Game/Addon/AddonLifecyclePooledArgs.cs
```diff
@@ -0,0 +1,107 @@
...
+
+using Dalamud.Game.Addon.Lifecycle.AddonArgTypes;
+
+namespace Dalamud.Game.Addon;
+
+/// <summary>Argument pool for Addon Lifecycle services.</summary>
+[ServiceManager.EarlyLoadedService]
+internal sealed class AddonLifecyclePooledArgs : IServiceType
+{
+    private readonly AddonSetupArgs?[] addonSetupArgPool = new AddonSetupArgs?[64];
```
KazWolfe, line 12: What's the significance of a size of 64? Under what conditions would we have this many entries? (Is this a per-addon limit or a per-frame limit?)

MidoriKami:
I would like it to be on the record somewhere that I think it would be best overall to revert AddonLifecycle to prior to #1555 , as the maintenance overhead now outweighs the very *marginal* performance benefit.

Additionally the changes posed in #1555 were with the idea that devs could edit the args and have the edited args forwarded to native code. This seems to be a feature that is not being utilized, and if someone did want to utilize it doing it via AddonLifecycle poses other limitations. At this time I believe that if a dev needs that behavior they should hook the addon themselves. AddonLifecycle still provides the necessary pointers to do most manipulations even if the edited arg data is not forwarded.

PR #1701, opened 2024-03-08
Give DownloadLinkUpdate a reason to exist again

rootdarkarchon:
During the migration to .NET5 the `DownloadLinkUpdate` used in the Plugin Manifest apparently got lost in the migration and sat around unused.
This PR remedies this and makes the URL being used again for update purposes.

Sidenote 1: the returned `HttpResponseMessage` and `Stream` are both implementing `IDisposable`, yet never get disposed. It could make sense to copy the returning `Stream` from `HttpResponseMessage` into a local `MemoryStream`, dispose the `HttpResponseMessage` completely after the download completes and dispose the generated `MemoryStream` at the end of the plugin installation. I did not want to add this to this PR, as changing this functionality was out of scope.
Sidenote 2: whitespace changes came from auto formatting.

### DIFF
--- a/Dalamud/Plugin/Internal/PluginManager.cs
+++ b/Dalamud/Plugin/Internal/PluginManager.cs
@@ -470,10 +470,18 @@ internal partial class PluginManager : IInternalDisposableService
                 try
                 {
                     var dllFile = new FileInfo(Path.Combine(versionDir.FullName, $"{pluginDir.Name}.dll"));
-                    var manifestFile = LocalPluginManifest.GetManifestFile(dllFile);
-
-                    if (!manifestFile.Exists)
+                    if (!dllFile.Exists)
+                    {
+                        Log.Error("No DLL found for plugin at {Path}", versionDir.FullName);
                         continue;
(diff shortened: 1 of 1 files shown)


KazWolfe:
Out of curiosity, what's the intent here? I can't see a good reason for plugins to receive different files between initial install and update, so all I can really think of is tracking or mirroring.

goaaats:
This requires backend changes, so I'll take care of merging when ready. We'll want to post an announcement for CRP authors.

KazWolfe:
I'd suggest looking at using a header here as well - `X-Dalamud-Update-Type`. It's non-breaking, keeps a single source for plugins (although I suppose that a server can still return different data depending on the header so _meh_), and feels cleaner in general.

I don't think we _have_ to do it this way, and I see no particular problem with different URLs either, just want to surface the opportunity to decide if there's a better/simpler way of handling this.

rootdarkarchon:
I will reopen this later, syncing my master to this master yeeted the PR

PR #1708, opened 2024-03-13
Add Object Table Cache

rootdarkarchon:
Many plugins are iterating over the ObjectTable and for each iteration new objects are being generated.
This pollutes the GC with unnecessary object creation and increases execution time for plugins utilizing ObjectTable.

This PR creates a local cache for the ObjectTable that is automatically generated every tick. From my brief testing with my local plugin load it improved the times from 1.5ms average to 1.0ms average, further reducing GC spikes during plugin framework execution.
Downside of this implementation is that it will generate the object cache regardless of whether a plugin utilizes the access or not.


Dalamud/Game/ClientState/Objects/ObjectTable.cs
```diff
@@ -104,6 +109,20 @@ public unsafe IntPtr GetObjectAddress(int index)
...
         };
     }
+
+    /// <inheritdoc />
+    public void Dispose()
+    {
+        this.framework.Update -= this.FrameworkOnUpdate;
+    }
+
+    private void FrameworkOnUpdate(IFramework framework)
```
KazWolfe, line 119: T/N: this *should* be happening before any plugin (since this delegate would get registered very early), but this may cause object table updates to be delayed by a frame depending on what runs before this. I'm unsure if this would matter to Dalamud internal services, but we should probably check.

KazWolfe: Only place I can see where this would matter would be in `ClientState`. Aside from a circular dependency there on `LocalPlayer`, `ClientState` will register first meaning its framework event gets called prior to the cache update.

This looks like it'd delay login notifications and some other pieces by one frame, though I'm uncertain this would be a notable issue. There _may_ be other uses of `LocalPlayer` that I'm just not immediately seeing.

Would a `.GetDirect(int)` make sense to bypass the cache? I guess we already (indirectly) have it, but it may make sense to make it into a proper internal method for use in cases like LocalPlayer. This will wipe out some of the performance savings introduced by this PR however, as that's a not-insignificant use case.

rootdarkarchon: After adding an internal BeforeUpdate this issue should be remedied.

KazWolfe (approved):

Limiana:
With this change, accessing underlying object struct from other hooks may no longer be safe as address can be stale (OnDraw specifically). That same issue was present during .net framework 4.8 times, developers probably will do good with a warning about it

PR #1709, opened 2024-03-13
Fix CreateImGuiRangesFrom to omit null char

Soreepeong:
UnicodeRanges.BasicLatin is [0, 127], but ImGui stops reading the glyph range list on encountering a zero. Fixed that by ensuring that 0 never appears in the glyph range list.


Dalamud/Interface/ManagedFontAtlas/IFontAtlas.cs
```diff
@@ -87,6 +87,11 @@ public interface IFontAtlas : IDisposable
     /// </summary>
     /// <param name="style">Font to use.</param>
     /// <returns>Handle to a font that may or may not be ready yet.</returns>
+    /// <remarks>
+    /// <para>Do not call this function during <see cref="BuildStepChange"/>, <see cref="UiBuilder.BuildFonts"/>,
+    /// <see cref="UiBuilder.AfterBuildFonts"/>, and alike. If it works, it is accidental.</para>
```
goaaats, line 92: If this works "sometimes", should it throw InvalidOperationException instead if it's unreliable in this case?

Soreepeong: Changed to throw exceptions, if called during `IFontAtlas.BuildStepChange` event, the callback for `IFontAtlas.NewDelegateFontHandle(callback)` method, `UiBuilder.BuildFonts` event, or `UiBuilder.AfterBuildFonts` event. For compatibility, `UiBuilder.GetGameFontHandle` is being treated as an exception.

goaaats (approved):

goaaats (approved):

PR #1712, opened 2024-03-14
Fix edge case in GameVersion instance creation and refactor

karashiiro:
**Bugfixes**
- Fixes JSON serialization: `JsonConstructorAttribute` requires that the constructor parameters match fields/properties of the target class. Previously, this meant that the JSON constructor would always throw an `ArgumentNullException`, as `Input` was not a class property. To fix this, I moved the attribute to the constructor that declares every version component.

**Refactors etc.**
- Test coverage has been added for the entire class
- The constructors for each component now call each other to reduce code duplication
- Replaced a bunch of null checks into exceptions with `ArgumentNullException.ThrowIfNull`
- Replaced some switch-like chunks with switch expressions (see `GameVersion.Parse`, `GameVersion.CompareTo`)
- Replaced hash code generation with `HashCode.Combine`
- Made some other bits less verbose (and tested them)

coverage bb
image

### DIFF
--- a/Dalamud.Test/Game/GameVersionTests.cs
+++ b/Dalamud.Test/Game/GameVersionTests.cs
@@ -1,10 +1,71 @@
+using System;
+
 using Dalamud.Common.Game;
+
+using Newtonsoft.Json;
+
 using Xunit;
 
 namespace Dalamud.Test.Game
 {
     public class GameVersionTests
     {
+        [Fact]
+        public void VersionComparisons()
+        {
+            var v1 = GameVersion.Parse("2021.01.01.0000.0000");
+            var v2 = GameVersion.Parse("2021.01.01.0000.0000");
+            Assert.True(v1 == v2);
+            Assert.False(v1 != v2);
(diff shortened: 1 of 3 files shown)


Dalamud.Test/Game/GameVersionTests.cs
```diff
@@ -54,5 +187,78 @@ public void VersionConstructorInvalid(string ver)
...
+                             {
+                                "Year": -1,
+                                "Month": -1,
+                                "Day": -1,
+                                "Major": -1,
+                                "Minor": -1,
+                             }
+                             """;
+            Assert.Throws<ArgumentOutOfRangeException>(() => JsonConvert.DeserializeObject<GameVersion>(serialized));
+        }
```
karashiiro, line 226: Oh yeah, this behavior is something I'm on the fence about - calling the 5-arg ctor means it's not possible to deserialize `GameVersion.Any`, but it's better than just always throwing an `ArgumentNullException` I guess

KazWolfe (approved):

PR #1714, opened 2024-03-14
Reuse PartyMember in PartyList and Status in StatusList

rootdarkarchon:
Instead of recreating instances of PartyMember and Status on access, this will reuse existing created objects and only adjust addresses on access.
Goes into a similar vein as the latest ObjectTable adjustments.


rootdarkarchon:
abandoning due to other internal changes for now

PR #1716, opened 2024-03-14
Add support for using Server GC

rreminy:
This PR Attempts to add support for using the Server GC instead of the default Workstation GC.

Server GC is another variant of Garbage Collector designed heavier workloads, optimized for throughput.

It tends to provide a smoother experience in my case however I have read that in some cases it may cause memory usage to grow drastically. While I have not seen it happen on my limited testing, YMMV.

Related launcher PR: https://github.com/goatcorp/FFXIVQuickLauncher/pull/1447

### DIFF
--- a/Dalamud.Boot/DalamudStartInfo.cpp
+++ b/Dalamud.Boot/DalamudStartInfo.cpp
@@ -108,6 +108,7 @@ void from_json(const nlohmann::json& json, DalamudStartInfo& config) {
     config.BootVehEnabled = json.value("BootVehEnabled", config.BootVehEnabled);
     config.BootVehFull = json.value("BootVehFull", config.BootVehFull);
     config.BootEnableEtw = json.value("BootEnableEtw", config.BootEnableEtw);
+    config.BootEnableGcServer = json.value("BootEnableGcServer", config.BootEnableGcServer);
     config.BootDotnetOpenProcessHookMode = json.value("BootDotnetOpenProcessHookMode", config.BootDotnetOpenProcessHookMode);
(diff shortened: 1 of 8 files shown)


KazWolfe:
Is this still desired/ready for review for API 10?

goaaats:
I would be extremely surprised if server GC would be advantageous for the type of work we do. Do you have some benchmarks or numbers?

rreminy:
Its difficult to actually make a benchmark whenever it comes to GC due to the YMMV nature, however I still did make a benchmark some time back then showing an allocation heavy workload where Server GC is excellent at:
- https://github.com/rreminy/GCOddities/tree/master/Results

In general while testing experiences been the following:
- On allocation-heavy workloads, Server GC perform way much better, eliminating most *(but not all)* hitches, however the remaining hitches tend to be much smaller, if any, and may be unnoticeable.
- On allocation-light workload however, while it performs as good or better than Workstation GC *(the default)*, it tends to accumulate a big collection every few minutes or so, which are harder to predict and may cause a hitch lasting about 200-400ms. This does not tend to happen during allocation heavy workloads.
- I read Server GC uses more memory, however during my testing the difference wasn't that big. This is likely another YMMV variable.

So Server GC mostly works as a work-around whenever you have plugins that are heavily allocating object, which can improve performance greatly where Workstation GC suffers. It still would be nice to have the option to enable it but this would be at most experimental and would need further testing.

This PR is technically ready but untested, however my other PR I got stuck because I couldn't get it to compile. I been testing using environment variables instead. You can monitor and change some GC parameters using my GC plugin:
- https://github.com/rreminy/GCPlugin

PR #1717, opened 2024-03-14
Add missing space in cross-world PF links

zeroeightysix:
Okay, I know this is a silly change, but I noticed while working with these links and it bugged me

For cross-world PF links, the game inserts a space between the text and icon:
image

Dalamud did not:
image


goaaats:
Not silly at all, we should really try to build these from localization data instead or just have game functions do it.

goaaats (approved):

Haselnussbomber:
FYI: kizer is working on a `SeStringEvaluator` ([work in progress](https://github.com/goatcorp/Dalamud/compare/master...Soreepeong:Dalamud:feature/sestring)), that will be part of the SeString ImGui renderer. With that, we can replace hardcoded strings like this and other manually-crafted SeStrings with the SeStrings from the sheets (Addon row 2265 in this case). It'll resolve payloads and expressions and populate any parameters with the values we pass etc. 🙂
Hopefully coming soon. 🙏

goaaats (approved):

PR #1718, opened 2024-03-15
Add tests for ReliableFileStorage

karashiiro:
Pretty self-explanatory, just tests for everything here, including a few for expected behaviors between different containers. I also added a broad concurrency test on top of everything else, just to catch any edge cases. Fortunately, none of these tests revealed any bugs in the existing implementation.

coverage
image


goaaats:
Thanks, appreciate it!

goaaats (approved):

PR #1720, opened 2024-03-15
Handle static declares in AddonLifecycleWidget

wolfcomp:
Just a fix to where if someone declares a static method for the event handler the `Delegate.Target` doesn't get the object out.

This switches it over to a `Type` inspection way of figuring out where the method is declared from.


MidoriKami:
Oh my apologies, I talked about how to fix this with Kizer and he guided me towards this solution, but I forgot to implement and PR it. Thanks!

goaaats (approved):

PR #1731, opened 2024-03-19
.NET 8

KazWolfe:


Dalamud/Plugin/Internal/PluginManager.cs
```diff
@@ -842,7 +844,8 @@ public void RemovePlugin(LocalPlugin plugin)
             this.installedPluginsList.Remove(plugin);
         }
 
-        PluginLocations.Remove(plugin.AssemblyName?.FullName ?? string.Empty, out _);
+        // NET8 CHORE
```
marzent, line 847: Do we want to remove these `// NET8 CHORE` comments, before merging?

KazWolfe: These look like they're killing off MonoMod, so it's not a blocker to release. We definitely will need to do a cleanup phase at some point though - will defer to Goat if he wants to do that before merging or after.

goaaats: We should do a pass for API 10.

goaaats (approved):

PR #1737, opened 2024-03-21
Add FontAwesome fixed width

rootdarkarchon:
This PR adds the `IconFontFixedWidthHandle` which is representing the FontAwesome font with a fixed width and vertically aligned height for each glyph.


goaaats:
@Soreepeong do you mind approving this if you think it's good?

Soreepeong:
Yeah this is fine; imo looks good enough to change later too if there arises any need to.

goaaats (approved):

Issue #1738, opened 2024-03-21
Notification SetIconTexture is not ergonomic


lojewalo:
The documentation for `IActiveNotification.SetIconTexture` states that the texture will be disposed on notification dismiss automatically.

I have a collection of a few icons I intend to set on a notification. Rather than parse the icons every single time I make a notification, I load them all at once at startup and cache them. On plugin dispose, I also dispose of the icons.

`SetIconTexture` makes this scheme impossible. Ideally, I would like to be able to set an icon multiple times throughout a notification's lifetime, so using `CreateWrapSharingLowLevelResource` is unwieldy. A `SetIconTexture`-derivative that marks the texture as not-to-be-disposed would be ideal.

Also unwieldy is the inability to compare the current icon texture to another texture. I see in the source that Dalamud has the ability to do so, so it would be nice if API consumers could do the same. Consider, for example, this code:

```cs
if (this.Icons.TryGetValue(state, out var texture)) {
    if (notif.Icon.Texture != texture) { // would love to be able to do this
        notif.SetIconTextureWithoutAutoDispose(texture);
    }
}
```

As a final aside, the documentation for `HardExpiry` is pretty heavy on the double-negatives:
> If neither INotification.HardExpiry nor INotification.InitialDuration is not MaxValue, then the notification will not expire after a set time.

Consider rewriting this to be more-easily parsed:

> If both INotification.HardExpiry and INotification.InitialDuration are MaxValue, then the notification will not expire after a set time.

Soreepeong:
The primary motivation for only supporting ownership transferal of textures was that the lifetime of texture would get confusing, especially when a notification may outlive a plugin. But now that I think about it, removing the textures for the notifications invoked by the plugin on plugin unload should work; I'll add `SetIconTexture( Task<IDTW?>|IDTW? texture, bool leaveOpen )` variant so that it will not dispose the texture upon the notification dismiss. Instead, if `leaveOpen` is set, the icon will be cleared if a notification is still visible after plugin is unloaded.

I'm thinking that texture comparison should be done by introducing a function instead of operators, as even for a single texture wrap there are various ways of consuming it. Does `bool ResourceEquals(this IDTW? left, IDTW? right);` sound acceptable, which compares the underlying D3D11 shader resource view?

Last one I'll fix that along with changing the rest.

lojewalo:
[...]

PR #1739, opened 2024-03-21
Add IconTexture/Wrap to INotification (#1738)

Soreepeong:
Notification record and IActiveNotification interface now supports setting or updating the texture wraps being used, and SetIconTexture has gotten more overloads to support leaveOpen mechanism that can commonly be found with Stream wrappers.

ImGui widget is updated to support testing setting "leaveOpen" and updating "IconTexture" property via setter, making it possible to check whether IDTW.Dispose is being called under given conditions.

Some changes to doccomments are made.

### DIFF
--- a/Dalamud/Interface/ImGuiNotification/INotification.cs
+++ b/Dalamud/Interface/ImGuiNotification/INotification.cs
@@ -22,19 +22,42 @@ public interface INotification
     /// <summary>Gets or sets the type of the notification.</summary>
     NotificationType Type { get; set; }
 
-    /// <summary>Gets or sets the icon source.</summary>
-    /// <remarks>Use <see cref="IActiveNotification.SetIconTexture(IDalamudTextureWrap?)"/> or
+    /// <summary>Gets or sets the icon source, in case <see cref="IconTextureTask"/> is not set or the task has faulted.
+    /// </summary>
+    INotificationIcon? Icon { get; set; }
+
+    /// <summary>Gets or sets a texture wrap that will be used in place of <see cref="Icon"/> if set.</summary>
+    /// <remarks>
+    /// <para>A texture wrap set via this property will <b>NOT</b> be disposed when the notification is dismissed.
+    /// Use <see cref="IActiveNotification.SetIconTexture(IDalamudTextureWrap?)"/> or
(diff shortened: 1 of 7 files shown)


Dalamud/Interface/ImGuiNotification/INotification.cs
```diff
@@ -22,19 +22,42 @@ public interface INotification
...
+    Task<IDalamudTextureWrap?>? IconTextureTask { get; set; }
 
     /// <summary>Gets or sets the hard expiry.</summary>
     /// <remarks>
     /// Setting this value will override <see cref="InitialDuration"/> and <see cref="ExtensionDurationSinceLastInterest"/>, in that
     /// the notification will be dismissed when this expiry expires.<br />
     /// Set to <see cref="DateTime.MaxValue"/> to make only <see cref="InitialDuration"/> take effect.<br />
-    /// If neither <see cref="HardExpiry"/> nor <see cref="InitialDuration"/> is not MaxValue, then the notification
+    /// If both <see cref="HardExpiry"/> and <see cref="InitialDuration"/> are MaxValue, then the notification
     /// will not expire after a set time. It must be explicitly dismissed by the user of via calling
```
sersorrel, line 61: while you're tidying docs:

```suggestion
    /// will not expire after a set time. It must be explicitly dismissed by the user or via calling
```

Soreepeong: :smh:

goaaats (commented):
How does this work when the plugin unloads, but still has active notifications, while owning the icon?

Soreepeong:
If icon ownership is not transferred, then the icon will be unset on plugin unload. See `RemoveNonDalamudInvocations`

goaaats:
Must have missed that, thanks. Wonder if that could race with the main thread rendering the icon, when the plugin is unloaded asynchronously?

Soreepeong:
(Draw) ... (CleanupImGuiResources)

NotificationManager touches a texture between `RenderImGui` and `CleanupImGuiResources`. If a plugin gets unloaded during that window, since DalamudTextureWrap is a deferred disposable, it will be disposed way after NM is done using it.

goaaats (approved):

PR #1747, opened 2024-03-24
DataWindow Enhancements

MidoriKami:
Subdivides DataWindow Widgets into IDebugWidget and IDataWidget, and provides a toggle to show or hide IDebugWidget's

We were starting to get a large number of widgets added, and this will help keep them manageable.

image

By default Debug Widgets are hidden, they can be enabled by clicking the bug icon.

The choice is saved to DalamudConfiguration so you won't have to enable it each time.

image

Debug widgets are identified with a bug icon at the end of the line.

Debug widgets are considered to be any widgets that don't simply display data, and/or are able to modify the state of the game.

image

Additionally provides a pop-out feature to pop widgets out into their own windows.

image

image


Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
```diff
@@ -235,4 +297,27 @@ private void Load()
...
+    private void OnDalamudConfigurationSaved(DalamudConfiguration dalamudConfiguration)
+    {
+        this.showDebugWidgets = dalamudConfiguration.DataShowDebug;
+    }
+    
+    private bool DrawToggleButtonWithTooltip(string buttonId, string tooltip, FontAwesomeIcon icon, ref bool enabledState)
+    {
+        var result = false;
+
+        var buttonEnabled = enabledState;
```
Infiziert90, line 310: No need to make a new variable for this bool, it isn't modified at all (neither does enabledState need to be a ref)

MidoriKami: This function is copied from ConsoleWindow, it's a generic helper function, there's no real need to make it specific to this window.

A more proper solution is to move the helper function to one of the imgui helper classes, but that's out of scope for this PR.

MidoriKami: Additionally, copying the ref bool is important here because it is used to push/pop style variables, and the ref'd value can potentially change between the push and the pop causing a crash/corruption issue.

Infiziert90: But it isn't modified at all, that's why the ref and variable are both unneeded instructions

If it is modified in the original function, okay, but in this copy they aren't touched

MidoriKami:
Is there enough interest in these features for me to resolve the merge conflicts?

KazWolfe:
I think there's interest in the pop-out widgets, though I'm not certain about hiding the so-called "debug widgets" as they still have plenty of significant usage.

At the very least, if we want to hide certain widgets, the UX will need to be modified to not just be an inscrutable icon button.

MidoriKami:
There is need to cut down on the UI clutter and so far no concrete suggestions have been made to address the problem.

If someone else wants to cut out the pop out feature into it's own PR I welcome them to do so.

MidoriKami:
Stale, closing. If someone wants to take part of all of this PR and resubmit it on its own, feel free.

PR #1748, opened 2024-03-25
Locale name standardize

wozaiha:
This commit changes the locale name to a more standardize state according to [ISO 15924](https://en.m.wikipedia.org/wiki/ISO_15924) which is the locale name options that other games/programs usually use.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/Settings/Widgets/LanguageChooserSettingsEntry.cs
+++ b/Dalamud/Interface/Internal/Windows/Settings/Widgets/LanguageChooserSettingsEntry.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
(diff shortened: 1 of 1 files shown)


goaaats (approved):

Soreepeong:
Ah no, leave Korean in English, or it'll display a `==` in the language list by default.

KazWolfe:
Holding this PR in general - if we’re fixing this, it would probably be a good idea to change from `tw` to whatever the proper localization name is (`zh-TW`? Unsure what .NET likes), alongside using the proper code for any other variants of Chinese. 

We’re looking at what it will take and will comment appropriately.

wozaiha:
I thought that Korean has been supported to show in Imgur since Korean fonts has been added?

Soreepeong:
It's only activated if the *current* language is set to Korean or the user has input Hangul via Korean IME at least once. Adding all 11152 characters at all times is pretty wasteful, and I don't think it's worth adding an exception of adding that 3 characters to the default atlas just to display the language name.

wozaiha:
Changed ko back to English for now.

KazWolfe:
Closing in favor of #1749. If you need any extra changes, please let me know there.

Thanks!

PR #1749, opened 2024-03-25
Change Chinese Language Localization Names

KazWolfe:
Supersedes #1748.

Change `tw` to be listed as 中文（繁體） in the UI, and `zh` to 中文（简体）. Also use proper capitalization for all languages in the list.

image

### DIFF
--- a/Dalamud/Localization.cs
+++ b/Dalamud/Localization.cs
@@ -70,14 +70,15 @@ public class Localization : IServiceType
     public CultureInfo DalamudLanguageCultureInfo { get; private set; }
 
     /// <summary>
-    /// Gets an instance of <see cref="CultureInfo"/> that corresponds to <paramref name="langCode"/>.
+    /// Gets an instance of <see cref="CultureInfo"/> that corresponds to a Dalamud <paramref name="langCode"/>.
     /// </summary>
     /// <param name="langCode">The language code which should be in <see cref="ApplicableLangCodes"/>.</param>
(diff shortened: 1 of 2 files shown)


Dalamud/Localization.cs
```diff
@@ -70,14 +70,15 @@ private Localization(Dalamud dalamud, DalamudConfiguration configuration)
     public CultureInfo DalamudLanguageCultureInfo { get; private set; }
 
     /// <summary>
-    /// Gets an instance of <see cref="CultureInfo"/> that corresponds to <paramref name="langCode"/>.
+    /// Gets an instance of <see cref="CultureInfo"/> that corresponds to a Dalamud <paramref name="langCode"/>.
     /// </summary>
     /// <param name="langCode">The language code which should be in <see cref="ApplicableLangCodes"/>.</param>
     /// <returns>The corresponding instance of <see cref="CultureInfo"/>.</returns>
     public static CultureInfo GetCultureInfoFromLangCode(string langCode) =>
```
goaaats, line 77: What is this for anyway, should we just deprecate it and change the codes to something that can differentiate them properly for API 10?

KazWolfe: This translates our language codes to something `CultureInfo` can work wiht more precisely. It works around our prior language code choices.

Ideally we would deprecate this come API 10, but that will also require plugin devs mitigate on their end (and possibly update a lot of translation files/infra that uses this assumption).

goaaats (approved):

PR #1758, opened 2024-04-05
Sanitize IME strings to UCS-2

Soreepeong:
Our ImGui has ImWChar defined as ushort, and it does not support UCS-2. ImGui would filter the input characters to UCS-2 range, but our IME implementation would forcefully set the buffer for the text input. This filters all candidate and composition strings to fit in UCS-2 range, notifying the user that some candidates are not properly supported.

image


goaaats:
Is this something we should fix in our ImGui natives? If so, should we add a comment explaining the reason for the conversion?

Soreepeong:
This is strictly a problem in our IME integration with ImGui; we defined ImWChar to be uint16 and that makes adding a codepoint that goes outside that range going against the assumptions ImGui is making.

goaaats (approved):

PR #1761, opened 2024-04-09
Replace U+202F with Indent payload

Infiziert90:
This is a fix for users that have the French number format using U+202F as thousand split.

The PR checks each message text payload if it contains any U+202F and replaces each of them with the Indent payload (0x1D) to render correctly in chat.


goaaats (approved):

goaaats (approved):
fix build

goaaats (dismissed):

goaaats (approved):

PR #1769, opened 2024-04-13
Wine stability improvements

marzent:
This uses MinHook for the Addon Hooks, as there seems to be a race with Reloaded somewhere that leads to execution of random code after a while.

Also makes `DALAMUD_LEGACY_CORRUPTED_STATE_EXCEPTION` configurable, which is highly undesirable to have enabled with Wine, as this can mask underlying issues and lead to random crashes as well.


KazWolfe:
n.b. holding this PR as we're allegedly seeing some issues with MinHook with certain plugins (including Simple Tweaks). See #linux-and-dev channel in the Dalamud Discord.

We're also looking at how best to handle `DALAMUD_LEGACY_CORRUPTED_STATE_EXCEPTION` since *ideally* we'd disable this on Windows as well, but this may break certain plugin behaviors.

goaaats:
We definitely can't "just" turn it off, we'll have to have some kind of migration plan. Either way it's a hard sell to devs, we'll talk about it internally, it's good to have the option for now though.

sersorrel:
> I dropped the hook changes part for now...

is the thinking that the problems those changes addressed are now fixed in a different way, or will be fixed with Dawntrail, or something else?

marzent:
I believe it to be an issue somewhere inside reloaded for stacked hooks; am currently trying to come up with a better solution.

MidoriKami:
On a related note, I'm working on rewriting AddonLifecycle to only hook when an event listener is actually active or upon request.

KazWolfe:
@marzent Is this still applicable?

PR #1778, opened 2024-04-20
Expose DTR entries in public interface

### LINKED ISSUE #1777: [RFC] Proposal to DtrBar API Enhancements
This proposal would make it possible for plugin developers to create plugins 
that would effectively replace the native DTR bar from the game with a custom
implementation by reading the registered DTR bar entries from Dalamud and
using a custom UI to display them.

In order to keep it streamlined and not conflict with Dalamud's existing config
options when it comes to DTR bar entry visibility and ordering, the exposed API
solely focuses on providing read-only access to the DTR bar entries.

Due to the specific use-case of this API and its immutability, it is
recommended to use a separate service interface that plugins can consume in
order to gain access to the DTR bar entries.

### `IDtrBarEntries`

The `IDtrBarEntries` interface provides read-only access to the DTR bar
entries, as well as the ability to subscribe to changes in the DTR bar entries.

```csharp
public interface IDtrBarEntries
{
    /// <summary>
    /// Gets the DTR bar entries.
    /// </summary>
    public IReadOnlyList<ReadOnlyDtrBarEntry> Entries { get; }

    /// <summary>
    /// Invoked when an entry has been added.
    /// </summary>
    public event Action<ReadOnlyDtrBarEntry> EntryAdded;
    
    /// <summary>
    /// Invoked when an entry has been removed.
    /// </summary>
    public event Action<ReadOnlyDtrBarEntry> EntryRemoved;
}
```

### `ReadOnlyDtrBarEntry`

While the existing `DtrBarEntry` class is mutable and can be modified by
a plugin that registered it, the `ReadOnlyDtrBarEntry` class is immutable and
is intended to be shared with consuming plugins. This ensures that the original
DTR bar entry cannot be modified by any other plugin than the one that
originally registered it.

```csharp
public class ReadOnlyDtrBarEntry
{
    /// <summary>
    /// Gets the name of the DTR bar entry.
    /// </summary>
    public string Name { get; }

    /// <summary>
    /// Gets the tooltip of the DTR bar entry.
    /// </summary>
    public string Tooltip { get; }

    /// <summary>
    /// Gets the visibility of the DTR bar entry.
    /// </summary>
    public bool IsVisible { get; }

    /// <summary>
    /// True if the DTR bar entry is interactive, meaning it has an OnClick
    /// event listener attached to it.
    /// </summary>
    public bool IsInteractive { get; }
    
    /// <summary>
    /// Gets the render-order of the DTR bar entry.
    /// </summary>
    public int SortIndex { get; }
    
    /// <summary>
    /// Invokes the OnClick event of the DTR bar entry.
    /// </summary>
    public void Click();
}
```

### Service Behavior

When a plugin registered a DTR bar entry, the `EntryAdded` event is invoked and
the `ReadOnlyDtrBarEntry` object will be part of the `Entries` list. When a
plugin unregisters a DTR bar entry, the `EntryRemoved` event is invoked and the
`ReadOnlyDtrBarEntry` object will no longer be part of the `Entries` list.

Using these events allows a plugin to only update its own state or UI when it
is necessary, instead of polling the DTR bar entries list for changes on every
frame.

> [!NOTE]
> A user can disable specific DTR bar entries in Dalamud's settings. In this case,
> the `ReadOnlyDtrBarEntry` object will no longer be part of the `Entries` list.
> The `EntryRemoved` event will be invoked if it was previously part of the list.
> The `EntryAdded` event will be invoked if it was not previously part of the list.
> This implicitly forces plugins to behave correctly when it comes to handling
> visibility of individual DTR bar entries.

### Usage Example

```csharp

internal class MyPlugin : IDalamudPlugin
{
    private readonly DalamudPluginInterface _pluginInterface;
    private readonly IDtrBarEntries _dtrBarEntries;
    
    public MyPlugin(DalamudPluginInterface pluginInterface, IDtrBarEntries dtrBarEntries)
    {
        _pluginInterface = pluginInterface;
        _dtrBarEntries = dtrBarEntries;
        
        _pluginInterface.UiBuilder.Draw += OnDraw;
        
        dtrBarEntries.EntryAdded += OnEntryAdded;
        dtrBarEntries.EntryRemoved += OnEntryRemoved;
    }
    
    public Dispose()
    {
        _pluginInterface.UiBuilder.Draw -= OnDraw;
        
        dtrBarEntries.EntryAdded -= OnEntryAdded;
        dtrBarEntries.EntryRemoved -= OnEntryRemoved;    
    }
    
    private void OnEntryAdded(ReadOnlyDtrBarEntry entry)
    {
        // Handle entry added
    }
    
    private void OnEntryRemoved(ReadOnlyDtrBarEntry entry)
    {
        // Handle entry removed
    }
    
    private void OnDraw()
    {
        // Draw DTR bar entries directly from the exposed list.
        foreach (var entry in _dtrBarEntries.Entries)
        {
            // Draw entry
        }   
    }
}
```

goaaats:
Closes #1777

### DIFF
--- a/Dalamud/Game/Gui/Dtr/DtrBarEntry.cs
+++ b/Dalamud/Game/Gui/Dtr/DtrBarEntry.cs
@@ -6,9 +6,78 @@ using FFXIVClientStructs.FFXIV.Component.GUI;
 namespace Dalamud.Game.Gui.Dtr;
 
 /// <summary>
+/// Interface representing a read-only entry in the server info bar.
+/// </summary>
+public interface IReadOnlyDtrBarEntry
+{
+    /// <summary>
+    /// Gets the title of this entry.
+    /// </summary>
+    public string Title { get; }
+    
+    /// <summary>
+    /// Gets a value indicating whether this entry has a click action.
+    /// </summary>
+    public bool HasClickAction { get; }
+    
+    /// <summary>
+    /// Gets the text of this entry.
+    /// </summary>
+    public SeString Text { get; }
+    
+    /// <summary>
+    /// Gets a tooltip to be shown when the user mouses over the dtr entry.
+    /// </summary>
+    public SeString Tooltip { get; }
+    
+    /// <summary>
+    /// Gets a value indicating whether this entry is visible.
+    /// </summary>
+    public bool Shown { get; }
+    
+    /// <summary>
+    /// Triggers the click action of this entry.
+    /// </summary>
+    /// <returns>True, if a click action was registered and executed.</returns>
+    public bool TriggerClickAction();
+}
+
+/// <summary>
+/// Interface representing an entry in the server info bar.
+/// </summary>
+public interface IDtrBarEntry : IReadOnlyDtrBarEntry
+{
+    /// <summary>
+    /// Gets or sets the text of this entry.
+    /// </summary>
+    public new SeString? Text { get; set; }
+    
+    /// <summary>
+    /// Gets or sets a tooltip to be shown when the user mouses over the dtr entry.
+    /// </summary>
+    public new SeString? Tooltip { get; set; }
+    
+    /// <summary>
+    /// Gets or sets a value indicating whether this entry is visible.
+    /// </summary>
+    public new bool Shown { get; set; }
+    
+    /// <summary>
+    /// Gets or sets a action to be invoked when the user clicks on the dtr entry.
+    /// </summary>
+    public Action? OnClick { get; set; }
+    
+    /// <summary>
+    /// Remove this entry from the bar.
+    /// You will need to re-acquire it from DtrBar to reuse it.
+    /// </summary>
+    public void Remove();
+}
+
+/// <summary>
 /// Class representing an entry in the server info bar.
 /// </summary>
-public sealed unsafe class DtrBarEntry : IDisposable
+public sealed unsafe class DtrBarEntry : IDisposable, IDtrBarEntry
 {
     private bool shownBacking = true;
     private SeString? textBacking = null;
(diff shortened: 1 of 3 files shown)


Dalamud/Game/Gui/Dtr/DtrBarEntry.cs
```diff
@@ -5,10 +5,79 @@
...
+    
+    /// <summary>
+    /// Gets a tooltip to be shown when the user mouses over the dtr entry.
+    /// </summary>
+    public SeString Tooltip { get; }
+    
+    /// <summary>
+    /// Gets a value indicating whether this entry is visible.
+    /// </summary>
+    public bool Shown { get; }
```
haroldiedema, line 36: This one seems to be only managed by the owner of the entry. 

The reason for a separate `IsVisible` flag is because visibility can also be modified via `/xlsettings` by adding some entries to the ignore list.

The `DtrBarService` iterates over the entries, then compares if an entry is in an _ignore-list_ to then just skip over it. 
The `Shown` flag remains `true`, even though the entry is invisible to the user. See https://github.com/goatcorp/Dalamud/blob/bac42ac3f139482fc0062447d364fd06decc583f/Dalamud/Game/Gui/Dtr/DtrBar.cs#L202

Ideally, plugins should have access to a property that exactly mimics the `isHide` variable that is used there.

haroldiedema (changes requested):

Dalamud/Game/Gui/Dtr/DtrBarEntry.cs
goaaats: Added a separate prop for that.

haroldiedema (approved):
Amazing stuff! :+1:

haroldiedema (approved):

PR #1780, opened 2024-04-20
Update Lumina to 3.17.0

Haselnussbomber:
This PR updates Lumina to 3.17.0, which includes the following changes:

**https://github.com/NotAdam/Lumina/pull/78:**
Reverts the output of `SeString.ToString()` back to `RawString`, as the change was causing some confusion.
The macro code style output, which was introduced in 3.16.0, was moved to new method `ToMacroString()` instead.

**https://github.com/NotAdam/Lumina/pull/79:**
Improves the performance of `GetFile<T>` when using multiple threads.

> [!IMPORTANT]
> - This adds a dependency on v9.0.0-preview.1.24081.5 of `Microsoft.Extensions.ObjectPool`.  
    Dalamud currently uses v8.0.1, which would be a downgrade. I updated it to use the same version (in 1c34267f2f0dbdc1b2c656ae231fa33d280ccaa3) so that it will build successfully.
> - `GameData` implements `IDisposable` now, so I added a Dispose call to `DataManager.DisposeService` (561ee89ac3d1dc52d2f637b6e55c277cb8e2d010).

**https://github.com/NotAdam/Lumina/pull/80:**
Fixes the output of `BasePayload.ToString()` for StringExpressions, a bug introduced with PR 78 above.

**https://github.com/NotAdam/Lumina/pull/81:**
Adds a new efficient way to read and build SeStrings.
With this change, Lumina now also targets net8.0.

---

Full list of changes: https://github.com/NotAdam/Lumina/compare/3.16.0...3.17.0

### DIFF
--- a/Dalamud.CorePlugin/Dalamud.CorePlugin.csproj
+++ b/Dalamud.CorePlugin/Dalamud.CorePlugin.csproj
@@ -27,7 +27,7 @@
     </ItemGroup>
 
     <ItemGroup>
-        <PackageReference Include="Lumina" Version="3.16.0" />
+        <PackageReference Include="Lumina" Version="3.17.0" />
         <PackageReference Include="Lumina.Excel" Version="6.5.2" />
         <PackageReference Include="Newtonsoft.Json" Version="13.0.2" />
         <PackageReference Include="StyleCop.Analyzers" Version="1.2.0-beta.333">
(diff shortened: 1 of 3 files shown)


goaaats:
Does it depend on a .NET9 preview of that lib on purpose? Seems odd to me, but if it doesn't cause any issues, eh

Haselnussbomber:
I'm not sure whats up with that. All I found out is:

- the sources were last updated Nov 9, 2023: https://github.com/dotnet/aspnetcore/tree/main/src/ObjectPoo
- v8.0.1 was released on Jan 9, 2024: https://github.com/dotnet/aspnetcore/releases/tag/v8.0.1
- and apicompat couldn't find any breaking changes between v8.0.1 (net8.0) and v9.0.0-preview.1.24081.5 (net9.0).

🤷‍♂️

Haselnussbomber:
I assume @Soreepeong did it by accident.
(Also, I'm just here for the update and have no affiliation with Lumina. 🙂)

goaaats (approved):

PR #1786, opened 2024-04-22
[master] Update ClientStructs

bleatbot:


wolfcomp:
Gets flagged as Incompatible API due to member function signature change in `InfoModule` and `AgentModule` this is purely done due to new exporter so we have enum names on the functions instead of number but the other function still exists.

KazWolfe (approved):

PR #1793, opened 2024-04-28
Fix RGB returning wrong color and change it to RGBA

Infiziert90:
The current function returns a color with the red channel set to FF instead the actual value, 
so either the function uses `UIForeground >> 8` or it returns the valid RGBA value, as none of the sheet colors has any real alpha.

For usage in ImGui you would also need the alpha channel back, so directly returning RGBA seems more useful.

Edit:
Also added a helper for converting from RGBA to ABGR, and expose these values directly in the payloads for easy usage with imgui

### DIFF
--- a/Dalamud/Utility/Util.cs
+++ b/Dalamud/Utility/Util.cs
@@ -477,12 +477,12 @@ public static class Util
             case "MacOS": return OSPlatform.OSX;
             case "Linux": return OSPlatform.Linux;
         }
-        
+
         // n.b. we had some fancy code here to check if the Wine host version returned "Darwin" but apparently
         // *all* our Wines report Darwin if exports aren't hidden. As such, it is effectively impossible (without some
         // (very cursed and inaccurate heuristics) to determine if we're on macOS or Linux unless we're explicitly told
         // by our launcher. See commit a7aacb15e4603a367e2f980578271a9a639d8852 for the old check.
-        
+
         return IsWine() ? OSPlatform.Linux : OSPlatform.Windows;
     }
 
@@ -545,7 +545,7 @@ public static class Util
                     }
                 }
             }
-        } 
+        }
         finally
         {
             foreach (var enumerator in enumerators)
@@ -586,7 +586,7 @@ public static class Util
     {
         WriteAllTextSafe(path, text, Encoding.UTF8);
     }
-    
+
     /// <summary>
     /// Overwrite text in a file by first writing it to a temporary file, and then
     /// moving that file to the path specified.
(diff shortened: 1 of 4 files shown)


haroldiedema:
Would you mind adding an ABGR "helper" getter to this as well? That way not every plugin needs to do the conversion themselves anymore. The method you posted on Discord yesterday works well for this case.

haroldiedema (approved):
Nice! 🥳

Dalamud/Utility/Util.cs
```diff
@@ -634,13 +634,24 @@ public static unsafe void WriteAllBytesSafe(string path, byte[] bytes)
...
+    /// <summary>
+    /// Turns a RGBA color value into ABGR for usage in ImGui.PushColor
+    /// </summary>
+    /// <param name="rgba">Color value with byte order of RGBA.</param>
+    /// <returns>The color value in byte order of ABGR, the format that ImGui uses.</returns>
+    public static uint RGBAToABGR(uint rgba)
+    {
+        var tmp = ((rgba << 8) & 0xFF00FF00) | ((rgba >> 8) & 0xFF00FF);
+        return (tmp << 16) | (tmp >> 16);
+    }
```
Haselnussbomber, line 653: That's just an endian swap. .NET already provides [BinaryPrimitives.ReverseEndianness](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.binary.binaryprimitives.reverseendianness). I remember @KazWolfe linking it in Discord.

Infiziert90: Yeah it is just an endian swap, if wanted i can use the .net method, tho i think having a Util explicitly named by what it is used for in dalamud can provide some benefits

Haselnussbomber: Yeah, I get it. Although it can also be used as ABGRToRGBA. 🤷‍♂️ It's up to Team Dalamud, I guess. 😄

Haselnussbomber: Besides, I think this would fit better in [Dalamud/Interface/ColorHelpers.cs](https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Interface/ColorHelpers.cs). 🙂

KazWolfe: I think a utility function is fine, but please do use the endian swap rather than re-implement our own. I'd probably want a better name though, seeing as this is technically bidirectional.

Infiziert90: Which name would you suggest instead?

KazWolfe (approved):

PR #1795, opened 2024-04-29
DalamudPluginInterface functions for opening Settings and DevMenu

meoiswa:
Changes to DalamudPluginInterface:

- Adds `OpenDalamudSettings()`
- Adds `OpenDeveloperMenu()`
- Changes `OpenPluginInstaller` signature to `OpenPluginInstaller(bool filterByPluginInternalName = true)` so opening the window without filtering by plugin name is possible, without breaking backwards compatibility

### DIFF
--- a/Dalamud/Interface/Utility/Raii/EndObjects.cs
+++ b/Dalamud/Interface/Utility/Raii/EndObjects.cs
@@ -1,4 +1,5 @@
 using System.Numerics;
+using System.Text;
 
 using ImGuiNET;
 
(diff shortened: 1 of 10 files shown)


MidoriKami:
Might consider changing OpenDalamudSettings to also open it to a specific tab within DalamudSettings, as I assume this would be for something like "Configure Dalamud DTR bar settings here [button]" for example.

Dalamud/Interface/Utility/Raii/EndObjects.cs
```diff
@@ -103,8 +104,43 @@ public static IEndObject TabBar(string label, ImGuiTabBarFlags flags)
     public static IEndObject TabItem(string label)
         => new EndConditionally(ImGui.EndTabItem, ImGui.BeginTabItem(label));
 
-    public static unsafe IEndObject TabItem(byte* label, ImGuiTabItemFlags flags)
```
Infiziert90, line 106: This change breaks a plugin on stg, AllaganTools in this case

PR #1802, opened 2024-05-04
Fix DutyCategory names

zeroeightysix:
To reflect what's in the game now. This is also the layout of the enum as remote-party-finder currently [sees it](https://github.com/zeroeightysix/remote-party-finder/blob/fb5198b9113b4659009da8a248cdb4f6d1fd1e58/server/src/listing.rs#L235)

* Quest battles can no longer be recruited for. Hence I've just removed the variant - any usages of it should be wrong
  * In its place has come the Gold Saucer category
* Adventuring Forays I've deprecated as it's still the same content, but since a patch in shb is called Field Operations


zeroeightysix:
From a quick github search, I think only BetterPartyFinder is affected by this change (removal of QuestBattles)

https://github.com/sersorrel/BetterPartyFinder/blob/658b8d710435421118ca4e84193ed62bfd8cdca9/BetterPartyFinder/PluginUi.cs#L727

MidoriKami:
If you want to pr a breaking change, pr it to the ApiX branch, else mark the current enum obsolete, and add the new value for quest battles.

zeroeightysix:
> If you want to pr a breaking change, pr it to the ApiX branch, else mark the current enum obsolete, and add the new value for quest battles.
> 
> Dalamud maintainers don't care about the scope of a breakage, if it's a API break they won't allow it.

Thanks. Didn't know - changed base.

PR #1805, opened 2024-05-09
[master] Update ClientStructs

bleatbot:


KazWolfe (approved):
For methods like the below, would it be possible to make the fact that they're an `out` var clearer (e.g. adding `out_` as a prefix)?

PR #1807, opened 2024-05-13
Make DalamudPluginInterface into IDalamudPluginInterface

Critical-Impact:
- Moves LanguageChangedDelegate and ActivePluginsChangedDelegate into the interface which will break plugins relying on them
- Adds a IDisposable but only on DalamudPluginInterface, called by LocalPlugin do unregister the events

### DIFF
--- a/Dalamud.CorePlugin/PluginImpl.cs
+++ b/Dalamud.CorePlugin/PluginImpl.cs
@@ -55,7 +55,7 @@ namespace Dalamud.CorePlugin
         /// </summary>
         /// <param name="pluginInterface">Dalamud plugin interface.</param>
         /// <param name="log">Logging service.</param>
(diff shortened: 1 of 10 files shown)


Critical-Impact:
Rebased and I have interfaced UiBuilder, while I don't think it's perfect and still a little leaky it should be enough of a stop gap until we split the font handling/etc into it's own service

The impact is very little, internally some services do have to refer to a LocalUiBuilder but that's because they need to access internal properties of the real UiBuilder. I don't like it but I think it does highlight parts of dalamud that could use some refactoring

KazWolfe (approved):

KazWolfe:
Approved for merge with #1808 and after builds pass. Same comments from that PR apply w/r/t possible over-interfacing.

goaaats (approved):

PR #1808, opened 2024-05-14
Add interfaces to non public/sealed classes referenced in public interfaces

Critical-Impact:
Interfaces now available as they are part of public APIs and should themselves be interfaced:
- AetheryteEntry(IAetheryteEntry)
- AetheryteList(IAetheryteList)
- BuddyMember(IBuddyMember)
- BuddyList(IBuddyList)
- GameObject(IGameObject)
- PlayerCharacter(IPlayerCharacter)
- BattleChara(IBattleChara)
- Character(ICharacter)
- Fate(IFate)
- PartyMember(IPartyMember)
- CommandInfo(ICommandInfo)
- MenuItem(IMenuItem)
- MenuArgs(IMenuArgs)
- MenuItemClickedArgs(IMenuItemClickedArgs)
- MenuOpenedArgs(IMenuOpenedArgs)
- DtrBarEntry(IDtrBarEntry)
- PartyFinderListing(IPartyFinderListing)
- PartyFinderListingEventArgs(IPartyFinderListingEventArgs)
- TitleScreenMenuEntry(ITitleScreenMenuEntry and IReadOnlyTitleScreenMenuEntry)

Changes:
- CommandManager now tracks the assembly from which a command was loaded for display later
- MenuItem no longer has a function to calculate the PrefixedName, it has been moved to ContextMenu as MenuItem is public and it should up to the service to determine the name and not the MenuItem
- The scoped TitleScreenMenu service returns read only entries now. This was done to create a seperation between what should be accessible to the plugin and what is required by dalamud.

I can pull out individual changes or redo anything if required. Probably the one I'm the most iffy about is CommandManager.


KazWolfe:
Can you please provide justification for this PR so we have something on record?

Critical-Impact:
Sure, the justification is that these are already part of public facing apis. Interface or not they are already exposed to plugin developers, they really should have interfaces. 

On top of that with interfaces added to them it makes it easier to create mock or test versions of them for unit/integration tests.

https://discord.com/channels/581875019861328007/860813266468732938/1239514326335881278
That's also roughly where it was discussed

KazWolfe:
Assigning goat to this to see if this is something we think would be beneficial further. 

I'm a bit hesitant to merge this as-is, since a ton of these interfaces are one-shots for data classes which [seems a bit unnecessary](https://softwareengineering.stackexchange.com/q/390084). The intent here seems to be related to mocking Dalamud, but this feels somewhat odd given we don't really have any internal unit tests against our own code for our own modules.

Critical-Impact:
I mean you are partially right, but these parts of dalamud are exposed to plugins already, the only reason I'm adding interfaces to these is because plugins might want to mock them for their own unit tests. Without interfaces that's impossible(as a lot of them are internal classes with no way to construct). 

In regards to your point about data classes not needing to be interfaced, my argument would be that it gives us control later to add to the implementation classes without worrying about accidentally exposing things we might not. On top of that we can rework the internals without having to worry about breaking compatibility.


But yes perhaps goat has some further thoughts

KazWolfe:
> In regards to your point about data classes not needing to be interfaced, my argument would be that it gives us control later to add to the implementation classes without worrying about accidentally exposing things we might not. On top of that we can rework the internals without having to worry about breaking compatibility.

I think YAGNI is going to come into play here. We're tentatively scheduling API bumps for every major patch (so 7.1, 7.2, etc.), and we have API compatibility tooling to catch specific issues now. Furthermore, an interface will not save us from a behavioral change which can also be just as much of a headache. I do agree that not exposing new things that _shouldn't_ be exposed would be a decent benefit, but I'm unsure how likely it would be to come up. 

I think all of these _are_ fair points and that in an ideal world they would be nice to have, but that does become an issue especially for the sake of unit testing. I think Dalamud has a testing problem, and there *probably* needs to be some better way of fixing said testing problem. I'm just unsure littering the codebase with interfaces is really going to be as beneficial as we'd think (and even so - a lot of those tests should *probably* be in Dalamud).

Critical-Impact:
I definitely have no issue with classes that I've interfaced that can be constructed from plugins already, I can definitely roll those back and leave them as it. 

Classes that cannot be constructed and are returned by an interface should either have an interface on them or be wrapped in a class that has an interface. Interfaces are a contract and if I can't actually fulfill the contract with an alternative implementation of said contract then why have the interface at all?

This isn't just YAGNI, I am relying on a lot of this already to develop plugins without needing the game running(See DalaMock). Being able to construct these services in full without having wrappers would mean that a majority of plugins could attach a nuget package, add a project for a mock and boot their plugin with minimal changes. For working on IMGUI it is a massive boon to be able to quickly iterate and see changes to the UI. Being parked in the game to develop is kind of cumbersome especially if you are say away traveling and want to develop on a PC that does not have the horsepower to run the game.

So I suppose yes I am doing it with a specific purpose in mind but I find it odd to provide interfaces for the plugin based services if the ultimate goal is not to allow for them to be substituted(at least properly)

I don't really have anything else to add, I feel we've both expressed our opinions for and against and I can definitely see where you are coming from, I'd just disagree that there's no benefit from doing this.

goaaats (changes requested):
I think this is fine to go in,  I guess it's a natural consequence of the interface work in general and will make refactors easier in the future. I generally agree with Kaz's assessment, pure data classes probably don't need this in a lot of cases but it doesn't really hurt for the aforementioned reason.
Some nitpicks:
* Docs are duplicated now. The implementation should use ``<inheritdoc/>``, if it's not too much trouble.
* The implementations are still public. Is that intentional?

Critical-Impact:
Too easy @goaaats 

Things I've done:

1. Rebased and reworked any of the interfaces affected.
2. Put in inheritdoc where possible or where it made sense. The class/interface do have differing summaries so I left those as is.
3. Fixed a couple of issues I found with how I had interfaced the gameobject/player character inheritance tree
4. I haven't reverted the changes I made to the the data classes as it sounds like you are ok but if you need me to just let me know
5. Made the now interfaced implementations internal however I do need your thoughts on 6
6. Both GameObject and Fate have 3 static methods, IsValid, == and !=(the latter 2 being operator overloads). 
IsValid feels like a shortcut method and perhaps needs to be exposed elsewhere.
For Fate I believe it could go into FateTable, GameObject I'm not really sure
The operators, no idea, again they feel like shortcuts that maybe people are using but I see no references internally in dalamud

MidoriKami:
This is somewhat unrelated, but if this is going through, could you add a few more properties to IFate, ClientStructs has things like the IconIds that dalamud is missing at the moment.

Critical-Impact:
> This is somewhat unrelated, but if this is going through, could you add a few more properties to IFate, ClientStructs has things like the IconIds that dalamud is missing at the moment.

Not a bad idea, have pushed a commit adding in the fields that were missing in IFate and Fate

Critical-Impact:
Have rebased

goaaats (approved):

PR #1814, opened 2024-05-25
AutoTranslatePayload: Add public property for group and key

Infiziert90:
For use in plugins that require both, the resolved text and building parts of the original string parameter

### DIFF
--- a/Dalamud/Game/Text/SeStringHandling/Payloads/AutoTranslatePayload.cs
+++ b/Dalamud/Game/Text/SeStringHandling/Payloads/AutoTranslatePayload.cs
@@ -16,22 +16,10 @@ public class AutoTranslatePayload : Payload, ITextProvider
     private string text;
 
     [JsonProperty]
-    private uint group;
+    public uint Group { get; private set; }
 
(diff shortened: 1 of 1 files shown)


Dalamud/Game/Text/SeStringHandling/Payloads/AutoTranslatePayload.cs
```diff
@@ -21,6 +21,18 @@ public class AutoTranslatePayload : Payload, ITextProvider
...
+    /// Gets the sheet group this entry belongs to
+    /// </summary>
+    [JsonIgnore]
+    public uint Group => this.group;
+
+    /// <summary>
+    /// Gets the sheet key for this entry
+    /// </summary>
+    [JsonIgnore]
+    public uint Key => this.key;
```
KazWolfe, line 34: Why not just replace the fields with these properties and use the relevant autos? The fields are `private` already so that shouldn't be a breaking change.

KazWolfe (dismissed):
good if it builds

KazWolfe (approved):

PR #1820, opened 2024-05-31
IGameInteropProvider Additional Overloads

MidoriKami:
Also uses `nint` instead of `IntPtr`

These overloads make it easier to use these functions without requiring the user to cast to `nint` every call manually.

### DIFF
--- a/Dalamud/Hooking/Internal/GameInteropProviderPluginScoped.cs
+++ b/Dalamud/Hooking/Internal/GameInteropProviderPluginScoped.cs
@@ -47,7 +47,7 @@ internal class GameInteropProviderPluginScoped : IGameInteropProvider, IInternal
     }
 
     /// <inheritdoc/>
-    public Hook<T> HookFromFunctionPointerVariable<T>(IntPtr address, T detour) where T : Delegate
+    public Hook<T> HookFromFunctionPointerVariable<T>(nint address, T detour) where T : Delegate
     {
         var hook = Hook<T>.FromFunctionPointerVariable(address, detour);
         this.trackedHooks.Add(hook);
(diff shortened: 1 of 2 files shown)


MidoriKami:
I'm assuming these are a non issue as they are additive?
```
  API compatibility errors between 'left\Dalamud.dll' (left) and 'right\Dalamud.dll' (right):
  CP0006: Cannot add interface member 'Dalamud.Hooking.Hook<T> Dalamud.Plugin.Services.IGameInteropProvider.HookFromAddress<T>(System.UIntPtr, T, Dalamud.Plugin.Services.IGameInteropProvider.HookBackend)' to right\Dalamud.dll because it does not exist on left\Dalamud.dll
  CP0006: Cannot add interface member 'Dalamud.Hooking.Hook<T> Dalamud.Plugin.Services.IGameInteropProvider.HookFromAddress<T>(void*, T, Dalamud.Plugin.Services.IGameInteropProvider.HookBackend)' to right\Dalamud.dll because it does not exist on left\Dalamud.dll
  API breaking changes found. If those are intentional, the APICompat suppression file can be updated by specifying the '--generate-suppression-file' parameter.
Error: Dalamud.dll did not pass. Please review it for problems.
```

Seems kinda silly to error on additions.

KazWolfe:
> Seems kinda silly to error on additions.

Yeah, they're fine since we don't particularly care about this case. Adding new members to an interface is a breaking change, as it means everyone else who implements those interfaces must also update their code. See:

> ❌ DISALLOWED: Adding a member to an interface
> 
> If you [provide an implementation](https://learn.microsoft.com/en-us/dotnet/csharp/advanced-topics/interface-implementation/default-interface-methods-versions), adding a new member to an existing interface won't necessarily result in compile failures in downstream assemblies. However, not all languages support default interface members (DIMs). Also, in some scenarios, the runtime can't decide which default interface member to invoke. For these reasons, adding a member to an existing interface is considered a breaking change.
> <sub>[source: Library Change Rules](https://learn.microsoft.com/en-us/dotnet/core/compatibility/library-change-rules)</sub>

MidoriKami:
I hadn't actually considered this `Adding new members to an interface is a breaking change, as it means everyone else who implements those interfaces must also update their code.` as it was intended for just Dalamud to implement the interfaces.
RIP Dalamock I guess =D

KazWolfe (approved):

PR #1821, opened 2024-05-31
[apiX] dalamud fixes for CS updates

aers:
Updates CS to latest & fixes Dalamud build errors.

Since CS is still receiving the odd field rename/struct move we don't need to merge this just yet, its mostly for keeping track of updates.

### DIFF
--- a/Dalamud/Game/Addon/Events/AddonEventManager.cs
+++ b/Dalamud/Game/Addon/Events/AddonEventManager.cs
@@ -173,14 +173,14 @@ internal unsafe class AddonEventManager : IInternalDisposableService
     {
         try
         {
-            var atkStage = AtkStage.GetSingleton();
+            var atkStage = AtkStage.Instance();
             
             if (this.cursorOverride is not null && atkStage is not null)
             {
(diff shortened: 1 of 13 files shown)


aers:
Latest commit changes BattleChara.CastTargetObjectId to a ulong, which it is (it uses the full size GameObjectId).

Not sure what Dalamud's approach will be going forward, but I'll write a brief explanation of the object id thing here just to have it written.

EntityId - the 4 byte id stored in the GameObject (name comes from a LUA function called GetEntityId)
GameObjectId - 8 byte id that for some types but not all will have the EntityId in the lower 4 bytes and a type in the upper 4 bytes; you can see the rules [here](https://github.com/aers/FFXIVClientStructs/blob/6e02f500cb0140b3a72f56b7115778da75988ff1/FFXIVClientStructs/FFXIV/Client/Game/Object/GameObject.cs#L104)

Some native functions use the EntityId, some use the GameObjectId, and at least on the CS side the motivation to changing the 4 byte name to "EntityId" is to try and stop plugin authors from making the mistake of using the EntityId where the game wants a GameObjectId and running into edge cases where it matters.

PR #1830, opened 2024-06-08
Add basic nameplate service

Pilzinsel64:
grafik

Keep in mind that this is just a basic (but futureorientate) implementation, re-implemented inspired by Pilz.Dalamud. The current implementation uses the `SetPlayerNameplateDetourDelegate` event and provides just a basic way to edit the existing elements of a player nameplate.

It's not possible yet to directly get/edit/add/remove nodes. This can be done in a future PR by a future implementation of the Nameplate service wich uses another Event (more low-level wich then access AktResNode etc.). But I had not the time to get involved with this Event. Most plugins uses this one and that's enough for now imao.

I know the other event is more low-level that gives you more options. But I have no expierience with it and time to investigate and it's not really required for this basic PR in my opinion.

Also, the manual allocation has it's reason. `MemoryHelper.WriteString` didn't work. The parameters like `titlePtr` are addresses to raw byte arrays, so `Utf8String` isn't working too.


Dalamud/Game/Gui/Nameplates/EventArgs/NameplateUpdateEventArgs.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// Gets an object that holds some infos about the nameplate.
+    /// </summary>
+    public NameplateInfo NameplateInfo { get; } = nameplateInfo;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether <see cref="NameplateInfo"/> has been changed.
+    /// <br/>Set this to <see cref="T:true"/> if you changes to <see cref="NameplateInfo"/> should take affect.
+    /// </summary>
+    public bool HasChanged { get; set; }
+}
```
MidoriKami, line 27: Per <https://github.com/goatcorp/Dalamud/pull/1808> It'll probably be best to create an interface for this type.

Dalamud/Game/Gui/Nameplates/Model/NameplateInfo.cs
```diff
@@ -0,0 +1,46 @@
...
+    /// Gets or sets a value indicating whether the title is visible.
+    /// <br/>If the title value is true, the title is visible. Otherwise the title text is not visible and ignored.
+    /// </summary>
+    public bool IsTitleVisible { get; set; }
+
+    /// <summary>
+    /// Gets or sets the icon shown on the nameplate. Mostly used for the status icon.
+    /// </summary>
+    public StatusIcons IconID { get; set; }
+}
```
MidoriKami, line 46: Per <https://github.com/goatcorp/Dalamud/pull/1808> It'll probably be best to create an interface for this type.

Dalamud/Game/Gui/Nameplates/Model/NameplateObject.cs
```diff
@@ -0,0 +1,13 @@
...
+/// Represents a nameplate object (mostly like player or NPC).
+/// </summary>
+/// <param name="pointer">The pointer for the nameplate object.</param>
+public class NameplateObject(IntPtr pointer)
+{
+    /// <summary>
+    /// Gets the pointer for the nameplate object.
+    /// </summary>
+    public IntPtr Pointer { get; } = pointer;
+}
```
MidoriKami, line 13: Per <https://github.com/goatcorp/Dalamud/pull/1808> It'll probably be best to create an interface for this type.

Dalamud/Game/Gui/Nameplates/Model/StatusIcons.cs
```diff
@@ -0,0 +1,91 @@
...
+    /// <summary>
+    /// The status for PvP mentors.
+    /// </summary>
+    MentorPvP = 061544,
+
+    /// <summary>
+    /// The status for a player that recently took a break playing FFXIV.
+    /// </summary>
+    Returner = 061547,
+}
```
MidoriKami, line 91: Does this mimic the datasheet for `OnlineStatus`?

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// </summary>
+    [ServiceManager.ServiceConstructor]
+    private NameplateGui(TargetSigScanner sigScanner)
+    {
+        // Services
+        this.gameGui = Service<GameGui>.Get();
+        this.objectTable = Service<ObjectTable>.Get();
+
+        // Pointers
+        this.namePlatePtr = this.gameGui.GetAddonByName("NamePlate", 1);
```
MidoriKami, line 43: This is likely to explode when the player logs out and logs back in.

The Addon is re-created when you login.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,178 @@
...
+            {
+                Title = MemoryHelper.ReadSeStringNullTerminated(titlePtr),
+                Name = MemoryHelper.ReadSeStringNullTerminated(namePtr),
+                FreeCompany = MemoryHelper.ReadSeStringNullTerminated(freeCompanyPtr),
+                Prefix = MemoryHelper.ReadSeStringNullTerminated(prefixPtr),
+                IsTitleAboveName = isTitleAboveName,
+                IsTitleVisible = isTitleVisible,
+                IconID = (StatusIcons)iconId,
+            };
+
```
MidoriKami, line 126: Is it necessary to parse this data every time? This is a lot of SeString parsing that is happening when it might not even be required. Might be better for NameplateInfo to store a address internally and offer this parsing as a get-only property.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
MidoriKami: Possible fix includes using AddonLifcycle to listen for the Setup and Finalize events for NamePlate, and use that to manage the pointer.

MidoriKami:
@Infiziert90 noticed that you are using `IntPtr` everywhere, should use `nint` instead.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,178 @@
...
+                var freeCompanyRaw = nameplateInfo.FreeCompany.Encode();
+                var freeCompanyNewRaw = nameplateInfo.FreeCompany.Encode();
+                if (!freeCompanyRaw.SequenceEqual(freeCompanyNewRaw))
+                    MemoryHelper.WriteSeString(freeCompanyPtr, nameplateInfo.FreeCompany);
+
+                // Get new Prefix string content
+                var prefixRaw = nameplateInfo.Prefix.Encode();
+                var prefixNewRaw = nameplateInfo.Prefix.Encode();
+                if (!prefixRaw.SequenceEqual(prefixNewRaw))
+                    MemoryHelper.WriteSeString(prefixPtr, nameplateInfo.Prefix);
```
Haselnussbomber, line 160: Please do not write to the pointers.
The data comes from StringArrayData and you don't know how much space the game has allocated for these strings.
Instead, you should use `Unsafe.AsPointer` to get pointers of your encoded byte arrays and pass them to the Original function. There is no need to free them, since the data is on the stack.
The only thing you probably have to do is to make sure they have a null terminator at the end, so concat the arrays with `new byte[] { 0 }`.

Dalamud/Game/Gui/Nameplates/EventArgs/NameplateUpdateEventArgs.cs
Pilzinsel64: Done!

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Pilzinsel64: True. Moved that to `NameplateNode.Text`. It will be parsed when the getter is called the first time.

Pilzinsel64:
> @Infiziert90 noticed that you are using `IntPtr` everywhere, should use `nint` instead.

Changed everything to nint now.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Pilzinsel64: Done!

Dalamud/Game/Gui/Nameplates/Model/NameplateInfo.cs
Pilzinsel64: Done!

Dalamud/Game/Gui/Nameplates/EventArgs/NameplateUpdateEventArgs.cs
Pilzinsel64: Well, I removed that event args. Never seen them as interface and even don't need them anymore now.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,177 @@
...
+        
+        if (namePlateIndex < 0 || namePlateIndex >= AddonNamePlate.NumNamePlateObjects)
+            return null;
+
+        // Get the nameplate info array
+        var nameplateInfoArrayPtr = nint.Zero;
+        unsafe
+        {
+            var framework = FFXIVClientStructs.FFXIV.Client.System.Framework.Framework.Instance();
+            nameplateInfoArrayPtr = new nint(&framework->GetUIModule()->GetRaptureAtkModule()->NamePlateInfoArray);
```
Pilzinsel64, line 97: Seems like this doesn't work anymore. Any idea?

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -92,72 +91,58 @@ public void DisposeService()
...
-                // Get new Prefix string content
-                var prefixRaw = nameplateInfo.Prefix.Encode();
-                var prefixNewRaw = nameplateInfo.Prefix.Encode();
-                if (!prefixRaw.SequenceEqual(prefixNewRaw))
-                    MemoryHelper.WriteSeString(prefixPtr, nameplateInfo.Prefix);
+                if (node.HasChanged)
+                {
+                    var textRaw = node.Text.Encode();
+                    node.Pointer = MemoryHelper.Allocate(textRaw.Length);
+                    ptrToFree.Add(node.Pointer);
```
Haselnussbomber, line 155: Btw, the reason I told you to use Utf8String is because it will add a null terminator after the last byte of the string.
The way you did it now doesn’t have that, so it doesn’t know where the string ends.
You can just push Utf8Strings to the list and then call Dispose in the ForEach

Pilzinsel64: Tried that but can't call `Dispose` as it's non-public.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Haselnussbomber: Should still be there. It’s inherited from UIModuleInterface now via the generators. Just build it and maybe restart your IDE.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Pilzinsel64: Oh, do you mean `Utfa8String.Dtor(bool)`? Missed that before.

Haselnussbomber: Might have to cast it to IDisposable, like
`(utf8string as IDisposable).Dispose()`

Haselnussbomber: Dispose calls Dtor(false), so that would work too, yep,

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -170,7 +170,7 @@ private nint HandleSetPlayerNameplateDetour(nint playerNameplateObjectPtr, bool
             iconId);
 
         // Free pointers
-        ptrToFree.ForEach(n => MemoryHelper.Free(n));
+        ptrToFree.ForEach(n => n.Dtor(true));
```
Haselnussbomber, line 173: You wrote Dtor(true), this will backfire. You didn’t let the game allocate the Utf8Strings, so it can’t free them. Call Dtor(false), so that it will free the StringPtr inside if needed (I say that because Utf8String has an internal buffer too) and let C#s garbage collector handle the Utf8Strings.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Pilzinsel64: I _think_ I got it now. Thanks! I even could simplify the method `GetNameplateGameObject` a bit more.

Dalamud/Game/Gui/Nameplates/Model/INameplateNode.cs
```diff
@@ -0,0 +1,30 @@
+using Dalamud.Game.Text.SeStringHandling;
+
+namespace Dalamud.Game.Gui.Nameplates.Model;
+
+/// <summary>
+/// Represents a nameplate text node.
+/// </summary>
+public interface INameplateNode
```
nebel, line 8: I wonder if it's better to just call this NameplateUpdateString or NameplateUpdateArgument or something else which doesn't use "node". When I think "node" in terms of the addons like NamePlate I think of the persistent UI nodes which make up the plates.

Pilzinsel64: I used "Element" before. But then I used "Node" as it the potential to match future expansion of the code.

MidoriKami:
Is there a particular reason or reasons why this is still a draft?

Pilzinsel64:
@MidoriKami I didn't actually tested this via the Tesplugin yet as I had no time this week. :/
Aaand I didn't took a look to the Status Icon Excel Sheet you asked me for to check.

MidoriKami:
Understandable, not trying to rush, it just seemed like it was pretty close to being good to go, we can always make minor adjustments later.

Pilzinsel64:
Alright, ready for review/merge now!
grafik

Keep in mind that this is just a basic (but futureorientate) implementation, re-implemented inspired by Pilz.Dalamud.

It's not possible yet to directly get/edit/add/remove nodes. This can be done by a future implementation of the Nameplate service wich uses another Event (more low-level wich then access AktResNode etc.). But I had not the time to get involved with this Event. Most plugins uses this one and that's enough for now imao.

Also, the manual allocation has it's reason. `MemoryHelper.WriteString` didn't work. The parameters like `titlePtr` are addresses to raw byte arrays, so `Utf8String` isn't working too.

Dalamud/Game/Gui/Nameplates/Model/StatusIcons.cs
Pilzinsel64: To be checked after merge. Had not the time until feature freeze. But can be deprecated in the future easily, if there is another way to provide that.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /// <inheritdoc/>
+    public T? GetNameplateGameObject<T>(INameplateObject nameplateObject) where T : GameObject
+    {
+        return this.GetNameplateGameObject<T>(nameplateObject.Pointer);
+    }
+
+    /// <inheritdoc/>
+    public T? GetNameplateGameObject<T>(nint nameplateObjectPtr) where T : GameObject
```
nebel, line 76: I wonder if it makes sense to cache the result of this in `NameplateObject` or similar. I think finding either the index or the object is going to be necessary for most users as that's how they're likely to either find the nameplate owner. As-is each consumer is likely to repeat this call, so we'll end up looking up the index and scanning the object table multiple times.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,211 @@
...
+        if (nameplateObjectArrayPtr == nint.Zero)
+            return null;
+
+        // Determine the index of the nameplate object within the nameplate object array
+        var namePlateObjectSize = Marshal.SizeOf(typeof(AddonNamePlate.NamePlateObject));
+        var namePlateObjectPtr0 = nameplateObjectArrayPtr + namePlateObjectSize * 0;
+        var namePlateIndex = (nameplateObjectPtr.ToInt64() - namePlateObjectPtr0.ToInt64()) / namePlateObjectSize;
+        
+        if (namePlateIndex < 0 || namePlateIndex >= AddonNamePlate.NumNamePlateObjects)
+            return null;
```
nebel, line 91: I think it might be a good idea to provide index lookup as a separate function independent from GameObject lookup. If a plugin is doing its own modifications to plates it likely needs to know the index so it can look up the UI Node or find certain data in whatever index it uses internally.

Dalamud/Game/Gui/Nameplates/Model/INameplateNode.cs
nebel: I see. I think if actual UI nodes are ever added to this there will likely need to be new field/event/etc. types, so I'm not sure there is really any forwards-compatibility to calling these "Nodes" here. If not "arguments", maybe "elements"? Things like prefix, name and free company are actually concatenated into the final name string and aren't even separate nodes in the final plate.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Pilzinsel64: Yup, makes sence. I removed the generic type and added a property to `INameplateObject`. Also removed `GetNameplateGameObject<T>` from `INameplateGui`.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
Pilzinsel64: Maybe also makes sense, your right. I added a property to `INameplateObject`. Also splitted `NameplateGui.GetNameplateGameObject<T>` into two methods.

kalilistic:
Will this be merged in with apix? No pressure, but we're talking about falling back to the library if not and want to plan accordingly.

goaaats:
That would be nice, but it'd have to go in this weekend. Could you please re-review if you have time? @Haselnussbomber, @nebel and @MidoriKami - thank you!

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,212 @@
...
+    [ServiceManager.ServiceDependency]
+    private readonly GameGui gameGui = Service<GameGui>.Get();
+
+    [ServiceManager.ServiceDependency]
+    private readonly ObjectTable objectTable = Service<ObjectTable>.Get();
+
+    [ServiceManager.ServiceDependency]
+    private readonly AddonLifecycle addonLifecycle = Service<AddonLifecycle>.Get();
+
+    [Signature("E8 ?? ?? ?? ?? E9 ?? ?? ?? ?? 48 8B 5C 24 ?? 45 38 BE", DetourName = nameof(HandleSetPlayerNameplateDetour))]
```
MidoriKami, line 32: This should be in a AddressResolver class, see `AddonLifecycleAddressResolver`, and `AddonLifecycle` for an example.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,212 @@
...
+    private unsafe delegate nint SetPlayerNameplateDetourDelegate(nint playerNameplateObjectPtr, bool isTitleAboveName, bool isTitleVisible, nint titlePtr, nint namePtr, nint freeCompanyPtr, nint prefix, int iconId);
+
+    /// <inheritdoc/>
+    public event INameplatesGui.OnNameplateUpdateDelegate OnNameplateUpdate;
+
+    /// <inheritdoc/>
+    public void DisposeService()
+    {
+        this.setPlayerNameplateDetourHook.Dispose();
+    }
```
MidoriKami, line 60: You need to also Unregister the AddonLifecycle Listeners here.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,212 @@
...
+        return this.objectTable.SearchById(objectId);
+    }
+
+    private void AddonLifecycle_Event(AddonEvent type, AddonArgs args)
+    {
+        if (type == AddonEvent.PostSetup)
+            this.namePlatePtr = args.Addon;
+        else if (type == AddonEvent.PreFinalize)
+            this.namePlatePtr = nint.Zero;
+    }
```
MidoriKami, line 116: I strongly suggest using two different function for Setup and Finalize, it'll make intent more clear and help prevent bugs like what was fixed earlier.

However this is just a style suggestion, feel free to disregard if you prefer.

Dalamud/Game/Gui/Nameplates/NameplateGui.cs
```diff
@@ -0,0 +1,212 @@
...
+        var pointer = Marshal.AllocHGlobal(bytes.Length + 1);
+        Marshal.Copy(bytes, 0, pointer, bytes.Length);
+        Marshal.WriteByte(pointer, bytes.Length, 0);
+        return pointer;
+    }
+
+    private void PluginFree(nint ptr)
+    {
+        Marshal.FreeHGlobal(ptr);
+    }
```
MidoriKami, line 211: I feel like there has to be a better way, but we can look into that later, as this is implementation details that we can address at any time without breaking the API.

Dalamud/Plugin/Services/INameplatesGui.cs
```diff
@@ -0,0 +1,21 @@
+using Dalamud.Game.ClientState.Objects.Types;
+using Dalamud.Game.Gui.Nameplates.Model;
+
+namespace Dalamud.Game.Gui.Nameplates;
+
+/// <summary>
+/// This class handles interacting with native Nameplate update events and management.
+/// </summary>
+public interface INameplatesGui
```
MidoriKami, line 9: I would strongly prefer `INameplateGui` as I and many others are likely going to not notice that it's plural, where as other services are singular.

MidoriKami (changes requested):
Some concerns that I think need to be addressed immediately before it can be merged, and some suggestions that would be nice to have.

I won't be available for follow up reviews until I have completed MSQ.

nebel:
> That would be nice, but it'd have to go in this weekend. Could you please re-review if you have time? @Haselnussbomber, @nebel and @MidoriKami - thank you!

There is currently an issue with this service which could be quite difficult to address, namely that the main function we hook to do all the work (`SetPlayerNamePlate`) has been fully inlined in 7.0. The original function still exists but it is no longer called, and our hook does nothing. Until we can figure out how to fix this, I think we need to hold off on merging this.

goaaats:
Thanks for the assessment. I assume this will be a problem for kal too then - maybe a solution can be worked out together.

Pilzinsel64:
Arg, that was unintended... Maybe they moved it to the paren function now. 

Hm, there is another function we could hook to. But I don't know how it's called and not what the sig is. I know it's more low level and is called for every single namplate when updating the stringdata, so also NPCs, if I'm right. But I have absolutely no experience with it.

Dalamud.CorePlugin/PluginImpl.cs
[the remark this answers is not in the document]
Pilzinsel64: Wasn't the plan. I prevented commiting it long time but seems somewhen I just missed it. Reverting it.

Dalamud/Game/Gui/Nameplates/Model/INameplateInfo.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// </summary>
+    /// <param name="name">The common namplate element name to serch for.<br/>Note: <see cref="NameplateElementName.Unknown"/> is not supported.</param>
+    /// <returns>Returns the found element, or null if nohting has been found.</returns>
+    INameplateElement? GetElement(NameplateElementName name);
+
+    /// <summary>
+    /// Gets all available nameplate elements.
+    /// </summary>
+    /// <returns>Returns an array of all nameplate element.</returns>
+    INameplateElement[] GetAllElement();
```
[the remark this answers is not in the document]
Pilzinsel64: Not really. It's readonly. :D

Pilzinsel64:
Deprecated due #1915 (big thanks to SE :/ )

PR #1831, opened 2024-06-09
AddonLifcycle Add Show/Hide events.

MidoriKami:
Some addons trigger Show or Hide multiple times, this is expected behavior, the same thing would happen to anyone that tries hooking the addon Show or Hide methods.

There is a LOT of added copy-paste code here, there's probably some room to condense the logic, but I decided to go with the safe approach, because I know that what I copied worked, I can trust the new systems to also work.


MidoriKami:
Approach is fundamentally flawed, closing.

PR #1833, opened 2024-06-11
AddonLifecycle Performance Enchancement

MidoriKami:
Per Aers' somewhat obvious question in hindsight... ("Why do you need to hook it before its requested?"), I have tweaked it slightly to instead store the address of the ReceiveEvent function, and now only hooks it when it is requested by someone.


MidoriKami:
Whoops, PR'd to Wrong Branch.

PR #1838, opened 2024-06-14
Verify conditions before killing game

KazWolfe:
- Add new `ClientState#IsClientIdle` method to test if the client is currently in an "idle" state (nothing important is going on).
- Add new `Condition#AsReadOnlySet` method to receive a set of active conditions.
- Add new helper functions `Condition#OnlyAny` and `Condition#OnlyAll` to evaluate certain game conditions.
- Block kills unless the client is currently considered idle.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -1,6 +1,8 @@
+using System.Linq;
 using System.Runtime.InteropServices;
 
 using Dalamud.Data;
+using Dalamud.Game.ClientState.Conditions;
 using Dalamud.Game.ClientState.Objects;
 using Dalamud.Game.ClientState.Objects.SubKinds;
 using Dalamud.Game.Gui;
@@ -123,6 +125,27 @@ internal sealed class ClientState : IInternalDisposableService, IClientState
     /// Gets client state address resolver.
     /// </summary>
     internal ClientStateAddressResolver AddressResolver => this.address;
+    
+    /// <inheritdoc/>
+    public bool IsClientIdle(out ConditionFlag blockingFlag)
+    {
+        blockingFlag = 0;
+        if (this.LocalPlayer is null) return true;
+        
+        var condition = Service<Conditions.Condition>.GetNullable();
+        
+        var blockingConditions = condition.AsReadOnlySet().Except([
+            ConditionFlag.NormalConditions, 
+            ConditionFlag.Jumping, 
+            ConditionFlag.Mounted, 
+            ConditionFlag.UsingParasol]);
+
+        blockingFlag = blockingConditions.FirstOrDefault();
+        return blockingFlag == 0;
+    }
+    
+    /// <inheritdoc/>
+    public bool IsClientIdle() => this.IsClientIdle(out _);
 
     /// <summary>
     /// Dispose of managed and unmanaged resources.
@@ -270,6 +293,12 @@ internal class ClientStatePluginScoped : IInternalDisposableService, IClientStat
 
     /// <inheritdoc/>
     public bool IsGPosing => this.clientStateService.IsGPosing;
+
+    /// <inheritdoc/>
+    public bool IsClientIdle(out ConditionFlag blockingFlag) => this.clientStateService.IsClientIdle(out blockingFlag);
+
+    /// <inheritdoc/>
+    public bool IsClientIdle() => this.clientStateService.IsClientIdle();
 
     /// <inheritdoc/>
     void IInternalDisposableService.DisposeService()
--- a/Dalamud/Game/ClientState/Conditions/Condition.cs
+++ b/Dalamud/Game/ClientState/Conditions/Condition.cs
@@ -1,3 +1,6 @@
+using System.Collections.Generic;
+using System.Linq;
+
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
 using Dalamud.Plugin.Services;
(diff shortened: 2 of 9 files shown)


Caraxi:
I use the kill command when the game gets stuck in an invalid state because of something stupid I did, this is going to make the kill command useless

KazWolfe:
How on earth are you doing things such that condition flags stay stuck.

NotNite:
The command should never be blocked, ever; what about just the UI?

KazWolfe:
I disagree heavily - removing one way of killing the game will not stop users from using other ways of killing the game provided by Dalamud. Especially not if the most frequent method is not blocked.

NotNite:
So the goal is to stop users from being able to kill their game at all, instead of just making it so you can't accidentally do this? That seems kind of silly, especially given that's just gonna drive users to Task Manager.

/xlkill is a tool, like all other developer oriented commands, and I was under the impression this PR was to make it so you don't accidentally click "Kill Game" in the developer menu. I regularly use /xlkill to quickly close my game after I mess up game state, a dev plugin fails to reload, etc - I see no point in restricting it, especially because I've had to use it in scenarios like Distant Seas where my plugin's entire focus is an instanced duty.

KazWolfe:
> That seems kind of silly, especially given that's just gonna drive users to Task Manager.

That's not our problem as it's out of the game. We should not be providing users means *officially supported in Dalamud* to behave badly. Users using Task Manager or similar is not our problem, since it isn't something we provide.


> /xlkill is a tool, like all other developer oriented commands

As is the Object Table, but we hide that in PVP to prevent advantages. The issue is that people are using it for very non-dev-oriented purposes.

> I regularly use /xlkill to quickly close my game after I mess up game state, a dev plugin fails to reload, etc

Again, how often do you actually do this *in content*?

> especially because I've had to use it in scenarios like Distant Seas where my plugin's entire focus is an instanced duty.

My stance here is that plugins that have good reason to kill the game in instanced content can implement their own debug-build-only commands. I'd still likely advise against such testing as it does have the potential to impact other players, especially in content that requires cooperation.

Aireil:
ObjectTable in PvP and /xlkill is quite the unfair comparison. The former allows you to see data that is supposed to be hidden from the player with no simple alternative while the latter just saves the annoyance of having to open the task manager.
This reminds me of a quote ([source](https://discord.com/channels/581875019861328007/685275026156683280/1202645663385391235)):
> DRM makes everything worse for everyone, and won’t stop the alleged “problem”

wolfcomp:
I can see that certain conditions should block debug features such as PVP but out right blocking it fully in say solo play if you managed to brick it in a bad state where you can't say exit an explorer instance could very much be bad. 
Also just going to say this that a single line command to kill the game from terminal isn't so hard to craft.

sersorrel:
it feels like it might help reach a conclusion if the things this is trying to prevent are made explicit, as well as the things it's not trying to prevent.

having a way to kill the game if you get stuck crafting (something I remember seeing mentioned some time ago, I think in relation to a simpletweak) is probably good; having a way to trivially skip msq roulette cutscenes (something I've seen mentioned several times) is probably bad. are there other examples? is there a set of conditions, or other criteria (e.g. locations), that permit one while prohibiting the other?

MidoriKami:
I feel listing the ways the command can be exploited here is not a great idea, especially if this PR gets rejected, would serve as a list of ideas that we explicitly want to discourage.

MidoriKami:
How this for an idea, if there's dev plugins running, allow restricted commands, that'll cut down on the abuse significantly. As anyone doing dev will easily be able to bypass any limitations anyways.

avafloww:
I don't think this is a battle we should be fighting. At the end of the day, we'll never be able to fully prevent bad behaviour by users. Although it's reasonable imo that we make _some_ attempt to do so where it won't inconvenience legitimate users of a feature, this change has a non-zero chance of impacting developers.

Unskippable cutscene skipping has been happening in a much more efficient way, for as long as I remember, from custom repo plugins. Even discounting that, `taskmgr` / `taskkill /f /im ffxiv_dx11.exe` exist and are both trivial.

I don't think we should set precedent to inconvenience legitimate users (in this case, developers) of a core feature, with the aim to cut perceived abuse, when said abuse can very easily be replicated with tools built-in to the OS anyways.

KazWolfe:
> this change has a non-zero chance of impacting developers.

I'm not certain this is really the case - how realistic is it to get stuck in a condition without some way of escaping it? Even if one *does* manage to do that, devs have countless other ways to kill the game if they *somehow* have managed to place themselves in a situation where they're not idle (including those provided by the OS itself).

> I don't think we should set precedent to inconvenience legitimate users [...] when said abuse can very easily be replicated with tools built-in to the OS anyways.

See above for inconveniencing legitimate users, but also see further above for my general thoughts that even though the abusive behaviors in question *can* be easily replicated via other means, they're not provided by our core systems. I see this as aligning what Dalamud does via dev commands with what the game itself intends to allow in this case (especially as the game itself already enforces it own rules when one can or can't actually close it).

avafloww:
> See above for inconveniencing legitimate users, but also see further above for my general thoughts that even though the abusive behaviors in question _can_ be easily replicated via other means, they're not provided by our core systems. I see this as aligning what Dalamud does via dev commands with what the game itself intends to allow in this case (especially as the game itself already enforces it own rules when one can or can't actually close it).
> 
> The intent here is to prevent users from doing bad things _via systems provided by us_ with zero extra work or breaking out of our walled garden, nothing more or less. Stopping the actual problem is impossible (and in this case is fortunately pretty minor), but that doesn't mean our systems need to make it more accessible either. This entire change is, admittedly, a whole lot of nothing - it marginally impacts devs (who aren't really affected by it), and it marginally impacts bad actors (who have other ways of doing this), but I consider that a benefit - it's an easy way to clean up a loose end and no longer have to concern ourselves with a problem case.

The definition of bad behaviour is mostly one of opinion. If you ask Square Enix, Dalamud's mere existence is bad behaviour. If you ask some more vocal parts of the community, we have plugins on the main repository that enable bad behaviour - and I'd consider that to be pretty accessible out-of-the-box.

We do take a strong stance on trying to prevent the more egregious use of our project (i.e. obvious cheat plugins/otherwise unsanctioned plugins that some folks may consider "cheaty"), but I don't think being able to easily kill the game during development, and the possibility of users abusing that to skip cutscenes, is particularly egregious, _especially_ when we don't have any concrete examples of this being a widespread phenomenon. As much as I dislike the saying, this change feels like a solution in need of a problem.

PR #1839, opened 2024-06-15
refactor: use client struct for character name check

kalilistic:


kalilistic:
I'll redo this against APIX.

PR #1842, opened 2024-06-15
feat: New Condition APIs

KazWolfe:
- New methods for `ICondition`:
  - `AnyExcept` ensures the listed conditions are *not* present.
  - `OnlyAny` ensures that *only* the listed conditions are met.
  - `OnlyAll` ensures that the condition state matches the listed set.
- New `IsGameIdle` method in `IClientState` can be used to check if the player is not in any "active" game state.

### DIFF
--- a/Dalamud/Plugin/Services/ICondition.cs
+++ b/Dalamud/Plugin/Services/ICondition.cs
@@ -40,6 +40,12 @@ public interface ICondition
     
     /// <inheritdoc cref="this[int]"/>
     public bool this[ConditionFlag flag] => this[(int)flag];
+    
+    /// <summary>
+    /// Convert the conditions array to a set of all set condition flags.
+    /// </summary>
+    /// <returns>Returns a set.</returns>
+    public IReadOnlySet<ConditionFlag> AsReadOnlySet();
 
     /// <summary>
     /// Check if any condition flags are set.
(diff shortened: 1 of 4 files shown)


Dalamud/Plugin/Services/ICondition.cs
```diff
@@ -51,4 +59,26 @@ public interface ICondition
...
+    /// <returns>Returns a bool.</returns>
+    public bool OnlyAny(params ConditionFlag[] other);
+
+    /// <summary>
+    /// Check that *only* the specified flags are set. Unlike <see cref="OnlyAny"/>, this method requires that all the
+    /// specified flags are set and no others are present.
+    /// </summary>
+    /// <param name="other">The array of flags to check.</param>
+    /// <returns>Returns a bool.</returns>
+    public bool OnlyAll(params ConditionFlag[] other);
```
avafloww, line 83: Maybe a name like `EqualTo` or similar would be better, since this seems to confer "the current (set of) conditions are equal to the provided conditions".

Dalamud/Plugin/Services/ICondition.cs
```diff
@@ -51,4 +59,26 @@ public interface ICondition
...
+    /// <param name="except">The array of flags to check.</param>
+    /// <returns>Returns false if any of the listed conditions are present, true otherwise.</returns>
+    public bool AnyExcept(params ConditionFlag[] except);
+
+    /// <summary>
+    /// Check that *only* any of the condition flags specified are set.
+    /// </summary>
+    /// <param name="other">The array of flags to check.</param>
+    /// <returns>Returns a bool.</returns>
+    public bool OnlyAny(params ConditionFlag[] other);
```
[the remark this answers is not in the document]
KazWolfe: `OnlyAny`, as the docs state, are designed to check "are the only conditions currently present within this limited list." That is to say, nothing but the listed conditions are permitted.

This was named this way over `Only` (as in Goat's initial PR for auto-updates) as it's more clear  for "true if the only conditions currently active are in this list"

Dalamud/Plugin/Services/ICondition.cs
KazWolfe: `OnlyAll` was chosen to complement `OnlyAny` and make the distinction more clear, but we can change it to EqualTo or similar. That said, I will apply this change.

avafloww (approved):

PR #1843, opened 2024-06-15
Implement hook stacking and priorities

marzent:
Hook stacking without having multiple managed to native and native to managed transitions was the primary objective here, but priorities were relatively easy to implement on top of that.

Currently this implements the following priority bands:

* 0: After-notify only hooks (always run last)
* 1-127: Normal priority hooks
* 128-191: Dalamud internal hooks
* 192-254: High priority hooks
* 255: Before-notify only hooks (always run first)

Normal priority hooks convert to Dalamud internal ones automatically based on the calling assembly. For fine grained control inside a priority band, the optional `precedence` parameter can be used to lower or raise the effective priority.

This also makes some locking a bit stricter (like making sure jumps are followed only under the global hook lock).

There is some IL dynamically generated in 3 places (most of the time only in 1), I don't believe there is a reasonably performant alternative to that unfortunately.
When stacking hooks of incompatible signatures (like ST `private delegate void* GlobalEventDelegate(AtkUnitBase* atkUnitBase, AtkEventType eventType, uint eventParam, AtkResNode** eventData, uint* a5)` and Dalamuds `private delegate IntPtr AtkUnitBaseReceiveGlobalEvent(AtkUnitBase* thisPtr, ushort cmd, uint a3, IntPtr a4, uint* a5)`) a warning is logged and IL thunks are generated inside a `CastingHook<,>`.

`AsmHook` seem to exist a bit outside the other hooks, so these currently do not stack, but are now also under the global hook lock.

This also fixes the `callingAssembly` not resolving to the actual plugin and marking all hooks as "Dalamud Hooks".

### DIFF
--- a/Dalamud/Hooking/AsmHook.cs
+++ b/Dalamud/Hooking/AsmHook.cs
@@ -32,19 +32,22 @@ public sealed class AsmHook : IDisposable, IDalamudHook
     /// <param name="asmHookBehaviour">How the hook is inserted into the execution flow.</param>
     public AsmHook(IntPtr address, byte[] assembly, string name, AsmHookBehaviour asmHookBehaviour = AsmHookBehaviour.ExecuteFirst)
     {
-        address = HookManager.FollowJmp(address);
+        lock (HookManager.HookSyncRoot)
+        {
+            address = HookManager.FollowJmp(address);
 
-        // We cannot call TrimAfterHook here because the hook is activated by the caller.
-        HookManager.RegisterUnhooker(address);
+            // We cannot call TrimAfterHook here because the hook is activated by the caller.
+            HookManager.RegisterUnhooker(address);
 
-        this.address = address;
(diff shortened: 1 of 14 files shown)


marzent:
The current reloaded also uses a thread-local assembler iirc... 
But switching that out would be probably best done in its own PR I think, given that this one is already a bit unwieldy.

KazWolfe:
Another note: if Dalamud is starting to own hooks, can we have some way of tracking hook timings? That's right now our biggest blind spot in terms of figuring out what plugins are or aren't doing timing-wise.

marzent:
> Another note: if Dalamud is starting to own hooks, can we have some way of tracking hook timings? That's right now our biggest blind spot in terms of figuring out what plugins are or aren't doing timing-wise.

Yeah this can be done, wanted to clean up the priority calculation a bit more and will try to add that as well.

KazWolfe:
Closing this for the same reason as #2558.

We'd need a DIP to move forward on how exactly to handle hook order and management. (Also, this PR is very outdated now.)

PR #1850, opened 2024-06-21
Add /xlcopylog command

anna-is-cute:
Add a command to copy the dalamud.log file to the user's clipboard, behaving the same as copying the file in an Explorer window. Sorry about the messy code.

I added this to Heliosphere recently, and it has been useful during troubleshooting. Just thought I'd toss this your way in case you might also find it useful. If not, no problem.

You may want to change it to allow copying various log files instead of *just* `dalamud.log`.

Instead of telling users to go to `%appdata%\XIVLauncher` and find the text file called `dalamud`, you can instruct them to do `/xlcopylog` if they're able to open their game.

### DIFF
--- a/Dalamud/Interface/Internal/DalamudCommands.cs
+++ b/Dalamud/Interface/Internal/DalamudCommands.cs
@@ -466,6 +466,7 @@ internal class DalamudCommands : IServiceType
             }
             else
             {
+                hGlobal.Dispose();
                 chatGui.Print(string.Format(Loc.Localize("DalamudLogCopyFailure", "Could not copy log file to clipboard."), "default"));
             }
         }
--- a/Dalamud/Interface/Internal/DalamudCommands.cs
+++ b/Dalamud/Interface/Internal/DalamudCommands.cs
@@ -2,7 +2,6 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
-using System.Text;
 
 using CheapLoc;
 using Dalamud.Configuration.Internal;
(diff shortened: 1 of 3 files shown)


anna-is-cute:
Oh, I didn't intend to update clientstructs.

Dalamud/Interface/Internal/DalamudCommands.cs
```diff
@@ -406,4 +421,53 @@ private void OnOpenProfilerCommand(string command, string arguments)
...
+                    hGlobal);
+                Win32_PInvoke.CloseClipboard();
+
+                chatGui.Print(string.Format(Loc.Localize("DalamudLogCopySuccess", "Log file copied to clipboard."), "default"));
+            }
+            else
+            {
+                chatGui.Print(string.Format(Loc.Localize("DalamudLogCopyFailure", "Could not copy log file to clipboard."), "default"));
+            }
+        }
```
KazWolfe: Please consider moving this into `Util.cs` or similar as an `internal` method returning a `bool` if successful. This way, we can reuse it elsewhere, including possibly as a button in `/xllog`.

Dalamud/Interface/Internal/DalamudCommands.cs
```diff
@@ -11,6 +13,12 @@
 using Dalamud.Plugin.Internal;
 using Dalamud.Utility;
 using Serilog;
+using Windows.Win32.Foundation;
+using Windows.Win32.System.Memory;
+using Windows.Win32.System.Ole;
+using Windows.Win32.UI.Shell;
+
+using Win32_PInvoke = Windows.Win32.PInvoke;
```
KazWolfe, line 21: T/N (non-blocking): does this cause any problems in Linux/Wine?

KazWolfe (changes requested):
Hi! Thanks for the PR. See attached comments - ideally we'd genericize this function so that we can use it elsewhere.

See, for example, WPF's [`SetFileDropList (fileDropList)`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.clipboard.setfiledroplist?view=windowsdesktop-8.0#system-windows-clipboard-setfiledroplist(system-collections-specialized-stringcollection)) for a possible API surface to replicate.

goaaats:
It would be really nice to do this for a tspack. We should merge it like this now(just the log) and I can have a peek  at expanding it.

anna-is-cute:
I believe I've addressed the requests/comments. I also fixed a few small bugs I noticed while refactoring. I've tested using one and two files.

KazWolfe (approved):

PR #1854, opened 2024-06-21
AddonLifecycle AddressResolver Fix

MidoriKami:


KazWolfe:
Proposed alternate fix in 34abd19f78fb96ad31cedec9dd88159501520e73. If this is not sufficient, please re-submit this PR against the new HEAD for apiX.

Note that apiX *was* force-pushed, so please ensure you've force pulled and reset back to our version.

PR #1857, opened 2024-06-25
Use IconFontFixedWidth

MidoriKami:


goaaats:
Why are you not removing the offset from the API?

MidoriKami:
I think it's a good idea to remove the offset property, but was considering the off chance someone wanted to do something unusual with it there would be no harm keeping it.

I'll gladly remove it if your prefer.

PR #1860, opened 2024-06-28
Fix AddonLifecycle OnSetup hook

Haselnussbomber:
This PR fixes the OnSetup hook for AddonLifecycle.

Instead of using the CallHook, which writes the detour address to rax and then calls it, I copied the class and adapted it to our needs: the AddonSetupHook simply writes the detour address to r9, which is called just after 0x1406552A0 in both cases. 😎

Self-test result:
Self-Test

### DIFF
--- a/Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
+++ b/Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
@@ -1,0 +1,80 @@
+using System.Runtime.InteropServices;
+
+using Reloaded.Hooks.Definitions;
+
+namespace Dalamud.Hooking.Internal;
+
+/// <summary>
+/// This class represents a callsite hook used to replace the address of the OnSetup function in r9.
+/// </summary>
+/// <typeparam name="T">Delegate signature for this hook.</typeparam>
+internal class AddonSetupHook<T> : IDisposable where T : Delegate
+{
+    private readonly Reloaded.Hooks.AsmHook asmHook;
+    
+    private T? detour;
+    private bool activated;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="AddonSetupHook{T}"/> class.
+    /// </summary>
+    /// <param name="address">Address of the instruction to replace.</param>
+    /// <param name="detour">Delegate to invoke.</param>
+    internal AddonSetupHook(nint address, T detour)
+    {
+        this.detour = detour;
+
+        var detourPtr = Marshal.GetFunctionPointerForDelegate(this.detour);
+        var code = new[]
+        {
+            "use64",
+            $"mov r9, 0x{detourPtr:X8}",
+        };
+        
+        var opt = new AsmHookOptions
+        {
+            PreferRelativeJump = true,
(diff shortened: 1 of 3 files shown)


KazWolfe (approved):

KazWolfe:
Requesting review from @MidoriKami too.

Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    /// <summary>
+    /// Remove a hook from the current process.
+    /// </summary>
+    public void Dispose()
+    {
+        this.asmHook.Disable();
+        this.detour = null;
+    }
+}
```
MidoriKami, line 80: I was thinking of modifying CallHook and calling it something like CustomInstructionHook, where you pass in the instructions you want to write to the callsite via constructor arg or property setter, with a static field for what CallHook does currently being the default, this way we aren't duplicating a class in its entirety.

This is fine for now, as this is all internal I can clean it up later (after MSQ)

Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            "use64",
+            $"mov r9, 0x{detourPtr:X8}",
+        };
+        
+        var opt = new AsmHookOptions
+        {
+            PreferRelativeJump = true,
+            Behaviour = Reloaded.Hooks.Definitions.Enums.AsmHookBehaviour.DoNotExecuteOriginal,
+            MaxOpcodeSize = 5,
```
MidoriKami, line 38: I believe for this specific case we might want to set MaxOpcodeSize to 7, as it is a 7-byte instruction we are overwriting.

However, as I am unfamiliar with how Reloaded hook does its magic internally this might not be necessary.

MidoriKami (commented):
Thanks to taking a look at this so quickly, just some very minor complaints, no action is requested, just want these notes documented here.

Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
Haselnussbomber: Funny enough, I was thinking this myself. But then I was like: why not use Reloaded.Hooks.AsmHook directly instead of making a wrapper around it? But whatever.^^ 🤷‍♂️

Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
Haselnussbomber: This is not a problem, as it will fill the remaining bytes of the instruction with NOPs. :)

Screenshot

MidoriKami: Nice! Thank you for checking!

Dalamud/Game/Addon/Lifecycle/AddonSetupHook.cs
MidoriKami: There will likely eventually be some kind of larger hook management system in Dalamud to potentially better help with cleanup, metrics, and other things. So for now I do plan to make a slightly friendlier wrapper.

PR #1864, opened 2024-06-29
Allow InterfaceManager fonts to be null

Critical-Impact:
So this one doesn't directly fix an issue that any player would run into, however it's something I'm running into and my plan is to release a nuget package allowing for developers to boot plugins with minimal effort from within their IDE.

In respect to that, this minor change allows for calls to InterfaceManager.IconFont,etc to not explode if InterfaceManager is not available(it'll always be available when Dalamud is running otherwise something has gone massively wrong), and instead returning imgui's default font

No issues if this doesn't get merged, it really only helps me but it shouldn't hurt anyone either

### DIFF
--- a/Dalamud/Interface/Internal/InterfaceManager.cs
+++ b/Dalamud/Interface/Internal/InterfaceManager.cs
@@ -128,28 +128,28 @@ internal class InterfaceManager : IInternalDisposableService
     /// <strong>Accessing this static property outside of the main thread is dangerous and not supported.</strong>
     /// </summary>
     public static ImFontPtr DefaultFont =>
-        WhenFontsReady().DefaultFontHandle!.LockUntilPostFrame().OrElse(ImGui.GetIO().FontDefault);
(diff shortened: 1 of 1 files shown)


goaaats:
This seems dangerous to me. In "non-testing" environments where IM manages the ImGui context, it should still throw this exception when IM is null, since GetIO() and FontDefault willl definitely not be valid to call.

Critical-Impact:
I'm wanting to avoid relying on behaviour on Dalamud's part as much as possible, though you might be right that there needs to be something in Dalamud to make this easier. 

I've had a few ideas on how this could be accomplished assuming a custom build of dalamud was used:
1) new csproj called Dalamud.Moq/Mock , provides access to the Service container and allows for moq versions to be injected. As it'd be built with the custom version of dalamud, it'd bypass the internal barrier
2) As you suggested, some sort of testing configuration that only allows the null return when in that specific build mode

I'll close this for now but 1 is probably the one I like the most and I could add the csproj in some sort of automatic build process but it could be useful for unit testing parts of dalamud

PR #1865, opened 2024-06-29
Add font properties to IUiBuilder

Critical-Impact:
This provides properties that match the static UiBuilder counterparts. Ideally the statics should go but I've conceded that point, I understand the convience factor, though it does make for a leaky interface

For those using DI, unit tests/mocks/etc this should allow for parity

### DIFF
--- a/Dalamud/Interface/UiBuilder.cs
+++ b/Dalamud/Interface/UiBuilder.cs
@@ -118,6 +118,34 @@ public interface IUiBuilder
     IFontSpec DefaultFontSpec { get; }
 
     /// <summary>
+    /// Gets the default Dalamud font size in points.
+    /// </summary>
+    public float FontDefaultSizePt { get; }
+
+    /// <summary>
+    /// Gets the default Dalamud font size in pixels.
+    /// </summary>
+    public float FontDefaultSizePx { get; }
+
+    /// <summary>
+    /// Gets the default Dalamud font - supporting all game languages and icons.<br />
+    /// <strong>Accessing this static property outside of <see cref="Draw"/> is dangerous and not supported.</strong>
+    /// </summary>
+    public ImFontPtr FontDefault { get; }
+
+    /// <summary>
+    /// Gets the default Dalamud icon font based on FontAwesome 5 Free solid.<br />
+    /// <strong>Accessing this static property outside of <see cref="Draw"/> is dangerous and not supported.</strong>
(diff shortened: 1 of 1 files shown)


Dalamud/Interface/UiBuilder.cs
```diff
@@ -117,6 +117,34 @@ public interface IUiBuilder
...
+    /// Gets the default Dalamud icon font based on FontAwesome 5 Free solid.<br />
+    /// <strong>Accessing this static property outside of <see cref="Draw"/> is dangerous and not supported.</strong>
+    /// </summary>
+    public ImFontPtr FontIcon { get; }
+
+    /// <summary>
+    /// Gets the default Dalamud monospaced font based on Inconsolata Regular.<br />
+    /// <strong>Accessing this static property outside of <see cref="Draw"/> is dangerous and not supported.</strong>
+    /// </summary>
+    public ImFontPtr FontMono { get; }
```
KazWolfe, line 146: If this is dangerous, is there a way we can assert that we're inside a draw context?

Critical-Impact: I believe we could put something into this function
 
https://github.com/goatcorp/Dalamud/blob/97b01d697b9e409e19a2a2a11d4a7b251629cecc/Dalamud/Interface/Internal/InterfaceManager.cs#L1154
Like what we did with ThreadSafety we could mark as being inside RawDX11Scene.OnBuildUI and assert if they try to access the properties

Asserts could be added instantly to the new properties within IUiBuilder but the statics would need to throw a log message and not crash to maintain backwards compatibility IMO

@KazWolfe thoughts?

KazWolfe: If you can sneak this in now, it'd be great. Otherwise I'll just merge this as-is and we can api-next a proper fix.

Critical-Impact: https://github.com/goatcorp/Dalamud/blob/9956424c132e06a642d40ebb3c0cf663960258bf/Dalamud/Interface/ManagedFontAtlas/Internals/FontHandle.cs#L106 already warns them if they try to access fonts outside the main thread which I've been assured is the same thread the draw happens on so that should cover us.

Critical-Impact: We may also want to consider making that more than a warning at some point

PR #1868, opened 2024-06-29
Update MarketBoard related packet structs

Infiziert90:
- Fixed all packet structs for 7.0
- Addd new fields from the packets
- Made interfaces public
- Updates Data MarketBoard Window

**Note**
- Marks values as obsoletes if they are required by Universalis, but these contain fake values


Infiziert90:
@karashiiro best if you give it a quick look that nothing blows up for the upload

karashiiro (approved):

goaaats (approved):

PR #1871, opened 2024-06-30
Add clientstruct widget

Infiziert90:
A simple widget for xldata that shows all Proxy addresses currently known (addresses are Selectables)
More future updates could also show Agent addresses
image

### DIFF
--- a/Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/Data/DataWindow.cs
@@ -27,6 +27,7 @@ internal class DataWindow : Window, IDisposable
         new AetherytesWidget(),
         new AtkArrayDataBrowserWidget(),
         new BuddyListWidget(),
+        new ClientStructWidget(),
         new CommandWidget(),
         new ConditionWidget(),
         new ConfigurationWidget(),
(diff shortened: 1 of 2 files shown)


KazWolfe (approved):
As of now I don't really know if I see the value that this specific widget adds. This data can be captured in plenty of other places (e.g. ReClass), so the value of a specific tab for most of these addresses seems odd.

I feel like a lot of these addresses can instead be placed on specific debug tabs that more closely relate to them, e.g. how addon addresses are already present in the Addon Inspector.

Either things should be put more into context or this debugger tab needs to be more clear about what its value is and why we'd want to have it around IMO.

MidoriKami:
Might be better to call the tab InfoProxies, instead of an extremely generic "ClientStructs" term.
Might also be a good idea to stub out the info proxy info a little, or have click-to-copy-to-clipboard.

As this is right now it doesn't really seem to be particularly useful.

Infiziert90:
The plan for this widget was more of a place of all related addresses, I started with info proxy as they aren't in any of the other debug tools like addon > agent

And the use case for this is a simply tool to copy addresses from without needing to log them in a plugin or some other tool, instead you can directly go here and just copy the address you looking for

NotNite:
Minor nitpick: It's "ClientStructs", plural, not Struct.

MidoriKami:
> The plan for this widget was more of a place of all related addresses, I started with info proxy as they aren't in any of the other debug tools like addon > agent
> 
> And the use case for this is a simply tool to copy addresses from without needing to log them in a plugin or some other tool, instead you can directly go here and just copy the address you looking for And yes, exactly Tools like reclass would benefit from this as you can easily get addresses and paste them in there (similar to all of simpletweaks debug tools)

Then I would suggest maybe doing some fun reflection shennagains and displaying anything that has a `.Instance` method, and maybe even grouping them by type, ie Agent, Controller, InfoModule.

PR #1879, opened 2024-07-02
Switch INotificationManager and ITitleScreenMenu to use ISharedImmediateTexture

Critical-Impact:
**INotificationManager**

- Simplifies the API, can only pass in a `ISharedImmediateTexture`, if a user wants to pass in a already created texture they can use `ForwardingSharedImmediateTexture` or create their own implementation of `ISharedImmediateTexture`

- There is no longer an option to leave textures open, they can RentAsync a `ISharedImmediateTexture` or create a texture and dispose of it as they see fit

**ITitleScreenMenu**
- Users can now only pass in a `ISharedImmediateTexture` as per INotificationManager
- The icon will only be validated on draw(that it's 64x64), if it is invalid, the entry will be removed from the list and a error will be logged.

As mentioned above this PR also adds in a wrapper class for IDalamudTextureWrap if the user creates the texture wrap themselves called `ForwardingSharedImmediateTexture`

### DIFF
--- a/Dalamud/Interface/ImGuiNotification/Internal/ActiveNotification.cs
+++ b/Dalamud/Interface/ImGuiNotification/Internal/ActiveNotification.cs
@@ -13,6 +13,8 @@ using Serilog;
 
 namespace Dalamud.Interface.ImGuiNotification.Internal;
 
+using Textures;
+
 /// <summary>Represents an active notification.</summary>
 internal sealed partial class ActiveNotification : IActiveNotification
 {
@@ -22,9 +24,6 @@ internal sealed partial class ActiveNotification : IActiveNotification
     private readonly Easing hideEasing;
     private readonly Easing progressEasing;
     private readonly Easing expandoEasing;
-
-    /// <summary>Whether to call <see cref="IDisposable.Dispose"/> on <see cref="DisposeInternal"/>.</summary>
-    private bool hasIconTextureOwnership;
 
     /// <summary>Gets the time of starting to count the timer for the expiration.</summary>
     private DateTime lastInterestTime;
@@ -119,31 +118,10 @@ internal sealed partial class ActiveNotification : IActiveNotification
     }
 
     /// <inheritdoc/>
-    public IDalamudTextureWrap? IconTexture
+    public ISharedImmediateTexture? IconTexture
     {
         get => this.underlyingNotification.IconTexture;
-        set => this.IconTextureTask = value is null ? null : Task.FromResult(value);
-    }
-
-    /// <inheritdoc/>
-    public Task<IDalamudTextureWrap?>? IconTextureTask
-    {
-        get => this.underlyingNotification.IconTextureTask;
-        set
-        {
-            // Do nothing if the value did not change.
-            if (this.underlyingNotification.IconTextureTask == value)
-                return;
-
-            if (this.hasIconTextureOwnership)
-            {
-                _ = this.underlyingNotification.IconTextureTask?.ToContentDisposedTask(true);
-                this.underlyingNotification.IconTextureTask = null;
-                this.hasIconTextureOwnership = false;
-            }
-
-            this.underlyingNotification.IconTextureTask = value;
-        }
+        set => this.underlyingNotification.IconTexture = value;
     }
 
     /// <inheritdoc/>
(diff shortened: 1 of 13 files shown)


Dalamud/Interface/ImGuiNotification/Internal/ActiveNotification.cs
```diff
@@ -317,11 +268,6 @@ internal void RemoveNonDalamudInvocations()
         if (this.Icon is { } previousIcon && !IsOwnedByDalamud(previousIcon.GetType()))
```
Soreepeong, line 268: Change this line so that `IsOwnedByDalamud` check is also done for the IDTW stored inside `ForwardingSharedImmediateTexture`.

goaaats:
We only have two options here now, sadly:
1. Add wrapper APIs that automatically convert texture wraps to ForwardingSharedImmediateTexture and tag them as Api11Obsolete
2. Wait until 7.1 or whenever we can break it again.

Sorry, this is probably on me. We should have kicked this off a lot earlier.

Critical-Impact:
I've only given this a cursory test as I'll be busy for most of tonight but I've done what I can in terms of adding back the missing APIs and then having the redirect to the right place.

The primary caveat with this is that it awaits the result of the old Task based methods. This could mean potential hitching as it loads but without creating a level of indirection that I don't really want to do it's about the best I could come up with.

The ITextureProvider changes means that if they are actually providing a Task<IDalamudTextureWrap> I'd be very inclined to say it's already a loaded wrap so I'd say the above point is moot but opinion may differ

Dalamud/Interface/ImGuiNotification/Notification.cs
```diff
@@ -32,7 +32,33 @@ public sealed record Notification : INotification
...
+
+        set
+        {
+            if (value == null)
+            {
+                this.ImmediateIconTexture = null;
+            }
+            else
+            {
+                var dalamudTextureWrap = value.Result;
```
Soreepeong, line 57: Assigning failed tasks should not throw. IMO it's better to just leave previous properties/methods alone and add ISIT on its own, and only on drawing refer to ISIT at first and then check the rest to figure out the correct thing to draw.

Critical-Impact: I'd basically need to throw out everything I've already done if that's the route you want to go, as it'd be faster for me to start on a new branch and implement a ImmediateIconTexture there rather than trying to put back in what I've already removed

Have added in a try/catch, not pretty but does aleviate that above concern

Soreepeong: Hopefully nobody refers to the state of the task after assigning it to the notification, too. Not sure why would one do that, so I'd guess it's be safe for this property to be "inconsistent".

Critical-Impact: Only thing I could think of is creating a new Task.FromException<IDalamudTextureWrap?>(exception), storing it in a field and returning that when IconTextureTask.get is called

Then when a set resolves correctly, it sets the failed task field to null.

Not sure why would want to do it either, if I don't do something like above, it shouldn't cause a crash as they'll still have to null check, just a bit incosnsistent as you said

Soreepeong: Another possible way would be making ForwardingSIT store Task<IDTW>  instead?

Critical-Impact: I've considered it however it'd require the GetWrapOrEmpty function to call `Service<DalamudAssetManager>.Get().Empty4X4` which adds a dependency on an internal service.

Now if we made it so you could only get a ForwardingSharedImmediateTexture's from the ITextureProvider and it injected the 4x4 into the ForwardingSharedImmediateTexture that'd be fine but it adds a hard dependency to an object that could concievably be constructed outside of Dalamud

goaaats:
@Critical-Impact Sorry, could you maybe rebase this and change the API to remove the obsolete/legacy functionality for API12? Thank you!

Critical-Impact:
@goaaats I've reworked it as requested.
This still introduces ForwardingSharedImmediateTexture as previously discussed. 
Tested it and everything I tried worked as I'd expect it to.

goaaats (approved):

PR #1882, opened 2024-07-02
Add new default 'search score' sorting method for plugin search

nebel:
Adds a new sorting option "Search score" to the plugin installer window. The window now starts in "adaptive" mode when opened, and in this mode it will switch between "Search score" and "Alphabetical" based on whether or not there's text in the search box.

Selecting an entry from the dropdown manually will disable adaptive mode until the window is closed. It's also possible to select "Search score" manually as long as there's text in the search box. If "Search score" is selected manually but no text is entered, it will sort alphabetically.

I'm not sure how to handle the new localization needed for the "Search score" string or if we want to name it something more user friendly (e.g. "Best match"?). Please let me know!


goaaats (approved):

goaaats:
Looks great, thanks!

PR #1895, opened 2024-07-04
fix: Make autoupdates work again

KazWolfe:
- Always opportunistically schedule a future (fast) auto-update after a check/attempt has been made.
- Track the update notification dismiss to extend this time out.

This change relies on the fact that a second auto-update won't run so long as a notification is open.

### DIFF
--- a/Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
+++ b/Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
@@ -190,14 +190,12 @@ internal class AutoUpdateManager : IServiceType
         {
             this.hasStartedInitialUpdateThisSession = true;
             
-            // Schedule the next update check immediately. No matter what the result of this initial update is, we'll
-            // have the next one queued up (or provide the user the ability to defer).
-            this.nextUpdateCheckTime = DateTime.Now + TimeBetweenUpdateChecks;
-            
             var currentlyUpdatablePlugins = this.GetAvailablePluginUpdates(DecideUpdateListingRestriction(behavior));
             if (currentlyUpdatablePlugins.Count == 0)
             {
                 this.IsAutoUpdateComplete = true;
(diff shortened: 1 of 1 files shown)


Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
```diff
@@ -245,26 +249,7 @@ private IActiveNotification GetBaseNotification(Notification notification)
         if (this.updateNotification != null)
             throw new InvalidOperationException("Already showing a notification");
         
-        if (this.notificationHasStartedUpdate)
-            throw new InvalidOperationException("Lost track of notification state"); 
-        
         this.updateNotification = this.notificationManager.AddNotification(notification);
-        this.updateNotification.Dismiss += _ =>
-        {
-            this.updateNotification = null;
```
goaaats, line 254: We still need this. We don't do anything until this is null again.

Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
```diff
@@ -236,6 +236,10 @@ private void OnUpdate(IFramework framework)
                         this.NotifyUpdatesAreAvailable(
                             this.GetAvailablePluginUpdates(
                                 DecideUpdateListingRestriction(behavior)));
+                        
+                        // Schedule the next update check opportunistically.
+                        this.nextUpdateCheckTime = DateTime.Now + TimeBetweenUpdateChecks;
```
goaaats, line 241: You should still schedule this inside the callback. That was supposed to prevent the case of someone starting the game with the notification open, going away and coming back, then clicking it off and immediately having another update - someone actually complained about that already, which is why I changed it. We should only calculate the next time from the point a notification is dismissed.

goaaats (changes requested):

goaaats (approved):

PR #1897, opened 2024-07-05
fix: Make MarketBoardCurrentOfferings.ItemListings accessible

lichie567:
Not very familiar with this stuff but I am pretty sure this was supposed to be public.


Critical-Impact:
Yes this should be

Critical-Impact:
Yes, this appears to be an oversight on my part, should be public

goaaats (approved):

PR #1905, opened 2024-07-07
Introduce IAsyncDalamudPlugin to tidy up plugin lifecycle

goaaats:
* Introduces IAsyncDalamudPlugin to tidy up plugin lifecycle
* Makes sure that calling UnloadAsync() or LoadAsync() don't block the main thread unnecessarily

### DIFF
--- a/Dalamud/Plugin/IDalamudPlugin.cs
+++ b/Dalamud/Plugin/IDalamudPlugin.cs
@@ -1,8 +1,21 @@
+using System.Threading.Tasks;
+
 namespace Dalamud.Plugin;
 
 /// <summary>
-/// This interface represents a basic Dalamud plugin. All plugins have to implement this interface.
+/// This interface represents a basic Dalamud plugin.
 /// </summary>
+[Obsolete("Use IAsyncDalamudPlugin instead and make sure that your plugin can load and unload asynchronously. This interface will be removed in a future version. Please refer to http://ooo for more information.")]
 public interface IDalamudPlugin : IDisposable
 {
+}
+
+/// <summary>
+/// This interface represents a basic Dalamud plugin that can be loaded and unloaded asynchronously.
+/// </summary>
+public interface IAsyncDalamudPlugin : IAsyncDisposable
+{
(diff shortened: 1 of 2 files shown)


KazWolfe:
What's the intent for plugin lifecycle, then? I'd assume it's `ctor`, `LoadAsync`, <execution>, `UnloadAsync`, `Dispose`? In this case, this would also mean that the plugin's `ctor` would also be called in a random thread, yes?

If so, I'm not certain we need to create a new method to do this. Would it be sensible to just place these new methods into `IDalamudPlugin` and assign a default implementation to them until we're ready to force people to do more async work?

goaaats:
`LoadAsync` is really just for convenience. People have asked for it and there's no reason not to have it. The main reason for the new interface is implementing `IAsyncDisposable` instead and forcing people to rethink their disposal scheme.

KazWolfe:
> LoadAsync is really just for convenience. People have asked for it and there's no reason not to have it. 

Would still be good to document when and why it should be used I think. I agree it's useful to have, but having two entrypoints can be a bit confusing alone.

> The main reason for the new interface is implementing IAsyncDisposable instead and forcing people to rethink their disposal scheme.

Point taken. I'd ask we plan to refactor `IAsyncDalamudPlugin` back to just `IDalamudPlugin` in the future then, since the `Async` descriptor would become redundant if we're pushing everyone into async territory.

Critical-Impact:
Is there a risk that a plugin fails to complete it's DisposeAsync in a timely manner(or never completes) and if that happens should we have a timeout? (Apologies if there is one, I had a look through and I couldn't see one)

KazWolfe:
Think this is good for api12 targeting, though it needs to be rebased to the `api12` branch.

What'll our strategy be for migration? I'd still like to see us collapse down to just async, but we'd need to decide naming. Perhaps just soft-obsolete `IDalamudPlugin` and keep the `IAsyncDalamudPlugin` name forever? Or do the aforementioned rename come api13?

goaaats:
Superseded by #2678

PR #1913, opened 2024-07-10
GetFromFile without FileInfo

RyouBakura:
The new API to load textures from disk creates a new FileInfo object for every call every frame. This wastes disk I/O and creates unnecessary GC.

```csharp
        /// <inheritdoc cref="ITextureProvider.GetFromFile(string)"/>
        [MethodImpl(MethodImplOptions.AggressiveInlining)]
        public SharedImmediateTexture.PureImpl GetFromFile(string path) =>
            this.GetFromFile(new FileInfo(path));
```

I understand that this is necessary because users might provide a relative path. 

A solution would be to add a method that explicitly expects a full path, thus taking away the need for the FileInfo instantiation.


goaaats (approved):

goaaats:
I think this would be nice to have, the build looks to be failing though due to a missing implementation - could you have a look?

goaaats (approved):

PR #1918, opened 2024-07-10
TextGimmickHint Services

Small-Miao:
Add a Service to call game TextGimmick Hint
image

### DIFF
--- a/Dalamud/Game/Gui/TextGimmickHint.cs
+++ b/Dalamud/Game/Gui/TextGimmickHint.cs
@@ -1,0 +1,42 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading.Tasks;
+
+using Dalamud.Plugin.Services;
+
+using FFXIVClientStructs.FFXIV.Client.UI;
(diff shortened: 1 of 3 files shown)


KazWolfe:
Is there a compelling reason to make this into a new service versus adding it to `ToastGui`?

MidoriKami:
It's objectively not a toast, and it's likely we'll want to expand it's features later.

Dalamud/Game/Gui/TextGimmickHint.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Text;
+using System.Threading.Tasks;
+
+using Dalamud.Plugin.Services;
+
+using FFXIVClientStructs.FFXIV.Client.UI;
+
+using static Dalamud.Plugin.Services.ITextGimmickHint;
+
+namespace Dalamud.Game.Gui
```
MidoriKami, line 14: I think file scoped namespaces are preferred? Did stylecop not complain about this?

Haselnussbomber:
CS was updated, so this is now available via `RaptureAtkModule.Instance()->ShowTextGimmickHint(...)`
No service needed.

PR #1923, opened 2024-07-12
Take ImGuiScene into Dalamud

Soreepeong:
Smaller part of #1542.

* Nothing in `Dalamud.Interface.ImGuiBackend` should be visible to plugins. They must be internal.
* No special ReShade handling will be needed. Dalamud would receive DirectX objects behind ReShade.
* External windows will support transparency.
* Added `--debug-directx` to enable D3D11/DXGI debugging.

Probably needs more testing with ReShade installed, though.


Soreepeong:
On 3c84b9c: seems that reshade recently changed IDXGISwapChain wrapper class layout, so previous 0x10 would have become 0x08. https://github.com/crosire/reshade/commit/aedef466b817b480aa3666782b180d8bc26d3b84#diff-963981c848bbb3106d4ee73bfdf53d556bd1bacfe901680995c7eb5eeb4a600d

PR #1926, opened 2024-07-13
Add PluginScoped Nags

MidoriKami:


MidoriKami:
Didn't notice GameInventoryTestWidget required instances of a GameInventoryPluginScoped, as it was the last service nag I added, assumed it would just work.

The GameInventoryTestWidget shouldn't be using scoped instances to test core functionality.

MidoriKami:
Removed GameInventory Nags for now.

KazWolfe:
Rather than doing this on a per-service scope, would it be possible to do this at the service manager level if a specific `NeedsCleanup` var is present? Would be cleaner and have less code everywhere, I'd think.

MidoriKami:
stale closing

PR #1927, opened 2024-07-13
Obsolete IconButton without Id

MidoriKami:
This is a common trap for devs where when using multiple icon buttons they share the same ImGui Id and become linked to each other. We have a overload already that takes an Id string.

Recommend merging now, and then removing function in api11

### DIFF
--- a/Dalamud/Interface/Components/ImGuiComponents.IconButton.cs
+++ b/Dalamud/Interface/Components/ImGuiComponents.IconButton.cs
@@ -52,6 +52,7 @@ public static partial class ImGuiComponents
     /// <param name="activeColor">The color of the button when active.</param>
     /// <param name="hoveredColor">The color of the button when hovered.</param>
     /// <returns>Indicator if button is clicked.</returns>
+    [Obsolete("This version does not set an id for ImGui to use, recommended to use overload with Id field.", false)]
     public static bool IconButton(FontAwesomeIcon icon, Vector4? defaultColor = null, Vector4? activeColor = null, Vector4? hoveredColor = null)
         => IconButton($"{icon.ToIconString()}", defaultColor, activeColor, hoveredColor);
 


Dalamud/Interface/Components/ImGuiComponents.IconButton.cs
```diff
@@ -52,6 +52,7 @@ public static bool IconButton(string iconText)
     /// <param name="activeColor">The color of the button when active.</param>
     /// <param name="hoveredColor">The color of the button when hovered.</param>
     /// <returns>Indicator if button is clicked.</returns>
+    [Obsolete("This version does not set an id for ImGui to use, recommended to use overload with Id field.", false)]
```
KazWolfe, line 55: `PushId` is a thing as well, are you sure you want to require double ID work?

MidoriKami: Can you elaborate on what you mean double Id work?

Since the Id's are an internal construct, people that are using PushId already, can either remove it, or can put whatever additional id they want, this comes at no real cost to usability.

KazWolfe: As a quick pseudocode example:

```cs
foreach (var thing in myThings) {
  _ = using ImRaii.PushId($"thingList-{thing.Id}");
  
  ImGui.Text($"Thing: {thing.Name}");

  if (IconButton(FontAwesome.MagnifyingGlass)) {
    thing.PrintInfo();
  }

  if (IconButton(FontAwesome.Trashcan)) {
    myThings.Remove(thing);
  }
}
```

In this case, the button already gets a unique ID and this change would require double-declaration of an ID where it's not strictly necessary.

MidoriKami: I feel this is a minority of cases, and this'll solve problems, but add a minor hassle to a minority of devs.

Folks using the code as you describe above would indeed likely just copy paste their Id into the IconButton field, but that comes at no harm.

Limiana: > minority of cases

More like almost 300: https://github.com/search?q=%22ImGuiComponents.IconButton%28FontAwesomeIcon%22&type=code&ref=advsearch&p=1

Most of the times you either way either want to use unique button icon in your UI or you're displaying in cycle and use PushId/PopId anyway. Cases where you'd use same icon outside of the loop aren't so very common I believe.

MidoriKami:
Closing because this is clearly unwanted, and doesn't effect me anyways.

PR #1929, opened 2024-07-14
Update version information for CI/metadata

KazWolfe:
This PR aims to revise our versioning logic to be a bit more sane in light of recent CI changes.

- Restore `GitSha` to actually be the Git SHA in our Kamori metadata.
  - Write the commit hash to `commit_hash.txt`.
  - Always output this commit hash, appending `-dirty` if necessary.
  - Change `Util#GetGitHash` to actually return the full git hash.
- Create new property `SCMVersion` which holds the version as determined by `git describe`
  - Swap over most use cases in code to now read this version string instead.
  - Retain `Local build at...` logic for debug builds for now
  - Most systems will be interested in this version over GitHash.


KazWolfe:
Closing - split into separate PRs.

PR #1934, opened 2024-07-15
Add UsedImplicitly to IDalamudPlugin

ionite34:
This adds the `UsedImplictly` Annotation to `IDalamudPlugin` to clear up warnings regarding plugin classes not being instantiated.

### DIFF
--- a/Dalamud/Plugin/IDalamudPlugin.cs
+++ b/Dalamud/Plugin/IDalamudPlugin.cs
@@ -5,7 +5,7 @@ namespace Dalamud.Plugin;
 /// <summary>
 /// This interface represents a basic Dalamud plugin. All plugins have to implement this interface.
 /// </summary>
-[UsedImplicitly(ImplicitUseKindFlags.InstantiatedNoFixedConstructorSignature,ImplicitUseTargetFlags.WithInheritors)]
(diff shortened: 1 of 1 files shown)


KazWolfe (approved):

KazWolfe:
Does this actually work if your plugin doesn't have `JetBrains.Annotations` as a dependency? I thought all of these annotations were stripped by the compiler.

ionite34:
> Does this actually work if your plugin doesn't have `JetBrains.Annotations` as a dependency? I thought all of these annotations were stripped by the compiler.

Actually yeah since the Release builds of Dalamud are used for development this is going to be stripped. (I noticed the existing `MeansImplicitUse` attribute on `[PluginService]` doesn't work either). 

The other options are either 
1. Including Jetbrains.Annotations in release builds
2. Adding a xml external annotations file https://www.jetbrains.com/help/resharper/Code_Analysis__External_Annotations.html and either add it to the release output or make a separate NuGet Package that is referenced by the Dalamud SDK.

KazWolfe:
Apologies for delay on this. Yeah, I think external annotations XMLs would be totally fine.

KazWolfe:
Going to tidy up and close this PR for now. If you want to re-submit the annotations XML, it'd be more than welcome.

Thanks!

PR #1938, opened 2024-07-16
Add an ULD display to data window

wolfcomp:
Implements an ULD browser to see what the different partsId and part indexes are showing.

https://github.com/user-attachments/assets/99d39f37-1d82-483b-b0d4-510b8b9b125e


wolfcomp:
Still not fully done checking through the exe with the script to see if there is missed uld locations but ready for code review otherwise.

Dalamud/Interface/Internal/Windows/Data/Widgets/UldWidget.cs
```diff
@@ -0,0 +1,1210 @@
...
+/// Contains the raw data for the ULD widget.
+/// </summary>
+internal class UldWidgetData
+{
+    /// <summary>
+    /// All known uld locations referenced in executable.
+    /// Generated from FFXIVClientStructs\ida\ffxiv_extract_uld_names.py as of 7.01
+    /// </summary>
+    #pragma warning disable SA1401 // Fields should be private - this needs to be internal or public for above class to access it
+    internal static string[] UldLocations =
```
KazWolfe, line 290: This will need to be generated and manually updated on a per-patch cycle?

I assume there's no automatic way to get it, nor any better place this can go then...

wolfcomp: > This will need to be generated and manually updated on a per-patch cycle?
Yesn't... Depends what kind of patch it is since they might not add uld files but instead change said file with new info.

> I assume there's no automatic way to get it, nor any better place this can go then...
I didn't fully look into how to see if I could get data from the asset path that say `bannedplugins` is loaded from, as it could be also located there. 

I will note that the list that said script currently generate isn't complete, as there are certain ones that it can find by sig scanning.

KazWolfe: As-is, I'm a bit wary of having a list that we need to update anywhere inside Dalamud, especially for something non-critical like a developer system. Any changes to this list risk going unnoticed for a while.

Moving forward, I think we'd need to either:

1. Add a plain text field to allow developers to import raw ULD names (this is probably a good idea anyways)
2. Add some sort of smarter auto-detection logic to find ULDs. The game ultimately needs to know about all the different ULD kinds, so we should be able to use the same systems to find things.

We've discussed point 2 in the Discord a bit, and it seems harder than expected as there's no good way to match for `ui/uld/...` in the EXE. Most ULDs are loaded via a call to `AtkUnitBase#LoadUldByName`, where the passed value comes from a series of constant strings in the binary that don't even have the politeness to be a proper array.

wolfcomp: I think I found a nice way to do the second one for ULD detection, with modifying the `SigScanner` a tiny bit to allow getting every match in the `.text` area of the game memory. This possibly will break if people modify said area, but not entirely sure.

KazWolfe (approved):

PR #1939, opened 2024-07-17
feat: Add Dalamud Version to the Title Screen

KazWolfe:
Because an image tells a thousand words:

image

- Indicates the current version string (via `GetScmVersion()`) for the current instance of Dalamud
- Indicates the target branch, as reported by Kamori.
  - If `DalamudBetaKey` is invalid or incorrect at evaluation time, this will report the new target branch.
  - If Kamori could not be contacted, the branch name as found in config will be reported with a question mark, e.g. `release?`
- Indicates the number of plugins currently enabled.

~~At present this PR does not allow users to hide this version string. I don't necessarily think there's any case where we *do* want to hide it, given that there's an increasing amount of plugins that are modifying native UI as well.~~ This is now hidden with the rest of TSM. As a result, I do need to continue hijacking things, but that should be a light operation.

This PR depends on #1929 for branch and version information.


MidoriKami:
Nice, I definitely think this should be a permanent addition that can't be toggled.

I'm sure we'll get people going "I wanna stream and this makes it so I can't" where we can inform them the dangers of plugin's with native UI modifications.

KazWolfe:
New version submitted for now, depends on #1945 being merged first.

I've removed the branch resolution logic for now as I think we have plans to rework that overall.

goaaats (approved):

goaaats:
I think this is fine. I'm not sure if it should be red, since that makes it seem like there's an error to me.

PR #1944, opened 2024-07-18
Service: make TCS RunContinuationsAsynchronously

Soreepeong:


goaaats (approved):

Soreepeong:
While there are more usages of `TaskCompletionSource` elsewhere, fix those when those exhibit problems after evaluating if RunContinuationsAsynchronously behavior is desired. For service initialization, it is definitely desired.

PR #1946, opened 2024-07-18
Limit log files to 100MB

KazWolfe:
This pull request aims to add a sanity limit to Dalamud's log size, currently capped to 100MB.

This value was arbitrarily chosen as it's high enough that no sane environment can ever hit it, and if it *was* hit, our concern should be to stop the bleeding rather than anything else. When this limit is hit, no new log entries will be added to the file.

There is probably room here to perform log rotations, replace the last line of the log file, print a message to chat, etc. I don't think any of these are particularly necessary for this, nor are they going to be worth the additional complication/digging into Serilog internals. This is meant to be quick and dirty, and I doubt many people are ever going to really hit this limit unless some plugin is very broken and likely spamming an exception every frame.


goaaats:
Applied without the version bump. Thanks!

KazWolfe:
ref 25cf7fde10a61da259e606dd65714f1a0153dfa9

PR #1947, opened 2024-07-19
Add separate handler for process modules

Soreepeong:
`Process.GetCurrentProcess().ProcessModules` turned out to be a bottleneck, so this PR will cache the Process object which caches ProcessModules internally, and invalidate only when necessary.

### DIFF
--- a/Dalamud/Support/CurrentProcessModules.cs
+++ b/Dalamud/Support/CurrentProcessModules.cs
@@ -1,0 +1,138 @@
+using System.Collections.Concurrent;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Serilog;
+
+using TerraFX.Interop.Windows;
+
+namespace Dalamud.Support;
+
+/// <summary>Tracks the loaded process modules.</summary>
+[ServiceManager.EarlyLoadedService]
+internal sealed unsafe partial class CurrentProcessModules : IInternalDisposableService
+{
+    private static readonly ConcurrentQueue<string> LogQueue = new();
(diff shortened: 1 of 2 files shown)


KazWolfe:
Can we sanely start using this new service as the provider to eventually replace our other GetCurrentProcess calls?

Soreepeong:
Depends on the purpose of it; if any aspect of Process is cached, then this service will only provide up-to-date ProcessModuleCollection; any other information can be outdated anytime.

goaaats:
To replace the xivfix: No, there are too many calls in libraries that expect it to work. Inside Dalamud, that seems fine.

KazWolfe:
Yeah, my question is more geared towards random uses like `Process.GetCurrentProcess().MainWindowHandle`, `Process.GetCurrentProcess().MainModule`, and I suppose `Process.GetCurrentProcess().Kill()`. 

These should be cheap-ish calls enough, but if they can be made even cheaper without any impact, might be good to do.

Soreepeong:
* MWH will provide invalid value in some cases; IUiBuilder has main window handle and that should be used instead, which guarantees that it's the game window.
* MainModule: GetModuleHandle(null) can be used as the cheapest alternative.
* Kill: it'll be the last thing dotnet allocates, so it probably won't matter.

Probably best to revisit again when the current process has to be touched every frame.

KazWolfe (approved):
LGTM and Addresses window is nice and fast again.

I'd like to change formatting eventually but that's not something this PR needs.

Secondarily, what's the purpose of LogQueue? Logging should be a safe operation, so I'm curious if there's something I'm missing that means it can't just be called normally.

PR #1952, opened 2024-07-21
Miscellaneous fixes

Soreepeong:
* Plugin constructors are now ran on separate threads, so that `Task.Result` and `Task.Wait()` inside there will not result in blocking random task thread pool threads.
* Made `InterfaceManager` wait for swap chain before setting title bar color according to dark mode.
* Suppressed ignorable errors from `CurrentProcessModules` and `DalamudAssetManager`.
* Changed `Localization` to a provided service, so that `LoadingDialog` can use `Loc`.
* `LoadingDialog` now does not depends on `System.Windows.Forms`, and instead it calls the underlying functions directly. This requires setting the activation context, which is not exposed by `System.Windows.Forms`. This suppresses thread association errors that are safely suppress-and-continuable.
* Made console window display all logs ever since initializing Serilog.
* Made loading dialog show last 20 lines of logs.
image


Dalamud/IoC/Internal/ServiceContainer.cs
```diff
@@ -127,7 +128,26 @@ public void RegisterInterfaces(Type type)
             return null;
         }
 
-        ctor.Invoke(instance, resolvedParams);
+        var tcs = new TaskCompletionSource(TaskCreationOptions.RunContinuationsAsynchronously);
```
goaaats, line 131: Could you leave a comment here explaining why this is necessary?

goaaats (changes requested):
Looks good code-wise!

Dalamud/IoC/Internal/ServiceContainer.cs
Soreepeong: I'm not sure if this fixes problems; more of seeing if this fixes stalling problems on load time.

Soreepeong:
Made all TaskCompletionSource have RunContinuationsAsynchronously passed too.

Dalamud/IoC/Internal/ServiceContainer.cs
goaaats: Alright, let's leave it like this for now and see if things work out

goaaats (approved):

PR #1955, opened 2024-07-21
Implement ReShade addon interface

Soreepeong:


KazWolfe:
Does this supersede `new_im_hooks`?

Soreepeong:
Tested working with previous version (6.1.1) and current version (6.2.0).

Soreepeong:
> Does this supersede `new_im_hooks`?

It is more of a choice between having stuff implemented in c# and cpp.

Dalamud/Interface/Internal/Windows/Settings/Tabs/SettingsTabExperimental.cs
```diff
@@ -55,6 +70,45 @@ public class SettingsTabExperimental : SettingsTab
...
+                ReShadeHandlingMode.None => Loc.Localize(
+                    "DalamudSettingsReShadeHandlingModeNone",
+                    "Do not handle"),
+                _ => "<invalid>",
+            },
+            FriendlyEnumDescriptionGetter = x => x switch
+            {
+                ReShadeHandlingMode.ReShadeAddon => Loc.Localize(
+                    "DalamudSettingsReShadeHandlingModeReShadeAddonDescription",
+                    "Dalamud will register itself as a ReShade addon. Most compatibility is expected, but multi-monitor window option won't work too well."),
```
goaaats, line 101: Could you describe what "won't work too well" means? Will you be able to use it at all/will it be affected by shaders?

Dalamud/Interface/Internal/ReShadeHandling/ReShadeUnwrapper.cs
```diff
@@ -0,0 +1,205 @@
...
+namespace Dalamud.Interface.Internal.ReShadeHandling;
+
+/// <summary>
+/// Peels ReShade off stuff.
+/// </summary>
+[SuppressMessage(
+    "StyleCop.CSharp.LayoutRules",
+    "SA1519:Braces should not be omitted from multi-line child statement",
+    Justification = "Multiple fixed blocks")]
+internal static unsafe class ReShadeUnwrapper
```
goaaats, line 19: I guess this seems to work, I'm curious how stable it is. Will this break if the inheritance hierarchy of those COM objects changes?

Dalamud/Interface/Internal/ReShadeHandling/ReShadeAddonInterface.cs
```diff
@@ -0,0 +1,176 @@
...
+    {
+        Exports.ReShadeUnregisterEvent(AddonEvent.InitSwapChain, this.initSwapChainDelegate);
+        Exports.ReShadeUnregisterEvent(AddonEvent.DestroySwapChain, this.destroySwapChainDelegate);
+        Exports.ReShadeUnregisterEvent(AddonEvent.ReShadeOverlay, this.reShadeOverlayDelegate);
+        Exports.ReShadeUnregisterAddon(this.hDalamudModule);
+        this.addonModuleResolverHook.Disable();
+        this.addonModuleResolverHook.Dispose();
+    }
+
+    private BOOL GetModuleHandleExWDetour(uint dwFlags, ushort* lpModuleName, HMODULE* phModule)
```
goaaats, line 113: It's a shame that this is necessary, but I guess there's no way around it for now. We should remove this as soon as possible when crosire releases a version with the HMODULE variants.

goaaats (changes requested):

goaaats:
Seems fine, what do you recommend the default to be? I guess it really depends on how Multi-Monitor interacts with addon mode. I would prefer that if it works well.

Another question is if we should somehow indicate that you are using a ReShade without addon support, in that case I guess it would just fall back to no handling at all. I think it's fine to leave it out for now, since I don't know how common it is. If it is quite common, we might want to show a warning on the settings page.

Dalamud/Interface/Internal/Windows/Settings/Tabs/SettingsTabExperimental.cs
Soreepeong: Viewport windows worked fine for me with ReShade, but every time it gets created it did have to load all over again.

Dalamud/Interface/Internal/ReShadeHandling/ReShadeUnwrapper.cs
Soreepeong: The interface definitions are part of Windows API, so they will not change. They are guaranteed to be based on IUnknown, so I don't see this breaking anytime soon. The only possible breakage is when ReShade wrapper object places pointer to the original COM object past 3rd member field or creates another indirection to it.

Dalamud/Interface/Internal/ReShadeHandling/ReShadeAddonInterface.cs
Soreepeong: I'll add a TODO about it.

Soreepeong:
image

Added alternative swapchain vtable hooking mode. This obsoletes `new_im_hooks` branch.

Soreepeong:
> Seems fine, what do you recommend the default to be? I guess it really depends on how Multi-Monitor interacts with addon mode. I would prefer that if it works well.

Probably best to leave it as addon, since it's what ReShade is guaranteeing to work well with.
 
> Another question is if we should somehow indicate that you are using a ReShade without addon support, in that case I guess it would just fall back to no handling at all. I think it's fine to leave it out for now, since I don't know how common it is. If it is quite common, we might want to show a warning on the settings page.

If they do that, they'll quickly figure out that ReShade isn't working as soon as they log into the game, so it probably isn't necessary for us to handle that case.

KazWolfe:
> If they do that, they'll quickly figure out that ReShade isn't working as soon as they log into the game, so it probably isn't necessary for us to handle that case.

I think it is given that users will ask for support anyways. If it's a case we can easily identify and mitigate, I don't see the harm in doing so.

Soreepeong:
While ReShade addon registration function will return false on failure, it does not give us why; it only writes to its own log file. I guess it's possible to add a notification that ReShade integration failed to initialize along with possible reasons?

Mortalitas:
> While ReShade addon registration function will return false on failure, it does not give us why; it only writes to its own log file. I guess it's possible to add a notification that ReShade integration failed to initialize along with possible reasons?

You may be overthinking it, as there's a much more straightforward way to check for add-on support: The ReShade (and GShade) binaries with addon support are unsigned; the ReShade binary without addon support is signed.

Just check for that and throw the error accordingly.

Soreepeong:
image

Changed to check if the DLL has signatures, and show a warning notification accordingly.

KazWolfe:
> The ReShade (and GShade) binaries with addon support are unsigned; the ReShade binary without addon support is signed.

Well, that’s absolutely terrifying. I wonder why ReShade doesn’t sign their addon-enabled version and, more critically, what would happen if they started. 

I suppose we can deal with that whenever it happens, but we should probably note that this might need a rework.

NotNite:
Few questions:

- I presume the GShade addon interface will stay stable with the ReShade interface, so this should continue to work with both?
- This means support for older ReShade versions is dropped, right? I remember reading about minimum addon API versions somewhere. Would be nice to support older versions but I don't think it's feasible to maintain.

Soreepeong:
> I presume the GShade addon interface will stay stable with the ReShade interface, so this should continue to work with both?

If GShade also supports addons, then it'll work fine.

> This means support for older ReShade versions is dropped, right? I remember reading about minimum addon API versions somewhere. Would be nice to support older versions but I don't think it's feasible to maintain.

We probably can put API version 1 since all we do is fetching the native object interface using reshade_present.

[Commit introducing reshade_present.](https://github.com/crosire/reshade/commit/5b49eff54bc7a235c82ed568452e74108195a961) [API version.](https://github.com/crosire/reshade/blob/5b49eff54bc7a235c82ed568452e74108195a961/include/reshade.hpp)

goaaats (approved):

crosire:
> I wonder why ReShade doesn’t sign their addon-enabled version and, more critically, what would happen if they started.

Just came across this by chance, so to answer that question: ReShade without add-on support is signed since it's the version that anti-cheat vendors can allow-list and the signature allows them them to automatically include all current and future ReShade versions in that safely. ReShade with add-on support is intentionally not signed, so that it will always trigger anti-cheat. This is by design and not subject to change, so is a reasonable way to distinguish the two.

KazWolfe:
> ReShade with add-on support is intentionally not signed, so that it will always trigger anti-cheat. This is by design and not subject to change, so is a reasonable way to distinguish the two.

Interesting and good to know we can rely on it moving forward. I'd have expected the addon'd version to be signed too if only for provenance and chain-of-trust purposes, but makes sense. I'm guessing addons wouldn't necessarily show themselves to anticheats accordingly?

crosire:
Both ReShade setup tools including the non-add-on and add-on version DLLs are signed and that's what one downloads from the website. So can at least verify that what was downloaded is to be trusted. But the DLLs have the aforementioned distinction.
It's less about the add-ons that could be loaded (any decent anti-cheat would notice other DLLs being loaded through ReShade), but rather that the version of ReShade with full add-on support also does not disable any of its functionality when there is network activity, while the normal, signed version does (e.g. to prevent access to the depth buffer in competitive multiplayer games, hence why ReShade is deemed safe for some by anti-cheat).

PR #1958, opened 2024-07-22
feat: Encourage developers to use DLLs for devPlugins

NotNite:
Discussed [here](https://canary.discord.com/channels/581875019861328007/860813266468732938/1264949160357134458).

Adds a button to pick a .dll for the dev plugin path, as well as removes the directory scan logic. While directory search hasn't been encouraged for a while now, there may still be people who use it, so this probably is gonna break someone's workflow.

image


Haselnussbomber:
Text above the plugin needs to be updated. Still says can be either directory or dll path.

NotNite:
> Text above the plugin needs to be updated. Still says can be either directory or dll path.

I'm blind

Haselnussbomber:
Sorry meant button, not plugin. But you got it. 👍

KazWolfe:
Can we change this such that dev plugins are still loaded in recursively but only individual paths can be saved for now?

If so, we can deprecate that load behavior in API 11 and get a proper announcement out to everyone.

Dalamud/Plugin/Internal/PluginManager.cs
```diff
@@ -786,11 +786,7 @@ public void ScanDevPlugins()
             
             Log.Verbose("Scanning dev plugins at {Path}", setting.Path);
 
-            if (Directory.Exists(setting.Path))
-            {
-                devDllFiles.AddRange(new DirectoryInfo(setting.Path).GetFiles("*.dll", SearchOption.AllDirectories));
-            }
-            else if (File.Exists(setting.Path))
```
KazWolfe, line 793: For now, let's keep the recursive loading mode but just block setting folders in UI.

Removing this behavior should be an api11 todo so we have time to announce it.

Dalamud/Interface/Internal/Windows/Settings/Widgets/DevPluginsSettingsEntry.cs
```diff
@@ -68,7 +70,23 @@ public override void Draw()
             }
         }
 
-        ImGuiHelpers.SafeTextColoredWrapped(ImGuiColors.DalamudGrey, Loc.Localize("DalamudSettingsDevPluginLocationsHint", "Add dev plugin load locations.\nThese can be either the directory or DLL path."));
+        ImGuiHelpers.SafeTextColoredWrapped(ImGuiColors.DalamudGrey, Loc.Localize("DalamudSettingsDevPluginLocationsHint", "Add dev plugin load locations.\nThis must be a path to the plugin DLL."));
+
+        var locationSelect = Loc.Localize("DalamudDevPluginLocationSelect", "Select Dev Plugin Location");
```
KazWolfe, line 75: Please change this label to `Select DLL` or similar.

KazWolfe (changes requested):

Dalamud/Interface/Internal/Windows/Settings/Widgets/DevPluginsSettingsEntry.cs
KazWolfe, line 75:
```suggestion
        var locationSelect = Loc.Localize("DalamudDevPluginLocationSelect", "Select Dev Plugin DLL");
```

PR #1960, opened 2024-07-23
Add SerpentCombo to VPRGauge

attickdoor:
Weirdly, `git gui` tells me there were some hidden characters at the start of the `using` line in `VPRGauge.cs`.

### DIFF
--- a/Dalamud/Game/ClientState/JobGauge/Enums/SerpentCombo.cs
+++ b/Dalamud/Game/ClientState/JobGauge/Enums/SerpentCombo.cs
@@ -1,0 +1,42 @@
+namespace Dalamud.Game.ClientState.JobGauge.Enums;
+
+/// <summary>
+/// Enum representing the SerpentCombo actions for the VPR job gauge.
+/// </summary>
+public enum SerpentCombo : byte
+{
+    /// <summary>
+    /// No Serpent combo is active.
+    /// </summary>
+    NONE = 0,
+
+    /// <summary>
+    /// Death Rattle action.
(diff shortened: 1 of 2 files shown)


Dalamud/Game/ClientState/JobGauge/Enums/SerpentCombo.cs
```diff
@@ -0,0 +1,42 @@
+namespace Dalamud.Game.ClientState.JobGauge.Enums;
+
+/// <summary>
+/// Enum representing the SerpentCombo actions for the VPR job gauge.
+/// </summary>
+public enum SerpentCombo : byte
```
KazWolfe, line 6: Please use `PascalCase` for enum members.

KazWolfe (changes requested):

Dalamud/Game/ClientState/JobGauge/Enums/SerpentCombo.cs
attickdoor: I'm following the convention of many other enums in this folder, even as recently as `BeastChakra.cs`. Only the PCT and other VPR enums use `PascalCase`.

KazWolfe: I would still prefer we clean up and move everything over to PascalCase moving forward, I'm guessing we missed a fair bit.

That said, I'll merge this change in as-is, and we can schedule a cleanup in a future API version.

KazWolfe (approved):

PR #1966, opened 2024-07-24
ServiceScope fixes

Soreepeong:
* Changed `IServiceScope.CreateAsync` to never return null, and instead store the exception in the returned Task.
* Changed `ServiceScope` to use `ConcurrentDictionary<Type, Task<object>>` to manage created objects.
* Added `IDalamudPluginInterface.CreateAsync` to offer an alternative avoiding using `.Wait`. `IDalamudPluginInterface.Create` will work as it always did; only a small note has been added that describes what happens on failure, and its actual behavior did not change.
* Added `IDalamudPluginInterface.InjectAsync` for same reasons.


Soreepeong:
`IDalamudPluginInterface.Create` and `.Inject` should not be used if `IAsyncDalamudPlugin` from #1905 is to be used. `.Wait()` and `await` should not be mixed; `.Wait()` should not be used from anything that may be running under a Task context.

goaaats (approved):

PR #1970, opened 2024-07-24
Use ReShade addon event Present

Soreepeong:
https://discord.com/channels/586242553746030596/938951838227632218/1265785111660597248
https://discord.com/channels/581875019861328007/653504487352303619/1265782095771340841

capture_20240725_063005_001
https://github.com/crosire/reshade/commit/8100504d557f1949a5404e852b022f6b246bd6e5


Soreepeong:
We should use `reshade_overlay` on future ReShade versions, since it is possible that other ReShade addons might be postprocessing from `present` event callbacks, and the order of handling probably is undefined that other addons may postprocess over Dalamud UI.

KazWolfe (approved):
I don't understand anything this code does. Approved! 👍

PR #1973, opened 2024-07-25
add back DXGISwapChain::on_present hook as default for now

Soreepeong:
image

Default = hook on_present


Dalamud/Configuration/Internal/DalamudConfiguration.cs
```diff
@@ -444,7 +444,7 @@ public string EffectiveLanguage
     public bool WindowIsImmersive { get; set; } = false;
 
     /// <summary>Gets or sets the mode specifying how to handle ReShade.</summary>
-    public ReShadeHandlingMode ReShadeHandlingMode { get; set; } = ReShadeHandlingMode.ReShadeAddon;
+    public ReShadeHandlingMode ReShadeHandlingMode { get; set; } = ReShadeHandlingMode.Default;
```
goaaats, line 447: I'll rename this property after merging so that people that have used stg will get the new default

goaaats (approved):

PR #1978, opened 2024-07-27
Make DtrBar more threadsafe

Soreepeong:
* Changed DtrBar to use ReaderWriterLockSlim so that there exists only one storage of entries, preventing possible desync.
* DtrBarEntry will now hold a reference to the LocalPlugin that created the entry, so that DtrBarPluginScoped can defer plugin related handling to the main service.
* Marked DtrBarEntry class itself to be turned internal in API 11.
* Made IDtrBar.Entries return an immutable copy of underlying list of DtrBar entries, that will be freshly created whenever the list changes.

### DIFF
--- a/Dalamud/Game/Gui/Dtr/DtrBar.cs
+++ b/Dalamud/Game/Gui/Dtr/DtrBar.cs
@@ -403,7 +403,7 @@ internal sealed unsafe class DtrBar : IInternalDisposableService, IDtrBar
         var additionalWidth = 0;
         AtkResNode* collisionNode = null;
 
-        foreach (var index in Enumerable.Range(0, addon->UldManager.NodeListCount))
+        for (var index = 0; index < addon->UldManager.NodeListCount; index++)
         {
             var node = addon->UldManager.NodeList[index];
             if (node->IsVisible())
--- a/Dalamud/Game/Gui/Dtr/DtrBar.cs
+++ b/Dalamud/Game/Gui/Dtr/DtrBar.cs
@@ -317,11 +317,10 @@ internal sealed unsafe class DtrBar : IInternalDisposableService, IDtrBar
         foreach (var data in this.entries)
         {
             if (!data.Added)
-            {
-                data.TextNode = this.MakeNode(++this.runningNodeIds);
-                data.Added = this.AddNode(data.TextNode);
-                data.Dirty = true;
-            }
+                data.Added = this.AddNode(data);
+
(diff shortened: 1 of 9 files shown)


Dalamud/Game/Gui/Dtr/DtrBar.cs
```diff
@@ -290,21 +368,30 @@ private void Update(IFramework unused)
...
+            // Is the node going away already, or already has been added?
+            if (entry.ShouldBeRemoved || entry.Added)
+                continue;
+
+            // Add the entry to the end of the order list, if it's not there already.
+            if (!dtrOrder.Contains(entry.Title))
+                dtrOrder.Add(entry.Title);
         }
+
+        this.ApplySort(dtrOrder);
```
nebel, line 392: This sort used to run only when new entries were added, but now it runs each frame. Would it be possible to run this only if a change was found in the above loop?

Soreepeong: Changed to sort on adding to entries.

Soreepeong:
Also added mouse cursor follow the up/down button when reordering from Dalamud Settings dtr tab.

Dalamud/Plugin/Services/IDtrBar.cs
```diff
@@ -24,11 +26,13 @@ public interface IDtrBar
...
     /// <exception cref="ArgumentException">Thrown when an entry with the specified title exists.</exception>
-    public IDtrBarEntry Get(string title, SeString? text = null);
+    IDtrBarEntry Get(string title, SeString? text = null);
 
     /// <summary>
     /// Removes a DTR bar entry from the system.
     /// </summary>
     /// <param name="title">Title of the entry to remove.</param>
-    public void Remove(string title);
+    /// <remarks>Remove operation is not immediate. Attempts to call <see cref="Get"/> immediately after calling this
```
goaaats, line 35: How does it fail? How does it need to be handled by the plugin? Can we just revive the entry instead in Get so that the removal is a no-op or is it a multi-step process?

goaaats (changes requested):

Dalamud/Plugin/Services/IDtrBar.cs
Soreepeong: That was describing the previous behavior; changed so that remove/get pairs will always work.

goaaats (approved):

PR #1979, opened 2024-07-27
[master] Update ClientStructs

bleatbot:


MidoriKami:
As noted here these are known breaking changes https://github.com/aers/FFXIVClientStructs/pull/1076#issuecomment-2254189272 , suspected that there are no plugins this will effect, the layout and format of the data was incorrect pre 7.0, and 7.0 very much broke it much more.

DailyDuty uses values that have had their offsets corrected in this PR.

goaaats (approved):

PR #1983, opened 2024-07-29
SeString renderer: fix colors, add link support

Soreepeong:
* Moved ImGuiSeStringRenderer outside `Internal` namespace, so that `SeStringRenderStyle` can be exposed.
* Added link support.
* Fixed wrong colors.
image


Soreepeong:
Change DalamudLinkPayload so that it can be encoded with `<link(0x0E,commandIdNumber,extraNumber1,extraNumber2,\["pluginName"\, "extraText"\])>Link text goes here<link(0xCE)>`.

KazWolfe (approved):
You're insane.

PR #1985, opened 2024-07-29
Add events for ClassJob and Level change

Haselnussbomber:
This PR adds 2 new events to ClientState:

- `ClassJobChanged(uint classJobId)`
- `LevelChange(uint classJobId, uint level)`

The events are fired *after* Agents are informed of the change to make sure everything is already updated appropriately.

### DIFF
--- a/Dalamud/Plugin/Services/IClientState.cs
+++ b/Dalamud/Plugin/Services/IClientState.cs
@@ -10,9 +10,33 @@ namespace Dalamud.Plugin.Services;
 public interface IClientState
 {
     /// <summary>
+    /// A delegate type used for the <see cref="ClassJobChanged"/> event.
+    /// </summary>
+    /// <param name="classJobId">The new ClassJob id.</param>
+    public delegate void ClassJobChangeDelegate(uint classJobId);
+
+    /// <summary>
+    /// A delegate type used for the <see cref="LevelChanged"/> event.
+    /// </summary>
+    /// <param name="classJobId">The ClassJob id.</param>
+    /// <param name="level">The level of the corresponding ClassJob.</param>
+    public delegate void LevelChangeDelegate(uint classJobId, uint level);
+
+    /// <summary>
     /// Event that gets fired when the current Territory changes.
     /// </summary>
     public event Action<ushort> TerritoryChanged;
+
+    /// <summary>
+    /// Event that fires when a characters ClassJob changed.
+    /// </summary>
+    public event ClassJobChangeDelegate? ClassJobChanged;
+
+    /// <summary>
+    /// Event that fires when a characters level changed.<br/>
+    /// This event also gets fired when the player changes jobs.
+    /// </summary>
+    public event LevelChangeDelegate? LevelChanged;
 
     /// <summary>
     /// Event that fires when a character is logging in, and the local character object is available.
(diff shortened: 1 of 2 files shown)


KazWolfe (approved):
lgtm, thank you!

Dalamud/Plugin/Services/IClientState.cs
```diff
@@ -9,11 +9,35 @@ namespace Dalamud.Plugin.Services;
...
+    /// A delegate type used for the <see cref="ClassJobChanged"/> event.
+    /// </summary>
+    /// <param name="classJobId">The new ClassJob id.</param>
+    public delegate void ClassJobChangeDelegate(uint classJobId);
+
+    /// <summary>
+    /// A delegate type used for the <see cref="LevelChanged"/> event.
+    /// </summary>
+    /// <param name="classJobId">The ClassJob id.</param>
+    /// <param name="level">The level of the corresponding ClassJob.</param>
```
Haselnussbomber, line 22: Actually incorrect, whoops.

Dalamud/Plugin/Services/IClientState.cs
```diff
@@ -9,11 +9,35 @@ namespace Dalamud.Plugin.Services;
...
 
+    /// <summary>
+    /// Event that fires when a characters ClassJob changed.
+    /// </summary>
+    public event ClassJobChangeDelegate? ClassJobChanged;
+
+    /// <summary>
+    /// Event that fires when a characters level changed.<br/>
+    /// This event might get fired for a different ClassJob than the player is currently on (e.g. in PvP matches).
+    /// </summary>
```
KazWolfe, line 38: Quick docfix pass.

```suggestion
    /// <summary>
    /// Event that fires when <em>any</em> character level changes, including levels
    /// for a not-currently-active ClassJob (e.g. PvP matches, DoH/DoL).
    /// </summary>
```

Dalamud/Plugin/Services/IClientState.cs
```diff
@@ -9,11 +9,35 @@ namespace Dalamud.Plugin.Services;
 /// </summary>
 public interface IClientState
 {
+    /// <summary>
+    /// A delegate type used for the <see cref="ClassJobChanged"/> event.
+    /// </summary>
+    /// <param name="classJobId">The new ClassJob id.</param>
+    public delegate void ClassJobChangeDelegate(uint classJobId);
```
KazWolfe, line 16: Would it make sense to expose `LazyRow<ClassJob>` here (and elsewhere) instead for convenience? I don't have an objection to just the `uint`, but it's probably worth asking.

KazWolfe (changes requested):
Just a minor docfix and a question.

Dalamud/Plugin/Services/IClientState.cs
Haselnussbomber: I actually didn't think of that, because `TerritoryChanged` event also just passes an `ushort`.
If you want me to add it to the new events, sure.

KazWolfe: I don't really have a good answer atm if it'd be "more appropriate" to return the rich type or the plain ID. One makes it easier to grab related data (although in almost all cases, the plugin dev will already be able to resolve their own sheet), and the other would ensure maximum compatibility. 

We can always change this later I suppose, so I don't think there's any harm in leaving it as ID-only. I'll leave this open for a little bit though to let anyone with Strong Opinions:tm: make their feelings known.

PR #1986, opened 2024-07-29
Rework AtkArrayDataBrowserWidget

Haselnussbomber:
Complete rework of the Atk Array Data tab in /xldata.

I've included the name enums for now, because https://github.com/aers/FFXIVClientStructs/pull/1077 isn't merged yet.

- Number Arrays:
NumberArrays

- String Arrays:
StringArrays

  - Text Address can be toggled to Entry Address when holding shift (or should I make a new column?)
String Entry Address

  - Search for text:
StringArraysSearch

- Extend Arrays:
ExtendArrays
These just hold pointers to structs, so not much to see here.

- Addons that are subscribed to an array will be listed too:
Subscribed Addons Tooltip
Subscribed means if the array data is updated, their OnRequestedUpdate function will be called.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/Data/Widgets/AtkArrayDataBrowserWidget.cs
+++ b/Dalamud/Interface/Internal/Windows/Data/Widgets/AtkArrayDataBrowserWidget.cs
@@ -1,7 +1,13 @@
-﻿using System.Numerics;
+using System.Numerics;
 
-using Dalamud.Memory;
+using Dalamud.Interface.Utility.Raii;
+
+using FFXIVClientStructs.FFXIV.Client.UI;
+using FFXIVClientStructs.FFXIV.Component.GUI;
+
 using ImGuiNET;
+
+using Lumina.Text.ReadOnly;
 
 namespace Dalamud.Interface.Internal.Windows.Data.Widgets;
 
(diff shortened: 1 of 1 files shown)


Dalamud/Interface/Internal/Windows/Data/Widgets/AtkArrayDataBrowserWidget.cs
```diff
@@ -28,148 +46,645 @@ public void Load()
...
+
+            ImGui.TableNextColumn(); // Pointer
+            if (!isNull)
+                DrawCopyableText($"0x{(nint)array->DataArray[i]:X}", "Copy address");
         }
     }
+
+    // Temporarily stored in Dalamud until https://github.com/aers/FFXIVClientStructs/pull/1077 is merged.
+    #pragma warning disable SA1201
+    private enum NumberArrayType
```
KazWolfe, line 402: Please remove and move to CS.

(Yes, I know, we're blocked by CS here.)

KazWolfe (changes requested):
Everything else LGTM, just going to wait for CS to merge their side.

Dalamud/Interface/Internal/Windows/Data/Widgets/AtkArrayDataBrowserWidget.cs
Haselnussbomber: PR was merged, I removed the enums and checked that it builds (it does).
The CI build here will obviously fail because CS isn't updated.

KazWolfe (approved):

PR #1994, opened 2024-08-03
Fix ImDrawList::AddImageRounded

Soreepeong:
The function makes an assumption that there exists 1 font atlas texture, so that `ImDrawList._Data->TexIdCommon` ==
`ImDrawList._CmdHeader.TextureId`. Since we support multiple font atlas textures, that assumption is no longer true and
`ImDrawList::AddConvexPolyFilled` will create a new draw command as needed, giving `ImGui::ShadeVertsLinearUV` a clean draw command to work with.

This workaround forcefully sets *the* font atlas texture to be the texture the user is trying to draw for the duration of drawing polygons and shading those vertices again, so that no draw command change happens. Once the operation is done, font atlas texture is reverted back to what it was.

This fix is done without thread safety concerns, but an `ImDrawList` should not be touched from multiple threads at a single time, so this is fine.

### DIFF
--- a/Dalamud/Interface/Internal/ImGuiDrawListFixProvider.cs
+++ b/Dalamud/Interface/Internal/ImGuiDrawListFixProvider.cs
@@ -28,10 +28,12 @@ internal sealed unsafe class ImGuiDrawListFixProvider : IInternalDisposableServi
 {
     private const int CImGuiImDrawListAddPolyLineOffset = 0x589B0;
     private const int CImGuiImDrawListAddRectFilled = 0x59FD0;
+    private const int CImGuiImDrawListAddImageRounded = 0x58390;
     private const int CImGuiImDrawListSharedDataTexIdCommonOffset = 0;
 
     private readonly Hook<ImDrawListAddPolyLine> hookImDrawListAddPolyline;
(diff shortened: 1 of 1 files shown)


Soreepeong:
https://github.com/goatcorp/gc-imgui/blob/e4920dc10cbf32c0cd9b4a48a4a71c8814242460/imgui_draw.cpp#L1669-L1693

`ImDrawList::PathFillConvex` in turn calls `ImDrawList::AddConvexPolyFilled`, which will push and pop the common texture ID as needed during its scope, which will often end up adding an empty draw command after it's done.
https://github.com/goatcorp/gc-imgui/blob/e4920dc10cbf32c0cd9b4a48a4a71c8814242460/imgui_draw.cpp#L993-L1001

`ImGui::ShadeVertsLinearUV` works on the most recent draw command.
https://github.com/goatcorp/gc-imgui/blob/e4920dc10cbf32c0cd9b4a48a4a71c8814242460/imgui_draw.cpp#L1909-L1931

KazWolfe (approved):

KazWolfe:
At what point are we just going to hardfork ImGui?

Soreepeong:
We already did; we're just trying to not rebuild cimgui.dll for the time being. Proper fix should go in with ImGui version update

PR #2001, opened 2024-08-05
fix: XivChatType respects the configured chat channel if not set

KazWolfe:
Fixes a bug where messages created using `new XivChatType` would always go to the `Debug` channel, regardless of what the user has set.

Behavioral change, but also patching a bug.


KazWolfe:
API breaking due to type change (thanks, enums and apicompat). Holding for API 11.

PR #2003, opened 2024-08-06
Add ITextureProvider.ConvertToKernelTexture

Soreepeong:
Lets you obtain an instance of Kernel::Texture from IDalamudTextureWrap.

Context: https://discord.com/channels/581875019861328007/653504487352303619/1270253907485921280

Displaying contents that is difficult to represent using game's native UI toolkit might be a common use case (eg. unsupported glyphs for various languages), so supporting displaying textures on them could be nice.

### DIFF
--- a/Dalamud/Plugin/Services/ITextureProvider.cs
+++ b/Dalamud/Plugin/Services/ITextureProvider.cs
@@ -5,7 +5,6 @@ using System.Reflection;
 using System.Threading;
 using System.Threading.Tasks;
 
-using Dalamud.Interface.Internal;
 using Dalamud.Interface.Internal.Windows.Data.Widgets;
 using Dalamud.Interface.Textures;
 using Dalamud.Interface.Textures.TextureWraps;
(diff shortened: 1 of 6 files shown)


Dalamud/Plugin/Services/ITextureProvider.cs
```diff
@@ -281,4 +280,12 @@ public interface ITextureProvider
...
     /// <remarks><para>This function does not throw exceptions.</para></remarks>
     bool IsDxgiFormatSupportedForCreateFromExistingTextureAsync(int dxgiFormat);
+
+    /// <summary>Converts an existing <see cref="IDalamudTextureWrap"/> instance to a new instance of
+    /// <see cref="FFXIVClientStructs.FFXIV.Client.Graphics.Kernel.Texture"/>.</summary>
+    /// <param name="wrap">Instance of <see cref="IDalamudTextureWrap"/> to convert.</param>
+    /// <param name="leaveWrapOpen">Whether to leave <paramref name="wrap"/> non-disposed when the returned
+    /// <see cref="Task{TResult}"/> completes.</param>
+    /// <returns>Address of the new <see cref="FFXIVClientStructs.FFXIV.Client.Graphics.Kernel.Texture"/>.</returns>
+    nint ConvertToKernelTexture(IDalamudTextureWrap wrap, bool leaveWrapOpen = false);
```
goaaats, line 290: I wonder if it would be wise to write a little bit about lifecycle here? How are you supposed to use this, what happens when you call this with texture wraps from shared textures, what are you supposed to do with the texture you're replacing and do you need to do anything to what you get back here when the texture is no longer used by the game?

goaaats (changes requested):

goaaats (approved):

PR #2005, opened 2024-08-11
Add additional warning to plugin config reset modal

Xpahtalo:
Adds a little warning to the config reset modal if you are subscribed to the testing version. It also further explains that this will delete plugin settings, not something related to Dalamud's config about the plugin. Here's how it should look if you are subscribed to testing.
image
We can't stop anyone from spreading bad info, but hopefully having an extra warning might get people to stop and ask more questions.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
@@ -98,6 +98,7 @@ internal class PluginInstallerWindow : Window, IDisposable
 
     private bool deletePluginConfigWarningModalDrawing = true;
     private bool deletePluginConfigWarningModalOnNextFrame = false;
+    private bool deletePluginConfigWarningModalExplainTesting = false;
     private string deletePluginConfigWarningModalPluginName = string.Empty;
     private TaskCompletionSource<bool>? deletePluginConfigWarningModalTaskCompletionSource;
 
@@ -732,7 +733,7 @@ internal class PluginInstallerWindow : Window, IDisposable
             }
         }
     }
-
+    
     private void DrawFooter()
     {
         var configuration = Service<DalamudConfiguration>.Get();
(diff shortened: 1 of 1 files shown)


KazWolfe (commented):
Thinking in general:

* Might it be clearer to call this a "factory reset" of a given plugin? Does that convey the same intent in a clearer fashion?
* I'm not particularly a fan of "this won't remove you from testing" since users shouldn't be clicking in to this modal for that anyways, but that's I suppose why we got here in the first place...

Something to note though: for some plugins, you will need to reset configs to downgrade to stable. If the new config version is incompatible with the stable version of a plugin, there could very well be load errors and other annoyances.

reiichi001:
> Something to note though: for some plugins, you will need to reset configs to downgrade to stable. If the new config version is incompatible with the stable version of a plugin, there could very well be load errors and other annoyances.

In this case, wouldn't serialization typically fail and reset the config anyways? While it's possible to craft up edge case and unique scenarios where a field kept the same name, but has an invalid value for an older version of the plugin, that would be reasonably easy for support to catch and notify a user if they report an issue, given it'd throw exceptions. (And is another reason plugin devs are encouraged to version config or note breakages, but aren't exactly expected to support downgrades, and we shouldn't really require that either. A user opting into testing is already consenting to those risks, whether they read the warning or forgot it.)

We do also have a separate modal that shows when a user is switching from a test version to a lower stable version, warning them about config. (Whether or not the user reads this or understands the ramifications is outside our control).
https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs#L2881
https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs#L942
https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs#L4258

Xpahtalo:
> Might it be clearer to call this a "factory reset" of a given plugin? Does that convey the same intent in a clearer fashion?
I agree that renaming this is probably worth it. Either something like "factory reset" which is well understood to be destructive, or something like "delete all plugin settings" to be more obvious that you are getting rid of something. 

> I'm not particularly a fan of "this won't remove you from testing" since users shouldn't be clicking in to this modal for that anyways, but that's I suppose why we got here in the first place...
It's a bit ugly and means the code has to reference things that are unrelated to the current action, but we clearly have lots of users that could benefit from having explicit explanations. 

> Something to note though: for some plugins, you will need to reset configs to downgrade to stable. If the new config version is incompatible with the stable version of a plugin, there could very well be load errors and other annoyances.
The modal for disabling testing somewhat mentions this, but we could add "unless told to by the plugin author" to the warning. I'm of the opinion it's unnecessary, though. From what I've seen, the devs that heavily utilize the testing branch to request early feedback are also active in Discord to support those situations.

goaaats (approved):

PR #2008, opened 2024-08-11
Add IGameInventory.GetInventoryItems

Haselnussbomber:
This PR adds a new `GetInventoryItems` function to `IGameInventory`, which returns a `ReadOnlySpan<GameInventoryItem>` of all slots in the container.

I can't think of any inventory container that ues symbolic links. My guess is that they are only used in Agents and such, so we don't have to worry about it. I even checked the containers MailEdit and HandIn, surprisingly no symbolic links.
I guess we'll hear when someone has trouble with it.

Screenshot

### DIFF
--- a/Dalamud/Game/Inventory/GameInventory.cs
+++ b/Dalamud/Game/Inventory/GameInventory.cs
@@ -404,6 +404,9 @@ internal class GameInventoryPluginScoped : IInternalDisposableService, IGameInve
     public event IGameInventory.InventoryChangedDelegate<InventoryItemMergedArgs>? ItemMergedExplicit;
 
     /// <inheritdoc/>
+    public ReadOnlySpan<GameInventoryItem> GetInventoryItems(GameInventoryType type) => GameInventoryItem.GetReadOnlySpanOfInventory(type);
+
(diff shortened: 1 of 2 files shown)


Dalamud/Game/Inventory/GameInventory.cs
```diff
@@ -403,6 +403,9 @@ internal class GameInventoryPluginScoped : IInternalDisposableService, IGameInve
     /// <inheritdoc/>
     public event IGameInventory.InventoryChangedDelegate<InventoryItemMergedArgs>? ItemMergedExplicit;
 
+    /// <inheritdoc/>
+    public ReadOnlySpan<GameInventoryItem> GetInventoryItems(GameInventoryType type) => GameInventoryItem.GetReadOnlySpanOfInventory(type);
```
KazWolfe, line 407: How is it that this use case hasn't come up yet?

Haselnussbomber: Because everyone that needs inventory items uses `InventoryManager` directly? I don't know. 🤷‍♂️

goaaats (approved):

PR #2016, opened 2024-08-17
Make ServiceScope IAsyncDisposable

Soreepeong:
ServiceScope.Dispose was not waiting for scoped services to complete disposing. This had an effect of letting a new plugin instance register a DtrBar entry before previous plugin instance's entry got unregistered.

This change also cleans up unloading procedure in LocalPlugin.

### DIFF
--- a/Dalamud/Utility/TaskExtensions.cs
+++ b/Dalamud/Utility/TaskExtensions.cs
@@ -63,20 +63,34 @@ public static class TaskExtensions
 #pragma warning restore RS0030
     }
 
-    /// <summary>Ignores any exceptions thrown from the task.</summary>
-    /// <param name="task">Task to ignore exceptions.</param>
-    /// <returns>A task that completes when <paramref name="task"/> completes in any state.</returns>
-    public static async Task SuppressException(this Task task)
-    {
-        try
-        {
-            await task;
-        }
-        catch
-        {
-            // ignore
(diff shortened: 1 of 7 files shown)


Dalamud/Utility/TaskExtensions.cs
```diff
@@ -63,6 +63,21 @@ public static T GetResultSafely<T>(this Task<T> task)
 #pragma warning restore RS0030
     }
 
+    /// <summary>Ignores any exceptions thrown from the task.</summary>
```
goaaats, line 66: Might be good to note that this awaits the task and isn't a "modifier", looks good otherwise!

goaaats (approved):

Dalamud/Utility/TaskExtensions.cs
Soreepeong: It effectively is a modifier, as it gets and discards the exception raised if any. Caller can optionally either await for the task passed as the parameter, or the returned task. It should be taken as an another version of:
```cs
Task SuppressException(Task task) => task.ContinueWith(r => _ = r.Exception);
```

Soreepeong: Actually I think I get what you mean there; gonna add more docs there

goaaats (approved):

PR #2018, opened 2024-08-18
feat: Use Debug Command Handler for Dalamud Commands

KazWolfe:
This pull request aims to improve the command handler experience by attaching a hook to the InvokeDebugCommand method. This allows Dalamud to receive all inbound commands, including those that have been directly injected or normally suppressed by `/merror off`. This replaces the legacy regex method of command evaluation, increasing speed and stability in case of system changes.

This method keeps Dalamud as the last command handler in the chain, so both standard and debug commands will take precedence. 

---

The InvokeDebugCommand method is called on any input to `ExecuteCommandInner`, which includes chat messages/non-commands. If a certain result is marked as a command but is not recognized, this method returns -1. If a certain result is a non-command (does not start with `/`), this method returns -2. InvokeDebugCommand is not called at all for "real" commands.

This hook will call `InvokeDebugCommand` normally and check its return code. If it matches `-1`, the hook passes the message contents to the Dalamud Command Handler for processing. If DCH succeeds in finding a command, this method will return `0` indicating to the game that a debug command was handled. If DCH fails to find a command, this method will return the original -1 causing the game to return the standard "command not found" response.


Dalamud/Game/Command/CommandManager.cs
```diff
@@ -2,58 +2,52 @@
...
-        this.chatGui.CheckMessageHandled += this.OnCheckMessageHandled;
+        this.resolver = new CommandManagerAddressResolver();
+        this.resolver.Setup(scanner);
+        
+        this.testCommandHook = Hook<TestCommandDelegate>.FromAddress(
+            this.resolver.CommandErrorHandler,
+            this.OnTestCommand);
+        this.testCommandHook.Enable();
+
+        // this.chatGui.CheckMessageHandled += this.OnCheckMessageHandled;
```
goaaats, line 46: What's going on here? Good to remove?

Dalamud/Game/Command/CommandManagerAddressResolver.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// <inheritdoc/>
+    protected override void Setup64Bit(ISigScanner sig)
+    {
+        this.SetupFramework(sig);
+    }
+    
+    private void SetupFramework(ISigScanner scanner)
+    {
+        this.CommandErrorHandler = 
+            scanner.ScanText("E8 ?? ?? ?? ?? 83 F8 ?? 74 ?? 83 F8 ?? 0F 85 ?? ?? ?? ?? 48 8B 07");
```
goaaats, line 23: Could we have this in CS?

goaaats (changes requested):

Dalamud/Game/Command/CommandManagerAddressResolver.cs
KazWolfe: https://github.com/aers/FFXIVClientStructs/pull/1102

KazWolfe:
merge after 7.06hf2 stable release.

goaaats (approved):

PR #2022, opened 2024-08-19
Update to Lumina 5 (new Excel parsing)

WorkingRobot:
New Features:
- Transparent RSV resolution using a hook to `LayoutWorld.AddRsvString`.
- Removal of `ExcelResolver`. `Lumina.Excel.RowRef` works as a better and more featured alternative now that lazy construction is obsolete/useless.
- `IDataManager` service now accepts a `string? name` optional parameter for explicit sheet names.
- `IDataManager.GetSheet` is now split into `IDataManager.GetSubrowSheet` as well.
- `IDataManager.Get(Subrow)Sheet` functions now throw. Not sure if this is 100% a good idea, but I don't like quiet failures for stuff that should be constant and never fail (except for updates and sheet changes). If needed, `Lumina.GameData.Get(Subrow)Sheet` is an alternative to `Lumina.Excel.ExcelModule.Get(Subrow)Sheet` that throws only when the schema/struct type is invalid.


KazWolfe (changes requested):
~~Please remove the CS bump. If you need it, let us know so that we can upstream the proper changes.~~ Completed, thank you.

As noted as well, this'll be blocked until Lumina updates itself and/or we decide that we want to maintain our own repo/project separately.

KazWolfe:
Also, still somewhere on the todo list:

- [ ] Write migration guides for changing concepts.
  - [ ] Special callout to sheet extensions/additions/calculated methods.
  - [ ] Any relevant pitfalls that may occur as part of this new mechanism.
- [ ] Figure out how to rapidly update `EXDSheets` to get changes upstreamed in a timely manner.
- [ ] Deal with any other strange dev use-cases that might come up.

KazWolfe (approved):

PR #2031, opened 2024-08-27
Fix GameNetwork sig

Soreepeong:
Previous signature caught onto another call instruction's non-first byte, resulting in interpreting mid-bytecode as a call bytecode, which apparently jumped to some random bytecode of a pixel shader, making HBAOPlus initialization fail.

image


Soreepeong:
Possible alternate sig, using that the distance between caller and callee in question are <0x1000000: `E8 ?? ?? ?? 00 48 83 C4 28 C3 32 C0 48 83 C4 28 C3 CC`

Haselnussbomber:
Yeah, that's the caller sig. IDA didn't give me a good one, so I sigged the function start instead. 🤷‍♂️
Both are fine. 🙂

PR #2033, opened 2024-08-27
Update Lumina and use MacroStringParser

Soreepeong:
No longer requires hook and main thread requirements on compiling macro strings.

Needs lookup table fixing on Lumina; using reflection to fix for the time being.

### DIFF
--- a/Dalamud/Interface/ImGuiSeStringRenderer/Internal/SeStringRenderer.cs
+++ b/Dalamud/Interface/ImGuiSeStringRenderer/Internal/SeStringRenderer.cs
@@ -1,4 +1,3 @@
-using System.Buffers;
 using System.Collections.Generic;
 using System.Numerics;
 using System.Runtime.InteropServices;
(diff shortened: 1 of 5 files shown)


KazWolfe:
These changes are API breaking - am I to assume that the things they're replacing are still new enough to not see widespread adoption?

If not, we may need to defer this to api11.

Dalamud/Interface/ImGuiSeStringRenderer/Internal/SeStringRenderer.cs
```diff
@@ -91,13 +72,23 @@ private SeStringRenderer(DataManager dm, TargetSigScanner sigScanner)
...
+             !.GetValue(null) as sbyte[])
+            .AsSpan();
+        fixedTable.Fill(-1);
+        fixedTable['_'] = fixedTable['\''] = -2;
+        for (var i = '0'; i <= '9'; i++)
+            fixedTable[i] = (sbyte)(i - '0');
+        for (var i = 'A'; i <= 'F'; i++)
+            fixedTable[i] = (sbyte)(10 + (i - 'A'));
+        for (var i = 'a'; i <= 'f'; i++)
+            fixedTable[i] = (sbyte)(10 + (i - 'a'));
```
KazWolfe, line 91: How quickly can this fix be upstreamed?

Haselnussbomber: This is included in Lumina [v4.2.1](https://github.com/NotAdam/Lumina/releases/tag/4.2.1), so this PR needs to be updated.

Soreepeong:
Updated to 4.2.1 and it seems that extension methods will not be referred to if the original class has the correspondingly signatured functions from future compilation; restored `this` modifier to those deprecated functions.

Soreepeong:
The one breaking change in Lumina is from a base class that is not supposed to be implemented by plugins anyway, so it should be fine.

KazWolfe (approved):

PR #2035, opened 2024-08-27
Change type of JobFlags to ulong

Infiziert90:
The packet uses ulong for all the job flags, but dalamud's current enum implements them still as int, which breaks any math around Pictormancer as it has overflown the int value and turned negative


KazWolfe (approved):
Approved as breaking change; we can go bother the plugins in question to update if we need to. Don't think too many things touch PF.

PR #2037, opened 2024-08-30
Improve Position/Rotation documentation

ALymphocyte:


KazWolfe:
Hi!

First off, thank you for proposing this change. While I do agree with the intent of these doc updates, as-is they don't really add too much that isn't already easily discernible. The information contained here may be better suited as a page on `dalamud.dev` explaining the systems in question, as well as possibly providing an overview of game vs map coordinates (though this is already documented in `MapUtil` to an extent).

As such, I'm going to go ahead and close this PR pending a more thorough rewrite of the GameObject docs, which we'll eventually get to. If you'd like to try your hand on expanding things more, feel free to reopen this PR!

PR #2038, opened 2024-09-03
Update Addon Inspector

ItsBexy:
This updates the data window's Addon Inspector with lots of new features and functionality.

- Features from Caraxi's fork of UiDebug have been incorporated, such as the Element Selector UI and address search.
- Any addon or node can now pop out into its own window.
- Revised the visual style of node field/property information.
- Color values are now visually displayed.
- Any nodes or components that are referenced by fields within the addon will now show that field name in the inspector.
- Added editors for nodes, allowing complete control over most of their properties.
- Improved texture display for Image nodes (and Image node variant types). The active part of the texture is now highlighted, and the boundaries of other parts can be shown via mouseover.
- Highlighting of node bounds onscreen is now more accurate, factoring in rotation (including when using the Element Selector).
- Display of animation timelines has been revamped, showing a table of keyframes for each animation.

A standalone SamplePlugin-based version is available here: [https://github.com/ItsBexy/UiDebug2](https://github.com/ItsBexy/UiDebug2)

### DIFF
--- a/Dalamud/Interface/Internal/UiDebug2/Browsing/AddonTree.FieldNames.cs
+++ b/Dalamud/Interface/Internal/UiDebug2/Browsing/AddonTree.FieldNames.cs
@@ -16,8 +16,10 @@ public unsafe partial class AddonTree
 {
     private static readonly Dictionary<string, Type?> AddonTypeDict = [];
 
+    private static readonly Assembly? ClientStructs = AppDomain.CurrentDomain.GetAssemblies().SingleOrDefault(static a => a.GetName().Name == "FFXIVClientStructs");
+
     /// <summary>
-    /// Gets or sets a collection of names corresponding to pointers documented within the addon struct.
+    /// Gets or sets a collection of names for field offsets that have been documented in FFXIVClientStructs.
     /// </summary>
     internal Dictionary<nint, List<string>> FieldNames { get; set; } = [];
 
@@ -28,27 +30,24 @@ public unsafe partial class AddonTree
             return null;
         }
 
-        if (AddonTypeDict.TryAdd(this.AddonName, null))
+        if (AddonTypeDict.TryAdd(this.AddonName, null) && ClientStructs != null)
         {
-            foreach (var a in AppDomain.CurrentDomain.GetAssemblies())
+            try
             {
-                try
+                foreach (var t in from t in ClientStructs.GetTypes()
+                                  where t.IsPublic
+                                  let xivAddonAttr = (Addon?)t.GetCustomAttribute(typeof(Addon), false)
+                                  where xivAddonAttr != null
+                                  where xivAddonAttr.AddonIdentifiers.Contains(this.AddonName)
+                                  select t)
                 {
-                    foreach (var t in from t in a.GetTypes()
-                                      where t.IsPublic
-                                      let xivAddonAttr = (Addon?)t.GetCustomAttribute(typeof(Addon), false)
-                                      where xivAddonAttr != null
-                                      where xivAddonAttr.AddonIdentifiers.Contains(this.AddonName)
-                                      select t)
-                    {
-                        AddonTypeDict[this.AddonName] = t;
-                        break;
-                    }
+                    AddonTypeDict[this.AddonName] = t;
+                    break;
                 }
-                catch
-                {
-                    // ignored
-                }
+            }
+            catch
+            {
+                // ignored
             }
         }
 
@@ -108,7 +107,7 @@ public unsafe partial class AddonTree
         {
             try
             {
-                if (this.FieldNames.TryAdd(fieldAddr, [..path!, name]) && fieldType.DeclaringType == baseType)
+                if (this.FieldNames.TryAdd(fieldAddr, [..path, name]) && fieldType.DeclaringType == baseType)
                 {
                     this.PopulateFieldNames(field.GetValue(obj), fieldAddr, [..path, name]);
                 }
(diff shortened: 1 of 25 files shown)


Dalamud/Interface/Internal/Windows/Data/Widgets/AddonInspectorWidget.cs
```diff
@@ -1,11 +1,11 @@
-﻿namespace Dalamud.Interface.Internal.Windows.Data.Widgets;
+namespace Dalamud.Interface.Internal.Windows.Data.Widgets;
 
 /// <summary>
 /// Widget for displaying addon inspector.
 /// </summary>
 internal class AddonInspectorWidget : IDataWindowWidget
 {
-    private UiDebug? addonInspector;
+    private UiDebug2.UiDebug2? addonInspector;
```
KazWolfe, line 8: Let's keep both inspectors around for a little bit. Once we can get feedback from devs overall, we can delete the legacy `UiDebug`.

KazWolfe (changes requested):
See requested change - let's just keep both UI debug inspectors around so we can get feedback on the new one.

Infiziert90:
The ImRaii usage is a good beginning, but it should also be used with `using` statements and not manually disposing in the same function

Dalamud/Interface/Internal/UiDebug2/Browsing/TimelineTree.cs
```diff
@@ -0,0 +1,386 @@
...
+        }
+
+        return str;
+    }
+
+    private void PrintAnimation(AtkTimelineAnimation animation, int a, bool isActive, nint address)
+    {
+        var columns = this.BuildColumns(animation);
+
+        var col = ImRaii.PushColor(ImGuiCol.Text, isActive ? new Vector4(1, 0.65F, 0.4F, 1) : new(1));
```
goaaats, line 313: You're doing this all over so this is more of a general comment: ImRaii doesn't really do anything for safety if you're not also using it with a using block or statement. Shouldn't be very difficult to fix!

goaaats (changes requested):
Looks good. Some small stuff and Kaz's comment and I think this is good to go!

Dalamud/Interface/Internal/UiDebug2/Browsing/TimelineTree.cs
Haselnussbomber: It's also possible to pass the `isActive` to the condition parameter, so it only pushes the color when isActive is true:
```cs
using var col = ImRaii.PushColor(ImGuiCol.Text, new Vector4(1, 0.65F, 0.4F, 1), isActive);
```
In case isActive is false, it will not push the color and continue to use the current active one.
Just wanted to note this, but I didn't check context if it's necessary here to push a white color or not. 🙂

Dalamud/Interface/Internal/UiDebug2/Browsing/AddonTree.FieldNames.cs
```diff
@@ -0,0 +1,196 @@
...
+    private object? GetAddonObj(AtkUnitBase* addon)
+    {
+        if (addon == null)
+        {
+            return null;
+        }
+
+        if (AddonTypeDict.TryAdd(this.AddonName, null))
+        {
+            foreach (var a in AppDomain.CurrentDomain.GetAssemblies())
```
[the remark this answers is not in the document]
ItsBexy: Narrowed the search to the `FFXIVClientStructs` assembly, since you're right, that's the only place we want to look.

Dalamud/Interface/Internal/UiDebug2/Browsing/NodeTree.Text.cs
```diff
@@ -0,0 +1,122 @@
...
+    {
+        if (isEditorOpen)
+        {
+            return;
+        }
+
+        ImGui.TextColored(new(1), "Text:");
+        ImGui.SameLine();
+
+#pragma warning disable
```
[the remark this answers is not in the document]
ItsBexy: This was for `SeStringRenderer`, clarified now.

ItsBexy:
Implemented most of the requested changes. Thanks all for the heads-up on ImRaii, I hadn't used it before so I wasn't quite familiar with the proper usage.

Dalamud/Interface/Internal/UiDebug2/Utility/Gui.cs
```diff
@@ -0,0 +1,169 @@
...
+        using (ImRaii.PushColor(Text, new Vector4(0.6f, 0.6f, 0.6f, 1)))
+        {
+            ImGuiHelpers.ClickToCopyText(text, textCopy);
+        }
+
+        if (ImGui.IsItemHovered())
+        {
+            using (ImRaii.Tooltip())
+            {
+                using var f = ImRaii.PushFont(UiBuilder.IconFont);
```
Infiziert90, line 134: Scoping would make it instantly visible at what points this gets popped

Dalamud/Interface/Internal/UiDebug2/Browsing/AddonTree.cs
```diff
@@ -0,0 +1,242 @@
...
+    internal void Draw()
+    {
+        if (!this.ValidateAddon(out var addon))
+        {
+            return;
+        }
+
+        var isVisible = addon->IsVisible;
+
+        ImGui.Text($"{this.AddonName}");
```
Infiziert90, line 117: `TextUnformatted()` should always be used over `Text()`

The provided string formatting that `Text()` offers isn't useable in C#, but will instantly lead to crashes the moment it encounters a % symbol

PR #2055, opened 2024-10-13
fix: testing plugins not detected if testing API not present

Aireil:


Aireil:
Intended behavior, repos must use `TestingDalamudApiLevel` to distribute testing versions.

PR #2059, opened 2024-10-26
Use RCX to resolve entrypoint

Soreepeong:
RCX contains entry point; RDX contains PEB. While this behavior is not contractual, in reality commercial packers just assume that this is a fact, which probably means that we can rely on this behavior too.


goaaats:
Can you add a quick note about why we're doing this instead now? Is it just cleaner/less code? What is missing to merge it?

Soreepeong:
It's less code with a very minor performance benefit, tho the main purpose of this PR was to offer a test build for someone guessing that memory lookup is the culprit in slow game startup performance. If something about entrypoint load doesn't work in future this can be attempted again at that time.

PR #2060, opened 2024-10-29
Use hooks instead of lifecycle for NamePlateGui

nebel:
This changes `NamePlateGui` to hook `AddonNamePlate.OnRequestedUpdate` directly instead of using `AddonLifecycle`.

The reason for this is because `AddonNamePlate.Show` can call `AddonNamePlate.OnRequestedUpdate` directly in certain conditions, bypassing the callsite hook used by `AddonLifecycle`. As a result, some `OnNamePlateUpdate` and `OnDataUpdate` events are missed (especially at the end of certain loading screens) which can result in a plugin being unable to respond to an update and leaving the plates in an invalid state.

Knowing this can happen, I also added `OnPostNamePlateUpdate` and `OnPostDataUpdate` events for use by plugins which currently use an `AddonEvent.PostRequestedUpdate` lifecycle hook (and as a result, also miss the same calls we missed before). These plugins can then switch to their lifecycle hooks to these new events which are more reliable.

There is no change to the API of existing events, so this should be backwards compatible.


nebel:
Force-pushed to fix merge conflict.

PR #2063, opened 2024-11-05
Safely call events in DalamudPluginInterface

Haselnussbomber:
This changes the event handler to safely call `ActivePluginsChanged` and `OnLocalizationChanged` in `DalamudPluginInterface` to prevent exceptions in plugins stopping the execution of other code.

I noticed this on the API11 branch, when I changed the language in the Dalamud Settings and it wouldn't save because one of my plugins threw an exception in its `LanguageChanged` event handler due to Lumina changes.

### DIFF
--- a/Dalamud/Plugin/DalamudPluginInterface.cs
+++ b/Dalamud/Plugin/DalamudPluginInterface.cs
@@ -526,13 +526,40 @@ internal sealed class DalamudPluginInterface : IDalamudPluginInterface, IDisposa
     /// <param name="affectedThisPlugin">If this plugin was affected by the change.</param>
     internal void NotifyActivePluginsChanged(PluginListInvalidationKind kind, bool affectedThisPlugin)
     {
-        this.ActivePluginsChanged?.Invoke(kind, affectedThisPlugin);
+        if (this.ActivePluginsChanged is { } callback)
+        {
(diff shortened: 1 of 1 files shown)


Dalamud/Plugin/DalamudPluginInterface.cs
```diff
@@ -526,13 +526,40 @@ public void Dispose()
     /// <param name="affectedThisPlugin">If this plugin was affected by the change.</param>
     internal void NotifyActivePluginsChanged(PluginListInvalidationKind kind, bool affectedThisPlugin)
     {
-        this.ActivePluginsChanged?.Invoke(kind, affectedThisPlugin);
+        if (this.ActivePluginsChanged is { } callback)
```
goaaats, line 529: Would the extensions we already have work for this?
https://github.com/goatcorp/Dalamud/blob/5f62c703bff4137a1f887553fc1e0bd932d6dc6e/Dalamud/Utility/EventHandlerExtensions.cs#L22

goaaats (changes requested):

Dalamud/Plugin/DalamudPluginInterface.cs
Haselnussbomber: Nope, I tried to use them, but they don't work with the custom delegate we define in IClientState.

goaaats (approved):

PR #2065, opened 2024-11-05
[api11] Change ClientState event logic

Haselnussbomber:
With this PR, ClientState now completely relies on hooks instead of the Framework.Update loop or Conditions.

For the **Login** event, `ProcessPacketPlayerSetup` is hooked.
This is where the game sets a bool `IsLoggedIn` in `AgentLobby` to `true` and sets the ContentId to UI modules.
Listeners are invoked after the original function, so that the player data is already set.

For the **Logout** event, `LogoutCallbackInterface_OnLogout` is hooked.
This is where the game sets a bool `IsLoggedIn` in `AgentLobby` to `false` and resets UI modules.
The hook provides us with a `type` and `code` of the logout, and is also called when the game gets disconnected, but is seemingly not called when the game could not connect to the lobby server (since it isn't really a logout).
I'm unsure what exactly `type` is, but from what I've seen type 2 makes the AtkDialogue box show up.
When closing the game or normally logging out, the type is 1 and code is 10000.
When losing connection, the type is 2 and code is 90002.
Listeners are invoked before the original function, so that we can handle the logout before the game does.

**EnterPvP** and **LeavePvP** invokers have been moved into the `SetupTerritoryTypeDetour`.
`GameMain.IsInPvPArea` is exactly the same thing as reading `IsPvpZone` from the TerritoryType sheet, but because the TerritoryType row isn't loaded in GameMain when the detour is called, we have to read it off the sheet ourselves.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -14,6 +14,7 @@ using Dalamud.Logging.Internal;
 using Dalamud.Plugin.Services;
 using Dalamud.Utility;
 
+using FFXIVClientStructs.FFXIV.Application.Network;
 using FFXIVClientStructs.FFXIV.Client.Game;
 using FFXIVClientStructs.FFXIV.Client.Game.Event;
 using FFXIVClientStructs.FFXIV.Client.UI;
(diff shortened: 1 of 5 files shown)


KazWolfe (approved):

Dalamud/Game/ClientState/ClientState.cs
```diff
@@ -165,23 +166,46 @@ void IInternalDisposableService.DisposeService()
...
+        {
+            var isPvP = rowRef.Value.IsPvpZone;
+            if (isPvP != this.IsPvP)
+            {
+                this.IsPvP = isPvP;
+                this.IsPvPExcludingDen = this.IsPvP && this.TerritoryType != 250;
+
+                if (this.IsPvP)
+                {
+                    Log.Debug("EnterPvP");
```
KazWolfe, line 192: These debug statements ideally would be more descriptive.

PR #2070, opened 2024-11-11
Update ImGuiComponents & ImGuiHelpers

ItsBexy:
Took some helper functions created for `UiDebug2`, and incorporated them into `ImGuiComponents` and `ImGuiHelpers` instead

- `IconButton()` and its various overloads now include an optional size parameter
- `IconButtonSelect()` has been added, allowing a row or grid of IconButtons to serve as a radio-like input
- `HelpMarker()` now includes an optional color parameter
- `ClickToCopyText()` now includes an optional color parameter, and includes the `FontAwesome.Copy` icon in the tooltip.
- Updated methods to use ImRaii where applicable

These changes are intended not to break any existing calls in plugins or within Dalamud itself (knock on wood).

### DIFF
--- a/Dalamud/Interface/Components/ImGuiComponents.DisabledButton.cs
+++ b/Dalamud/Interface/Components/ImGuiComponents.DisabledButton.cs
@@ -1,5 +1,7 @@
 using System.Numerics;
 
+using Dalamud.Interface.Utility.Raii;
+
 using ImGuiNET;
 
 namespace Dalamud.Interface.Components;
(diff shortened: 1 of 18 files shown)


Dalamud/Interface/Components/ImGuiComponents.DisabledButton.cs
```diff
@@ -21,15 +23,17 @@ public static partial class ImGuiComponents
...
+        bool button;
+        using (ImRaii.PushFont(UiBuilder.IconFont))
+        {
+            var text = icon.ToIconString();
+            if (id.HasValue)
+            {
+                text = $"{text}##{id}";
+            }
+
+            button = DisabledButton(text, defaultColor, activeColor, hoveredColor, alphaMult);
```
Infiziert90, line 35: You can just return here, no need to store it just for return

Return ends the scope automatically

Dalamud/Interface/Utility/ImGuiHelpers.cs
```diff
@@ -167,17 +167,41 @@ public static List<Vector4> DefaultColorPalette(int swatchCount = 32)
...
+
+                ImGui.SameLine();
+                ImGui.TextUnformatted($"{textCopy}");
+            }
         }
 
-        if (ImGui.IsItemClicked()) ImGui.SetClipboardText($"{textCopy}");
+        if (ImGui.IsItemClicked())
+        {
+            ImGui.SetClipboardText($"{textCopy}");
```
Infiziert90, line 203: Could just directly use `textCopy`, it gets null checked above and is a string already

ItsBexy:
I've also put together a [GameIconButton](https://github.com/goatcorp/Dalamud/blob/dae0074d87976e00cdc9f0c3b75b9902f8c30bd6/Dalamud/Interface/Components/ImGuiComponents.GameIconButton.cs) feature, but I haven't included it in this PR because it relies on `ITextureProvider` and I'm not yet sure what would be the least clunky way to address that dependency.

.gitignore
```diff
@@ -327,4 +327,5 @@ ASALocalRun/
 *.nvuser
 
 # MFractors (Xamarin productivity tool) working folder 
-.mfractor/
\ No newline at end of file
+.mfractor/
+/Dalamud/Interface/Components/ImGuiComponents.GameIconButton.cs
```
KazWolfe, line 331: Eh?

PR #2073, opened 2024-11-13
UIDebug2 Fixes

ItsBexy:
- Quick-fix for retrieving UnitLists (switch back to using AtkStage instead of UIModule)
- Update references to AddonAttribute

These two changes got my standalone plugin version of the tool working again, so hopefully it'll work here too.


Haselnussbomber:
You should be working on the api11 branch, but also AddonAttribute was already updated and the offset of RaptureAtkUnitManager was already updated in CS. :)

ItsBexy:
> You should be working on the api11 branch, but also AddonAttribute was already updated and the offset of RaptureAtkUnitManager was already updated in CS. :)

Eurgh, thanks. Very Sleepy.

PR #2074, opened 2024-11-13
Extend Icon Browser range

ItsBexy:
Most status effect icons have been moved to IDs above 200,000, which the Icon Browser was hardcoded not to go past. Extending to 220,000 for now.


MidoriKami:
Huh interesting, I only set the limit to prevent it from trying to load excess things, its since been improved, if we suspect more things might be added, there's likely no harm in extending to 500,000

MidoriKami:
As far as I know only `// Remove range 170,000 -> 180,000 by default, this specific range causes exceptions.` causes any actual issues to try and load.

KazWolfe:
Implemented in fd6d52d33b94b769bf61c506b436c5e5c0e5e159.

PR #2083, opened 2024-11-15
Update JobGauge

Taurenkey:
Corrects the job gauge offset.


Infiziert90:
This should be fixed in CS instead :)

PR #2091, opened 2024-11-16
Update JobGauge

Taurenkey:
This resolves an issue where if a gauge was to try and be cached before the memory address is properly updated would result in caching an incorrect address. Most notably would happen if something tried to access it during login.


KazWolfe:
Superseded by #2094.

PR #2093, opened 2024-11-16
Revert ItemHover/ItemOut hook removal

Haselnussbomber:
I also updated the delegates. 😊


Haselnussbomber:
I replaced old delegates and signatures with delegates and vfunc addresses from CS, but it depends on:
- https://github.com/aers/FFXIVClientStructs/pull/1200

KazWolfe (approved):

PR #2097, opened 2024-11-16
Add InventoryWidget

Haselnussbomber:
I thought it may be useful to have a widget showing Inventory containers and their items.

Screenshot1

Collectables:
Screenshot2

High Quality:
Screenshot3


Haselnussbomber:
Updated to use `GameInventoryItem.GetReadOnlySpanOfInventory`.
Also now displays Stains and Materia if the item supports them.

Screenshot

KazWolfe (approved):
Approved if only I want a reason to check item IDs without Simple Tweaks.

Haselnussbomber:
It's not about you, it's about bringing popular debug features from a plugin into Dalamud, because said plugin might be down when we want to fix CS and Dalamud. But thanks for approving.

KazWolfe:
That was indeed the point.

PR #2106, opened 2024-11-18
Implement Print methods that work with Lumina SeString/ReadOnlySeString

Infiziert90:
Adds the same functions that already exist for `string` and `Dalamud.SeString` but instead with a `ReadOnlySpan<byte>` overload which all Lumina version implicitly convert to.

The NBSP code had to be replaced with a byte[] implementation.

### DIFF
--- a/Dalamud/Game/Gui/ChatGui.cs
+++ b/Dalamud/Game/Gui/ChatGui.cs
@@ -112,7 +112,14 @@ internal sealed unsafe class ChatGui : IInternalDisposableService, IChatGui
     /// <inheritdoc/>
     public void Print(XivChatEntry chat)
     {
-        this.chatQueue.Enqueue(new XivChatEntryRaw(chat.Name.Encode(), chat.Message.Encode(), chat.Type, chat.Timestamp, chat.Silent));
+        this.chatQueue.Enqueue(new XivChatEntryRaw
+        {
+            Type = chat.Type,
+            Timestamp = chat.Timestamp,
+            Message = chat.Message.Encode(),
+            Name = chat.Name.Encode(),
+            Silent = chat.Silent,
+        });
     }
 
     #region DalamudSeString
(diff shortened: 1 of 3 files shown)


Dalamud/Game/Text/XivChatEntry.cs
```diff
@@ -32,3 +32,34 @@ public sealed class XivChatEntry
...
+
+    /// <summary>
+    /// Gets the sender name.
+    /// </summary>
+    public byte[] Name { get; set; } = [];
+
+    /// <summary>
+    /// Gets the message.
+    /// </summary>
+    public byte[] Message { get; set; } = [];
```
KazWolfe, line 59: Are these the only two methods for chat entries, and are they always guaranteed to be SeStrings? If so, you can add these fields to `XivChatEntry` as `NameBytes` and `MessageBytes`, then have `Name`/`Message` read from/write to accordingly.

Dalamud/Game/Gui/ChatGui.cs
```diff
@@ -246,30 +288,32 @@ private void PrintTagged(string message, XivChatType channel, string? tag, ushor
 
         this.Print(new XivChatEntry
         {
-            Message = builder.AddText(message).Build(),
+            Message = builder.Build().Append(message),
```
KazWolfe, line 291: If this is just a `SeString`, why bother invoking the builder?

KazWolfe (changes requested):

Dalamud/Game/Gui/ChatGui.cs
Infiziert90: I haven't touched this method, the git diff is just confusing it with the removed `string` one

Dalamud/Game/Gui/ChatGui.cs
```diff
@@ -137,6 +143,24 @@ public void PrintError(SeString message, string? messageTag = null, ushort? tagC
         this.PrintTagged(message, XivChatType.Urgent, messageTag, tagColor);
     }
 
+    #endregion
+
+    #region LuminaSeString
```
KazWolfe, line 148: Not sure regions are necessary for such small segments.

(non-blocking).

KazWolfe (approved):

PR #2108, opened 2024-11-19
fix: Don't use a hook for ClientState#TerritoryType

KazWolfe:
- We can just read this from the game data.

Added as its own method because unsafe getters are ugly af.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -105,7 +105,7 @@ internal sealed class ClientState : IInternalDisposableService, IClientState
     public ClientLanguage ClientLanguage { get; }
 
     /// <inheritdoc/>
-    public ushort TerritoryType { get; private set; }
+    public ushort TerritoryType => this.GetCurrentTerritoryTypeId();
 
     /// <inheritdoc/>
     public unsafe uint MapId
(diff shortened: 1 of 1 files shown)


goaaats:
Why specifically, is it just because this isn't set when injecting while the game is running? I usually don't like this type of change as it makes it harder to reason about what order things are done in. i.e. the field we are reading from might be set after the hook is fired, so now subscribers that read from the field in the event won't have the correct territory.

goaaats:
I think that is a deal-breaker until we decouple that and also have two IDs for current and pending territory, if that is even necessary. I doubt anyone has ever needed it and our API might not need to support it, just seems confusing to people.

Haselnussbomber:
I just ran into this by accident in my plugin, actually.
Instead of using `ClientState.IsPvP`, I was using `GameMain.IsInPvPArea()`, and it was set to false a lot later than the event `ClientState.TerritoryChanged` was fired. Just to demonstrate this, I'll post the logs here:
```
03:39:00.085 | OnTerritoryChanged: 250
03:39:00.897 | IsInPvPArea: True
03:39:12.290 | OnTerritoryChanged: 339
03:39:12.291 | OnLeavePvP
03:39:14.493 | IsInPvPArea: False
```

wolfcomp:
IMO if this has a more correct return type than the hook this should be the defacto way to get the territory type.

goaaats:
More correct is not always == a good API, especially if it breaks existing assumptions.

wolfcomp:
This isn't about following API properties since we are explicitly having to deal with how SE codes the systems so if this follows better to how the system works in SE code this is the correct way and what is currently breaks the assumptions about the system.

PR #2113, opened 2024-11-21
fix: Only regenerate the title screen version string PostRequestedUpdate

KazWolfe:
draft pr, i still want this to update as plugins are enabled/disabled


KazWolfe:
Closed in favor of #2117, since that actually updates every frame.

PR #2116, opened 2024-11-23
Attempt to fix file path with special characters

Kobi-Blade:
This pull request aims to address the issue of special characters in file paths by implementing encoding. This change is related to a bug reported [here](https://github.com/goatcorp/FFXIVQuickLauncher/issues/1036).

**Testing:** Unfortunately, I do not have an account with special characters to properly test this fix.

### DIFF
--- a/Dalamud/Interface/Internal/InterfaceManager.cs
+++ b/Dalamud/Interface/Internal/InterfaceManager.cs
@@ -597,7 +597,11 @@ internal partial class InterfaceManager : IInternalDisposableService
             var startInfo = Service<Dalamud>.Get().StartInfo;
             var configuration = Service<DalamudConfiguration>.Get();
 
-            var iniFileInfo = new FileInfo(Path.Combine(Path.GetDirectoryName(startInfo.ConfigurationPath)!, "dalamudUI.ini"));
+            string filePath = Path.Combine(Path.GetDirectoryName(startInfo.ConfigurationPath)!, "dalamudUI.ini");
+            byte[] encodedFilePathBytes = Encoding.UTF8.GetBytes(filePath);
+            string encodedFilePath = Encoding.UTF8.GetString(encodedFilePathBytes);
+
+            var iniFileInfo = new FileInfo(encodedFilePath);
 
             try
             {


KazWolfe:
I'm not sure what this change is intended to do? You're encoding a string to bytes and then immediately un-encoding it.

Running the same code in LINQPad results in `filePath` being equal to `encodedFilePath`:

image

Can you please let me know what the intent here is?

KazWolfe:
Can you please provide a test case where this works as expected and the input is not equal to the output? I was unable to find such a case.

KazWolfe:
> Your current implementation fails when users have a simple accent mark (´) in their names, which is quite common in Latin names such as Cédric.

Same test harness:

```cs
const string TestHarness = "Cédric";

string filePath = Path.Combine(Path.GetTempPath(), TestHarness, "foo.txt");
byte[] encodedFilePathBytes = Encoding.UTF8.GetBytes(filePath);
string encodedFilePath = Encoding.UTF8.GetString(encodedFilePathBytes);

Console.WriteLine($"FilePath={filePath}  EncodedFilePath={encodedFilePath}");
Console.WriteLine($"EQUALITY: {filePath == encodedFilePath}");

var plainFileInfo = new FileInfo(filePath);
Console.WriteLine($"File size using plain: {plainFileInfo.Length}");

var encodedFileInfo = new FileInfo(encodedFilePath);
Console.WriteLine($"File size using Encoded: {encodedFileInfo.Length}");
```

```
FilePath=C:\Users\Kaz\AppData\Local\Temp\Cédric\foo.txt  EncodedFilePath=C:\Users\Kaz\AppData\Local\Temp\Cédric\foo.txt
EQUALITY: True
File size using plain: 11
File size using Encoded: 11
```

wolfcomp:
From looking at this I don't think the issue is in `Dalamud.dll` but in `Dalamud.Boot.dll` or `Dalamud.Injector.exe` that malforms the path string

KazWolfe:
Thank you for the contribution! We managed to chase down the problem to ImGuiScene, and I've opened PR #2122 that supersedes this one for the fix.

PR #2117, opened 2024-11-23
Avoid re-creating title version string every frame

Soreepeong:
* Added `PluginManager.LoadedPluginCount` which will count the number of loaded plugin without making a copy of the plugin list.
* Made TitleScreenMenuWindow.OnVersionStringDraw` update the title version text addon only if number of loaded plugin changes or the text is overwritten by the game.


KazWolfe (approved):
Accepting in favor of #2113 until we have an event for enabled plugin counts.

PR #2122, opened 2024-11-24
fix: Properly handle UTF-8 characters in ImGUI paths

KazWolfe:
- Fixes a bug where users with special characters in their filenames would not be able to save `dalamudUI.ini`.
- Throw a special warning if `dalamudUI.ini` doesn't exist, as it's not an error case.

Fixes goatcorp/FFXIVQuickLauncher#1036.

### DIFF
--- a/Dalamud/Interface/Internal/InterfaceManager.cs
+++ b/Dalamud/Interface/Internal/InterfaceManager.cs
@@ -604,9 +604,16 @@ internal partial class InterfaceManager : IInternalDisposableService
                 if (iniFileInfo.Length > 1200000)
                 {
                     Log.Warning("dalamudUI.ini was over 1mb, deleting");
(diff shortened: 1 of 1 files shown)


Dalamud/Interface/Internal/InterfaceManager.cs
```diff
@@ -617,6 +624,8 @@ private unsafe RawDX11Scene InitScene(IDXGISwapChain* swapChain)
             newScene.ImGuiIniPath = iniFileInfo.FullName;
             newScene.OnBuildUI += this.Display;
             newScene.OnNewInputFrame += this.OnNewInputFrame;
+            
+            Log.Warning($"IMGUI INI PATH: {newScene.ImGuiIniPath}");
```
goaaats, line 628: This meant to be here?

goaaats (dismissed):

goaaats (approved):

goaaats (approved):

PR #2126, opened 2024-11-27
[master] Update ClientStructs

bleatbot:


wolfcomp:
Complains about a `HouseId` property that was changed from `uint` to `long` this is explicitly done with knowledge about how the system worked as the `uint` was truncating half the information that is stored in said field.

goaaats (approved):

PR #2129, opened 2024-11-28
Use object table index lookup for NamePlateGui GameObject resolution

nebel:
Sometimes, for example when exiting GPose, nameplate object info pointers may briefly end up pointing to invalidated object table entries. These objects appear valid in some ways but are not, and can have partially corrupted data/vtables. With this change, rather than wrapping the (assumed valid) object with `CreateObjectReference`, we use the object's `ObjectIndex` to look it up in `ObjectTable`. This way the object table pointer is checked for null and we correctly refuse to return an invalidated `GameObject`.

This should fix a known crash in the Player Tags plugin.


goaaats (approved):

Pilzinsel64:
Nice! 👍

PR #2131, opened 2024-11-29
Only set BGM in About tab when client is idle

### LINKED ISSUE #2130: BGM set to None when opening About screen in an instance
Steps to reproduce:
1. Go into an instance
2. Open Dalamud Settings
3. Click on About
4. go out of About

Expected result: BGM goes back to playing the instance's current song
Actual result: No BGM playing.

Happend to me in FRU, getting to a new phase fixes the issue as new BGM starts playing.
I've also tested it in Emanation.

Haselnussbomber:
OnOpen and OnClose were changed to make sure the client is idle before setting the background music.
OnClose additionally checks if the background music was set in OnOpen to begin with.

**Currently untested**, but the change seems simple enough.
I'm just not sure if SetBgm has other side-effects. 🤷‍♂️

Fixes #2130

### DIFF
--- a/Dalamud/Interface/Internal/Windows/Settings/Tabs/SettingsTabAbout.cs
+++ b/Dalamud/Interface/Internal/Windows/Settings/Tabs/SettingsTabAbout.cs
@@ -1,9 +1,11 @@
-﻿using System.Diagnostics;
+using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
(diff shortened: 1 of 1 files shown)


goaaats:
Is there a reason you closed this? Is the ticket not valid?

Haselnussbomber:
Better version incoming. :)

PR #2132, opened 2024-11-30
Use MiniGame BGM scene for About tab

### LINKED ISSUE #2130: BGM set to None when opening About screen in an instance

Haselnussbomber:
I reversed the BGMSystem a bit and now we can make proper use of the BGM scenes. 👍

This PR depends on:
- https://github.com/aers/FFXIVClientStructs/pull/1234

To elaborate further, there are 12 BGM scenes the game can use:
- 0 = Event
- 1 = Battle
- 2 = MiniGame (RhythmAction, TurnBreak)
- 3 = Content
- 4 = GFate
- 5 = Duel
- 6 = Mount
- 7 = Unknown, no xrefs
- 8 = Unknown, via packet (near PlayerState stuff)
- 9 = Wedding
- 10 = Town
- 11 = Territory

The signature for `SetGlobalBgm` resolved to the setter for the BGM scene Content, which is why the instance music doesn't come back after the About tab changed it.
I changed `GameGui.SetBgm` to default to the MiniGame scene, as it is not used very often and mini games can easily be restarted.
I also added a `GameGui.ResetBgm` function that takes care of resetting the scene for us, instead of using an invalid BGM id.
Finally, I replaced the internal `GameGui.IsOnTitleScreen` function with a `IsInLobby` function that simply checks if the `CurrentUIScene` in `AtkModule` equals to `LobbyMain`, so we don't have to check for all the possible lobby addons.

Fixes #2130

### DIFF
--- a/Dalamud/Game/Gui/GameGui.cs
+++ b/Dalamud/Game/Gui/GameGui.cs
@@ -9,6 +9,7 @@ using Dalamud.Logging.Internal;
 using Dalamud.Plugin.Services;
 using Dalamud.Utility;
 
+using FFXIVClientStructs.FFXIV.Client.Game;
 using FFXIVClientStructs.FFXIV.Client.Game.Control;
 using FFXIVClientStructs.FFXIV.Client.Graphics.Kernel;
 using FFXIVClientStructs.FFXIV.Client.System.String;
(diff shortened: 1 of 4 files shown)


wolfcomp:
CS bump only requirement now for this

goaaats (approved):

PR #2142, opened 2024-12-10
Add timezone offset to system time in crash log

wolfcomp:


DalamudCrashHandler/DalamudCrashHandler.cpp
```diff
@@ -941,7 +941,8 @@ int main() {
             log << std::format(L"Dump at: {}", dumpPath.wstring()) << std::endl;
         else
             log << std::format(L"Dump error: {}", dumpError) << std::endl;
-        log << L"System Time: " << std::chrono::system_clock::now() << std::endl;
+        log << L"System Time: " << std::chrono::system_clock::now();
```
KazWolfe, line 944: Isn't this always UTC?

NotNite: Yes.

wolfcomp: No. It is not UTC. It is your local PCs time

NotNite: This isn't true, it's UTC going off of file creation dates of old logs (I live in America/New_York):
image

wolfcomp: Taken straight from the doc: `The system_clock's time value can be internally adjusted at any time by the operating system, for example due to NTP synchronization or the user changing the system's clock.`

wolfcomp: It will not always be UTC based due to that last cause.

wolfcomp: If you want a stable UTC clock its `std::chrono::utc_clock`

KazWolfe: From [cppreference](https://en.cppreference.com/w/cpp/chrono/system_clock), emphasis mine:

> The system_clock's time value can be internally adjusted at any time by the operating system, for example due to NTP synchronization or the user changing the system's clock. **Daylight Saving Time and time zone changes, however, do not affect it since it is based on the [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time) time-zone.**

However, I will also note that only C++ 20 or newer seem to require that `system_clock` return UTC - and Windows in particular is very well known for doing Clock Shenanigans. MSFT's [own documentation](https://learn.microsoft.com/en-us/cpp/standard-library/system-clock-structure) does not specify a time zone here.

@wolfcomp have you personally seen a use case where this is *not* UTC, e.g. on your own machine?

wolfcomp: ok must be the C++ 20 change then since when I looked it up it wasn't stated that it was required to be UTC and was using the local time on the computer for windows

KazWolfe: n.b. I'm not opposed to this change out of principle; including time zone information tends to be better than not doing it. I do, however, want to be sure we're making a well-defined decision here.

I do think, at a minimum, we should only get `std::chrono::system_clock::now()` once and carry that through for as long as we can. Past that, it would be ideal if we could only deal with formatting things once, though I'm not certain how much of a savings that would be.

wolfcomp: Seeing how it is UTC with C++ 20 and on ward most likely just going to append `<< L"+0000"` instead of the other one

KazWolfe: That's... what we're doing? I'm confused.

Re offset, I think it might still be better to calculate it just in case some system misbehaves, though I'm not sure how that would happen. 

Let's save `std::chrono::system_clock::now()` to a variable and format with `{0:%F} {0:%T} {0:%Ez}`. Does that work for everyone?

KazWolfe (approved):

PR #2143, opened 2024-12-11
Return nullable on XivChatType.GetDetails

Haselnussbomber:
Some `XivChatType` do not have a `XivChatTypeInfoAttribute` on them, so `XivChatType.GetDetails()` should return a nullable.


Haselnussbomber:
Wait. That's not a breaking change??

KazWolfe (approved):

PR #2147, opened 2024-12-17
UI cleanup and changes to ImGuiHelpers

Infiziert90:
- Removes most uses of formatted imgui function

ImGuiHelpers Class:
- Changes the existing `SafeTextWrapped` implementation from .Replace() to what ImGui does internally
- Adds two new "Safe" function `SafeTextDisabled` and `SafeTextColored`


KazWolfe:
Yet another argument for just forking ImGui.NET and locking our API surface to `*Unformatted` only.

Infiziert90:
Added `SafeSetTooltip`, this replaces the last formatted function which was missing.
Important to note is that this function does not 100% replicate the original behaviour, as ImRaii doesn't offer support for discarding previous Tooltips, nor do we have access to this flag in our bindings currently.

PR #2155, opened 2024-12-24
Build adjustments for code annotation

nathanctech:
This PR overrides Nuke's log handling and removes the problematic formatting the default analyzers were unable to process as well as improving the build flow to reduce duplicate warnings (and thus annotations). Building and testing have merged into a single step, and warnings are suppressed from the test rebuild since they're just refired from the initial build process anyway.

In addition, I adjusted the workflow so pushes to an existing pull request won't trigger both the push and PR builds to occur in some cases which also contributed to the warning duplication. In the end, file changes now have code-level annotations!


.github/workflows/main.yml
```diff
@@ -1,5 +1,9 @@
 name: Build Dalamud
-on: [push, pull_request, workflow_dispatch]
+on:
+  push:
+  pull_request:
+    types: [opened]
```
wolfcomp, line 5: This will cause subsequent builds to not trigger for PRs. This should not be the case and the default types for `pull_request` is `opened`, `syncronize`, `reopened` which all handle different cases with pushes to the pull request for commits and if a pull request is opened again after closing. Best is to revert this change as you want it to build for each time something happens to the commit tree for the PR.

nathanctech: Hmm, when testing on my fork the `push` trigger was firing whenever I made a commit to my draft PR. That could've been something on my end - I'll revert this portion though just to be safe.

build/DalamudBuild.cs
```diff
@@ -159,20 +163,32 @@ public class DalamudBuild : NukeBuild
                 .SetConfiguration(Configuration));
         });
 
+    Target SetCustomLogging => _ => _
```
KazWolfe, line 166: Please call this `SetCILogging`, since this should only really be set in the CI process.

KazWolfe (changes requested):
GitHub Actions sets an environment variable (`CI`) to `true`. please only ensure these changes are only active when `CI == true`.

.github/workflows/main.yml
```diff
@@ -32,10 +33,8 @@ jobs:
           ($env:REPO_NAME) >> VERSION
           ($env:BRANCH) >> VERSION
           ($env:COMMIT) >> VERSION
-      - name: Build Dalamud
+      - name: Build and Test Dalamud
```
KazWolfe, line 36: Is it not possible to keep these as separate steps?

nathanctech: Based on the runner behavior it wasn't, since the test was unable to find the previously built binaries and was triggering a rebuild, causing log duplication when the warnings appeared again. When combined this way, Nuke can handle all the steps at once instead of being called twice. I can look further into it though if needed.

build/DalamudBuild.cs
nathanctech: Will make this change!

.github/workflows/main.yml
KazWolfe: This specific one isn't a blocking thing, but I can see why this'd be annoying for local dev work especially as Nuke is somewhat required for certain workflows.

Ideally we'd use them as separate steps though. I'd say see if it's doable but don't worry _too_ much if it's not (unless we change our minds/goat says we need it split).

nathanctech:
Well...git's quite something and re-commited a bunch of things despite my best efforts - anyway:

It now checks for the CI variable before doing any changes to the logging, so local builds will function as normal. I also added `--skip-tests` as a possible parameter in case those need to be skipped for some reason. You can technically still run tests without everything else locally by passing `test` as with the previous process and it will build as usual, too.

.github/workflows/main.yml
nathanctech: Mentioned on discord, but I couldn't figure out how to make Nuke handle the artifact upload/download portion required for the splitting to work, so I added `--skip-tests` to mimic the compile-only functionality from before, and `build.ps1 test` still works as usual.

wolfcomp:
Maybe have skip test being the default on non CI builds?

nathanctech:
> Maybe have skip test being the default on non CI builds?

Also an option, yeah. Then you could invoke it manually which would make sense since you're already having to do that with the current process. I can make that change if that's the way to go.

KazWolfe:
The other side is to have a `ci-run` target that does compile+test. That might be cleaner in general, though we'd need to probably look more into Nuke and how it works.

nathanctech:
> The other side is to have a `ci-run` target that does compile+test. That might be cleaner in general, though we'd need to probably look more into Nuke and how it works.

From my research, the `compile` and `test` parameters being passed in the current process simply make Nuke trigger a given Target specification which then sets off the chain using the dependencies given. So, we could indeed have something like `build.ps1 ci` and have it trigger the build+test, with default being build only and untouched logging.

nathanctech:
Okay. I scratched the skip-tests thing entirely. Now, the build process runs `build.ps1 ci` which triggers the test after building. When done locally with `build.ps1 compile`, it only runs the build step as expected.

goaaats (approved):

goaaats:
Thank you! Really appreciate it, and sorry for taking so long on the merge!

PR #2160, opened 2024-12-30
Update Easing.cs to not have breaking changes mid API

wolfcomp:
Undoes breaking changes done with commit `1aada98`


wolfcomp:
Fixed with commit `b36bdb2`

sersorrel:
@wolfcomp if you don't surround commit hashes with backticks, they get linkified automatically :) e.g.: b36bdb2

PR #2176, opened 2025-01-27
feat: Allow feedback to unofficial endpoints

KazWolfe:
Requires goatcorp/dalamud-bugbait#10.

As an aside, we really need to redo manifests, they're getting painful.

Intended behavior:
* Plugins that have `AcceptFeedback == false` do not display feedback buttons.
* Mainline plugins will *always* send feedback to the Bugbait endpoint.
  * Bugbait will be responsible for forwarding the feedback to the webhook URL.
* Custom plugins will accept feedback only if they have a feedback URL set.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
@@ -110,6 +110,7 @@ internal class PluginInstallerWindow : Window, IDisposable
     private bool feedbackModalIncludeException = false;
     private IPluginManifest? feedbackPlugin = null;
     private bool feedbackIsTesting = false;
+    private bool feedbackIsCustomRepo = false;
 
     private int updatePluginCount = 0;
     private List<PluginUpdateStatus>? updatedPlugins;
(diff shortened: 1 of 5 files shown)


Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
```diff
@@ -2444,10 +2451,13 @@ private void DrawAvailablePlugin(RemotePluginManifest manifest, int index)
                 ImGuiHelpers.ScaledDummy(3);
             }
 
-            if (!manifest.SourceRepo.IsThirdParty && manifest.AcceptsFeedback && !isOutdated)
+            var feedbackEligible = manifest.AcceptsFeedback &&
+                                   (!manifest.SourceRepo.IsThirdParty || manifest.FeedbackWebhook != null);
```
Haselnussbomber, line 2455: Pretty sure `!manifest.SourceRepo.IsThirdParty` should be removed here since you moved it to `DrawSendFeedbackButton`.

KazWolfe: This still controls whether the button is visible or not. The draw method only tracks it to place into state (which then goes to bugbait).

This code is confusing in general, and we need to redo manifests at some point.

Haselnussbomber: Oh, nevermind. I understand now. Sorry!

Haselnussbomber:
Yeah. Well, I deleted my comments since I didn't see your response. All good tho.

KazWolfe:
Closing as stale, this doesn't seem like something we really care much about.

reiichi001:
I still think we should support this. 

Especially for certain cases where a developer is not on the discord and has no other means of accepting feedback.

PR #2182, opened 2025-02-26
Remove hooks when disposed to allow plugins to GC properly

Critical-Impact:
References to plugin are being held by HookInfo(as it stores the delegate for the hook). This does what the plugin stats screen does but in the background and on a timer. Have removed the same functionality from the plugin stats screen as it's no longer needed.


KazWolfe:
Ref #1803.

PR #2188, opened 2025-03-05
Add SeStringEvaluator service

Haselnussbomber:
This PR adds the public-facing **[ISeStringEvaluator](https://github.com/Haselnussbomber/Dalamud/blob/sestringevaluator/Dalamud/Plugin/Services/ISeStringEvaluator.cs)** service.

This service allows macro payloads and their expressions to be evaluated.

- Supports local and global parameters.
- Supports a large majority of macro codes, including almost all fixed string variations. (See [SeStringEvaluator.cs](https://github.com/Haselnussbomber/Dalamud/blob/sestringevaluator/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs#L183))
- Supports an optional language override to be specified, which is used instead of Dalamuds EffectiveLanguage, for internal sheet lookups inside macros.
- The API is currently designed for ease of use rather than extensibility. For example, `SeStringEvaluator.EvaluateFromAddon(8519, [info.Ward, info.Plot])`. The `SeStringParameter`s have inplicit casts to make it as easy as possible. It does not provide an option to handle specific macros differently (you have to pre/post-process the string for that), but that is not really an issue in my opinion.

Notes:

- Originally written by Kizer. I heavily expanded on it. Some code was commented out when I made the API simpler, but I left them in as a reminder where to place them, in case someone decides to change it in the future.
- There are sadly no unit tests.
- Does not provide feature parity to the games evaluator - there are some differences/advantages for us, for example in processing macros recursively and mixing macro and fixed string evaluation, but also we're missing a couple features/macro codes.
- Can be used independently of the games loaded excel language, which is a huge plus.
- `ISeStringEvaluator` is marked experimental for now.
- This does not support Dalamuds SeString type at all, as they don't have expressions. If needed, they can just as easily be cast to ReadOnlySeString. That said, certain Dalamud SeString payloads are fundamentally wrong (looking at you, AutoTranslatePayload - which is the fixed macro), and it might be better to use the SeStringEvaluator internally for those.
- A self test `SeStringEvaluatorAgingStep` was added, which is not very exhaustive, but covers processing of macros, local and global parameters, and generating expected results.

A xldata widget has been added to create SeStrings and preview their evaluated results:

Preview

It's possible to export the entries as macro string and to load strings from excel sheets:

Preview Import

A second tab displays a list of global parameters, their type, value and, if known, their description.

---

In order to not clutter the SeStringEvaluator class with other things, I moved some code into their own, internal services: the NounProcessor and the SheetRedirectResolver service.

Regarding the **NounProcessor**:

- The noun system is used in a couple of sheets (see below). This service helps in processing them, for example an Item name.
- It supports handling of different language-specific grammatical features based on the Attributive sheet for the following placeholders:
  - `[t]`: Article or grammatical gender (EN: the, DE: der, die, das).
  - `[n]`: The amount of an item (number).
  - `[a]`: Declension
  - `[p]` Plural
- Used in evaluation of the `janoun`, `ennoun`, `denoun` and `frnoun` macro codes in the SeStringEvaluator.
- C# versions of the games `Component::Text::Localize::NounEn/NounDe/NounFr/NounJa` classes.
	- NounCh (Chinese) and NounKo (Korean) were omitted, since we can't use them and I cant validate them anyway.

Notes:

- There are some variables from porting over the decompiled code to C# that I didn't rename yet, because I am lazy and got tired of it. Please forgive me.
- I spent hours upon hours to reverse this and make sense of it.
- Might need to get looked over by someone who studied languages - especially for French.
- A self test `NounProcessorAgingStep` was added to do spot tests.
- A xldata widget was added in order to help with the noun payloads:

<details>
<summary>Preview</summary>

Please note that the self-test export button is only available in local debug builds.

Japanese:

Preview Japanese

English:

Preview English

German:

Preview German

French:

Preview French

</details>

<details>
<summary>A list of the (to me) currently known sheets that use the noun system</summary>

- Aetheryte
- BNpcName
- BeastTribe
- DeepDungeonEquipment
- DeepDungeonItem
- DeepDungeonMagicStone
- DeepDungeonDemiclone
- ENpcResident
- EObjName
- EurekaAetherItem
- EventItem
- GCRankGridaniaFemaleText
- GCRankGridaniaMaleText
- GCRankLimsaFemaleText
- GCRankLimsaMaleText
- GCRankUldahFemaleText
- GCRankUldahMaleText
- GatheringPointName
- Glasses
- GlassesStyle
- HousingPreset
- Item
- MJIName
- Mount
- Ornament
- TripleTriadCard

</details>

---

Regarding the **SheetRedirectResolver**:

This service resolves some special (partially non-existant) sheet names, most importantly ActStr and ObjStr.

It is just a C# version of the games `Client::UI::Misc::RaptureTextModule.ResolveSheetRedirect` function.

A self test `SheetRedirectResolverAgingStep` was added to check if the results are matching the games resolver.

---

Notes:

- I coded this for my plugins and just ported it to be a Dalamud service. Technically this could be part of Lumina, but given that we fetch global parameters from the game and that it would be a *lot* of work to make the necessary API changes, I'll rather PR this to Dalamud.
- Originally, I was thinking of providing functions in a TextProvider service to get singular names from a variety of sheets, like in my [TextService](https://github.com/Haselnussbomber/HaselCommon/blob/e7d598576676b425ef4fbbf367512c820345c5a1/HaselCommon/Services/TextService.cs#L120-L235), but I felt this was a) too many functions, b) unnecessary since it's just caching and not even evaluating (except for the Fate sheet), and c) doesn't provide the capabilities of the NounProcessor (pluralization etc.).
- I added a couple Util functions and made them publicly available.

I'm sure it will need some updates in the future.

I hope this will be useful for a lot of plugins and I can't wait to no longer see those noun placeholders in names!

As a final word to plugin devs reading this: please strip [soft hyphens](https://en.wikipedia.org/wiki/Soft_hyphen) out of names.
They are basically everywhere in German sheets to indicate a word break opportunity and ImGui will render them as `-` because it can't handle it correctly. I've added a `string.StripSoftHypen()` extension function for that. If you use `ImGuiHelpers.SeStringWrapped()` you're already covered - the great Kizer made sure it renders correctly.
Lets make things work correctly in at least the 4 client languages we can support! :)

### DIFF
--- a/Dalamud/Dalamud.csproj
+++ b/Dalamud/Dalamud.csproj
@@ -10,7 +10,7 @@
 
     <PropertyGroup Label="Feature">
         <Description>XIV Launcher addon framework</Description>
-        <DalamudVersion>11.0.7.0</DalamudVersion>
+        <DalamudVersion>11.0.8.0</DalamudVersion>
         <AssemblyVersion>$(DalamudVersion)</AssemblyVersion>
         <Version>$(DalamudVersion)</Version>
         <FileVersion>$(DalamudVersion)</FileVersion>
--- a/Dalamud/Dalamud.csproj
+++ b/Dalamud/Dalamud.csproj
@@ -4,7 +4,7 @@
         <TargetFramework>net8.0-windows</TargetFramework>
         <PlatformTarget>x64</PlatformTarget>
         <Platforms>x64</Platforms>
-        <LangVersion>12.0</LangVersion>
+        <LangVersion>13.0</LangVersion>
         <EnableWindowsTargeting>True</EnableWindowsTargeting>
     </PropertyGroup>
 
--- a/Dalamud/Game/ActionKind.cs
+++ b/Dalamud/Game/ActionKind.cs
@@ -1,0 +1,89 @@
+namespace Dalamud.Game;
+
+/// <summary>
+/// Enum describing possible action kinds.
+/// </summary>
+public enum ActionKind
+{
+    /// <summary>
+    /// A Trait.
+    /// </summary>
+    Trait = 0,
+
+    /// <summary>
+    /// An Action.
+    /// </summary>
+    Action = 1,
+
+    /// <summary>
+    /// A usable Item.
+    /// </summary>
+    Item = 2, // does not work?
+
+    /// <summary>
+    /// A usable EventItem.
+    /// </summary>
+    EventItem = 3, // does not work?
+
+    /// <summary>
+    /// An EventAction.
+    /// </summary>
+    EventAction = 4,
+
+    /// <summary>
+    /// A GeneralAction.
+    /// </summary>
+    GeneralAction = 5,
+
+    /// <summary>
+    /// A BuddyAction.
+    /// </summary>
+    BuddyAction = 6,
+
+    /// <summary>
+    /// A MainCommand.
+    /// </summary>
+    MainCommand = 7,
+
+    /// <summary>
+    /// A Companion.
+    /// </summary>
+    Companion = 8, // unresolved?!
+
+    /// <summary>
+    /// A CraftAction.
+    /// </summary>
+    CraftAction = 9,
+
+    /// <summary>
+    /// An Action (again).
+    /// </summary>
+    Action2 = 10, // what's the difference?
+
+    /// <summary>
+    /// A PetAction.
+    /// </summary>
+    PetAction = 11,
+
+    /// <summary>
+    /// A CompanyAction.
+    /// </summary>
+    CompanyAction = 12,
+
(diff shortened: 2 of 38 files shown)


Haselnussbomber:
Alright, I think I've added enough. 😃 No more! 😆

KazWolfe:
Merging in to api12 as this is `[Experimental]` tagged, and I think it'd be good to have more people kick the tires of this to see what happens/doesn't happen.

This is obviously a very big (net-new) service, and it has a lot of potential edge cases, so I think we're going to keep that tag for A While.

PR #2192, opened 2025-03-12
Read .text section bytes from file instead of process memory

Nukoooo:
This PR addresses issue where .text section can be modified by other tools before Dalamud loads, which may cause some plugins to fail to find signatures

### DIFF
--- a/Dalamud/Game/SigScanner.cs
+++ b/Dalamud/Game/SigScanner.cs
@@ -496,13 +496,17 @@ public class SigScanner : IDisposable, ISigScanner
                     this.TextSectionOffset = Marshal.ReadInt32(sectionCursor, 12);
                     this.TextSectionSize = Marshal.ReadInt32(sectionCursor, 8);
 
-                    var pointerToRawData = Marshal.ReadInt32(sectionCursor, 20);
+                    if (this.IsCopy)
+                    {
+                        var pointerToRawData = Marshal.ReadInt32(sectionCursor, 20);
 
-                    Marshal.Copy(
-                                 fileBytes.AsSpan(pointerToRawData, this.TextSectionSize).ToArray(),
-                                 0,
-                                 this.moduleCopyPtr + (nint)this.TextSectionOffset,
-                                 this.TextSectionSize);
+                        Marshal.Copy(
+                                     fileBytes.AsSpan(pointerToRawData, this.TextSectionSize).ToArray(),
+                                     0,
+                                     this.moduleCopyPtr + (nint)this.TextSectionOffset,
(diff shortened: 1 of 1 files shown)


Dalamud/Game/SigScanner.cs
```diff
@@ -463,6 +462,8 @@ private static int[] BuildBadCharTable(byte[] needle, bool[] mask)
 
     private void SetupSearchSpace(ProcessModule module)
     {
+        var fileBytes = File.ReadAllBytes(module.FileName);
```
KazWolfe, line 465: This will _always_ read all file bytes into memory, no?

If so (and we're committing to always reading `.text` from disk), is there any downside to reading *everything* from disk instead? cc @Soreepeong.

Soreepeong: It works better because it almost rarely is intentional to sigscan over thunk-overwritten memory data, and as long as offsets match, even using different hooking libraries to nest hooks should work fine.

Though, since it will read the file every time a class instance is constructed, probably better to do that once and share the read-only memory across different instances.

Also, we probably have IMAGE_DOS_HEADER and IMAGE_NT_HEADERS64 importable, so it might be better to use those structs instead of Marshal.Read... with byte offsets.

Dalamud/Game/SigScanner.cs
```diff
@@ -51,12 +52,16 @@ public SigScanner(ProcessModule module, bool doCopy = false, FileInfo? cacheFile
...
         this.IsCopy = doCopy;
 
-        // Limit the search space to .text section.
-        this.SetupSearchSpace(module);
-
         if (this.IsCopy)
+        {
             this.SetupCopiedSegments();
 
+            fileBytes ??= File.ReadAllBytes(module.FileName);
```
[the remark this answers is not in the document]
Nukoooo: > Do this only if module.BaseAddress == Process.GetCurrentProcess().MainModule!.BaseAddress, if module.Equals(Process.GetCurrentProcess().MainModule) does not work.

I don't think that is necessary in this case as every construction of ``SigScanner`` (including ``TargetSigScanner``) is ``Process.GetCurrentProcess().MainModule!``

> you need to map file addresses to virtual addresses and apply relocations according to the address of the actually loaded module rather than the read buffer.

Do we have any cases that need to do this? I did the same thing in [CounterStrikeSharp](https://github.com/roflmuffin/CounterStrikeSharp/pull/414) and no one has reported issues yet

goaaats:
Sorry, missed this one when looking for PRs that are still pointed to api12. I will merge this manually.

PR #2197, opened 2025-03-19
SMNGauge: Fix typo, update CS

zbee:
This PR has some simple updates for `SMNGauge` that get it back in line with what CS has for the gauge.

- [X] Update note about how `Attunement` should be used
- [X] Add `AttunementCount` and `AttunementType`, which use `Attunement` as a bitfield
- [X] Fix `AttunmentTimerRemaining` (missing `e`) typo

### DIFF
--- a/Dalamud/Game/ClientState/JobGauge/Types/SMNGauge.cs
+++ b/Dalamud/Game/ClientState/JobGauge/Types/SMNGauge.cs
@@ -52,15 +52,15 @@ public unsafe class SMNGauge : JobGaugeBase<SummonerGauge>
     public byte Attunement => this.Struct->Attunement;
 
     /// <summary>
-    /// The count of attunement cost resource available.
+    /// Gets the count of attunement cost resource available.
     /// </summary>
     public byte AttunementCount => this.Struct->AttunementCount;
 
     /// <summary>
-    /// The type of attunement available.
+    /// Gets the type of attunement available.
     /// Use the summon attuned accessors instead.
     /// </summary>
-    public byte AttunementType => this.Struct->AttunementType;
+    public SummonAttunement AttunementType => (SummonAttunement)this.Struct->AttunementType;
 
     /// <summary>
     /// Gets the current aether flags.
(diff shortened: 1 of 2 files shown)


Dalamud/Game/ClientState/JobGauge/Types/SMNGauge.cs
```diff
@@ -40,10 +43,25 @@ internal SMNGauge(IntPtr address)
...
+    /// <summary>
+    /// The count of attunement cost resource available.
+    /// </summary>
+    public byte AttunementCount => this.Struct->AttunementCount;
+
+    /// <summary>
+    /// The type of attunement available.
+    /// Use the summon attuned accessors instead.
+    /// </summary>
+    public byte AttunementType => this.Struct->AttunementType;
```
KazWolfe, line 63: Would it make sense to declare an `enum` of attunement(s) and use that here? Or are attunements backed by a Lumina table?

zbee: I'm not sure if it is backed by a table or not, I can't readily find one at least.

An enum may well make sense.

At least in the plugin I am working on we were using `AttunementType == 1`, `2`, and `3`, but just replaced those with the summon accessors `IsIfritAttuned`, `IsTitanAttuned`, and `IsGarudaAttuned` successfully, and `0` seems to be "unattuned".

So I believe I have the data for an enum, if we want to do that, just don't know if it's really needed or not 🤷🏻

KazWolfe: Yeah, let's include an enum then. If someone needs the raw value, we can always cast to `byte`. I'd _technically_ prefer an enum over accessors (flexibility), but I'm not going to remove API surface just for that.

zbee: About to commit an enum then 👍🏻

KazWolfe (approved):

PR #2199, opened 2025-03-23
Update DalamudCommands.cs

Etupa:
Added a /xltogglemultimonitor command for toggling multimonitor.

Using Dalamud local build encounters Assertion Failure with Umbra plugin enabled. Otherwise works fine.

### DIFF
--- a/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
+++ b/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
@@ -1066,8 +1066,8 @@ internal class SeStringEvaluator : IServiceType, ISeStringEvaluator
         if (!enu.MoveNext() || !this.TryResolveUInt(in context, enu.Current, out var placeNameIdInt))
             return false;
 
-        var instance = packedIds >> 0x10;
-        var mapId = packedIds & 0xFF;
+        var instance = packedIds >> 16;
+        var mapId = packedIds & 0xFFFF;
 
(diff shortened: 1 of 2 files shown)


KazWolfe (dismissed):
Approving code changes, they seem fine as is. Going to ping @goaaats though to see if this is something we want to add.

Can you please explain the intent behind adding this command (i.e. what are you doing that would benefit from it)?

Dalamud/Interface/Internal/DalamudCommands.cs
```diff
@@ -152,6 +152,11 @@ private DalamudCommands(CommandManager commandManager)
                 "DalamudCopyLogHelp",
                 "Copy the dalamud.log file to your clipboard."),
         });
+        // Add the new command handler for toggling multi-monitor option
+        commandManager.AddHandler("/xlmulti", new CommandInfo(this.OnToggleMultiMonitorCommand)
```
goaaats, line 156: Could you change this to `/xltogglemultimonitor` to be a bit more descriptive?

goaaats (changes requested):

Dalamud/Interface/Internal/DalamudCommands.cs
Etupa: ofc ! was a bit lazy while testing

Etupa (commented):
applied requested change to /command name

PR #2202, opened 2025-03-24
fix: typo in JobFlags

status102:


KazWolfe:
Thanks for the PR! Given we're going to change this enum in the future and this is a doc-only change, we're not going to take action on this PR. When we next update JobFlags for Beastmaster, we'll fix this in-tree and close this PR.

PR #2223, opened 2025-03-28
Remove MateriaSlotCount check

Haselnussbomber:


KazWolfe (approved):
Approving as-is to patch the bug. If necessary, we can rollback to raw spans later.

As an aside, api13 should ideally deprecate these and return just `IReadOnlyList<Materia>` or similar, need to consider what's best for API surface there. I absolutely don't think there's a reason to keep returning `ReadOnlySpan<>` here.

PR #2230, opened 2025-03-31
feat: Set XIVLauncher AppID

KazWolfe:
- Sets the AppUserModelID of the FFXIV process to XIVLauncher, allowing the game to stack into the XL process.

Depends on FFXIVQuickLauncher#1675 to work properly, otherwise this'll just use the Squirrel setup (not bad, but not great).

### DIFF
--- a/Dalamud/Interface/Internal/InterfaceManager.cs
+++ b/Dalamud/Interface/Internal/InterfaceManager.cs
@@ -10,6 +10,7 @@ using System.Threading.Tasks;
 
 using CheapLoc;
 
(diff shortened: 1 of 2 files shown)


goaaats:
Does this have any side-effects, like resetting some settings that people have already set? Thinking audio levels, etc. I don't know what it's used for beyond what you wrote.

KazWolfe:
To my knowledge, no. That said, Windows likes being weird with this sorta stuff. This isn't particularly well documented.

It would behoove us to have the XL side working, though.

PR #2232, opened 2025-04-01
Safer AddonEventManager node removal

Haselnussbomber:
Services are disposed asynchronously ([here](https://github.com/Haselnussbomber/Dalamud/blob/remove-event-fix/Dalamud/IoC/Internal/ServiceScope.cs#L141)), which means when AddonEventManager wants to remove events from addons it might not do that on the Framework thread.

As posted in Discord, I crashed while it was looking for the event during removal:

Screenshot

I simply wrapped the executing code in a `Service<Framework>.Get().RunOnFrameworkThread(() => ... )`.

As an additional safety measure I added a check if UldManager is not unloaded yet, though the NodeCount is being set to 0 shortly afterwards anyway.

I hope this prevents the rare occurrence of a crash there.


Dalamud/Game/Addon/Events/AddonEventManager.cs
```diff
@@ -230,8 +230,11 @@ void IInternalDisposableService.DisposeService()
         {
             this.eventManagerService.ResetCursor();
         }
-        
-        this.eventManagerService.RemovePluginEventController(this.plugin.EffectiveWorkingPluginId);
+
+        Service<Framework>.Get().RunOnFrameworkThread(() =>
+        {
+            this.eventManagerService.RemovePluginEventController(this.plugin.EffectiveWorkingPluginId);
```
KazWolfe, line 236: T/N: would it be safe to `.Wait()` execution of this to prevent the post-dispose `this` call?

Haselnussbomber: I didn't think of that. Should be fine, since it's run async. I added it.

PR #2241, opened 2025-04-08
feat: Deprecate IGameNetwork

KazWolfe:
This PR is a **proposal** to remove the `IGameNetwork` service from Dalamud. With changes in Patch 7.2 (namely, additional packet obfuscation), this service no longer really provides useful data in some cases. In general, Dalamud has encouraged developers move away from opcode management in favor of hook(s) in the appropriate locations, and this solidifies that decision.

This service may still be used by certain plugins. As such, we are seeking input from relevant community members who may still have some use for these APIs about the deprecation path here.

Identified Uses:

* [PingPlugin](https://github.com/karashiiro/PingPlugin/blob/30e45e3186af72460056ab5d7d81f6ec73a8eaab/PingPlugin/PingTrackers/PacketPingTracker.cs#L59), @karashiiro
* [StanleyParableXIV](https://github.com/rekyuu/StanleyParableXiv/blob/36ae9d6fad2ce5aa92be73edcf53e42b753431c5/StanleyParableXiv/Events/PvpEvent.cs#L318), @rekyuu
* [DeepDungeonTracker](https://github.com/marconsou/deep-dungeon-tracker/blob/16089d0a45076c0b7a3c917a4c5b43da49d07985/DeepDungeonTracker/Data/Data.cs#L245), @marconsou 

At this time, the internal service is not planned to be deprecated, as it's still used for the debug window. This feature is still deemed beneficial as it can be used to identify whether something does (or doesn't) send a network request.

### DIFF
--- a/Dalamud/Game/Network/Internal/NetworkHandlers.cs
+++ b/Dalamud/Game/Network/Internal/NetworkHandlers.cs
@@ -56,7 +56,6 @@ internal unsafe class NetworkHandlers : IInternalDisposableService
 
     [ServiceManager.ServiceConstructor]
     private NetworkHandlers(
-        GameNetwork gameNetwork,
         TargetSigScanner sigScanner,
         HappyHttpClient happyHttpClient)
     {
--- a/Dalamud/Plugin/Services/IGameNetwork.cs
+++ b/Dalamud/Plugin/Services/IGameNetwork.cs
@@ -1,10 +1,13 @@
 ﻿using Dalamud.Game.Network;
+using Dalamud.Utility;
 
 namespace Dalamud.Plugin.Services;
 
 /// <summary>
 /// This class handles interacting with game network events.
 /// </summary>
+[Obsolete("To be removed with API13.")]
+[Api13ToDo("Remove")]
 public interface IGameNetwork
 {
     // TODO(v9): we shouldn't be passing pointers to the actual data here


KazWolfe:
In response to a Discord comment:

> If we can actually provide a hook that outputs good, deobfuscated packets that might be valuable [...]

I'm of the opinion that there's no longer any valid use case for this service. Even were we to move the hooks to provide deobfuscated information, there's very little value to providing such an API surface. 

There is a possibility that it may still be beneficial to have a singular hook interacting with the relevant function(s), but in practice this seems minimal at best. Nearly every plugin that would use this service can be better served hooking elsewhere, leaving the general-purpose methods untouched. 

While the singular-hook argument also means that events can be managed in a cleaner fashion and we can prevent plugins from bringing down the entire stack, it's still ultimately a plugin developer's responsibility to test their code and make sure things don't crash. This method already passed a `nint` to game data directly, meaning plugins that wanted to tamper with things already had that ability - multiple hooks doesn't really change this too much.

rekyuu:
I only really use `IGameNetwork` for PvP countdown and completion events, so it wouldn't be a huge loss but it would be a bit sad. Not sure if those events could be re-implemented somehow.

https://github.com/rekyuu/StanleyParableXiv/blob/main/StanleyParableXiv/Events/PvpEvent.cs#L334

PingPlugin and DeepDungeonTracker are obviously way more important though.

I agree that moving these sorts of things to an actual API would be much better and it would make things like this less confusing to read. I worry that it would also mean that there will prob be a lot of very minute, specific cases that someone might want implemented in the API, but would have to wait behind a PR/Dalamud update. Maybe that's not a bad thing, but it makes "should this specific packet be implemented into Dalamud" a possibly subjective thing.

Critical-Impact:
Anything currently using the network could track down one of the functions further the call tree from ProcessZonePacketDown using IDA/Ghidra and hook that instead. You've still gotta decode the network packets depending on how far down you go but there's definitely alternatives.

karashiiro:
PingPlugin has three different ping-tracking implementations and the one that relies on this is the last fallback - given that the packet handler has been giving obfuscated data for a few months now and it hasn't been an issue, I'll consider that as it not being very useful in the first place. I'll likely just remove that opcode-based ping tracker altogether and not replace it (that would leave the `COM` ping and `IpHlpApi` trackers).

KazWolfe:
Resolved per cb2932296838ab1cd972eea21a4b25332c37ed31.

PR #2247, opened 2025-04-15
Modernize CommandManager, prep for API13, add new experimental console-backed API

goaaats:


Dalamud/Game/Command/CommandManager.cs
```diff
@@ -41,12 +41,31 @@ private CommandManager(Dalamud dalamud)
             (nint)ShellCommands.MemberFunctionPointers.TryInvokeDebugCommand,
             this.OnTryInvokeDebugCommand);
         this.tryInvokeDebugCommandHook.Enable();
-
-        this.console.Invoke += this.ConsoleOnInvoke;
     }
 
     /// <inheritdoc/>
-    public ReadOnlyDictionary<string, IReadOnlyCommandInfo> Commands => new(this.commandMap);
+    [Api13ToDo("Make this sensible. Don't use exposed API for internal structures.")]
```
KazWolfe, line 47: API note: some people can/do use this, we may need to create an interface to expose commands. (See: SimpleTweaks)

Dalamud/Game/Command/CommandManager.cs
```diff
@@ -135,66 +184,88 @@ public bool AddHandler(string command, CommandInfo info, string loaderAssemblyNa
...
-        }
-
-        return true;
+        return this.AddHandler(command, info, null);
     }
 
     /// <inheritdoc/>
-    public bool RemoveHandler(string command)
+    [Experimental("Dalamud001")]
+    public bool AddCommand(string commandName, string helpMessage, Delegate func, bool showInHelp = true, int displayOrder = -1)
```
KazWolfe, line 192: Honestly not sure here re a bare `Delegate`. Maybe a new class like a `CommandExecutionContext` would be better? Would let us pass `args` as a bare string, "parsed" args, and other context info (e.g. if we're in a macro).

KazWolfe (commented):
We're moving from InternalName to GUID in a bunch of places. Is there a reason for this vs just passing in a reference to that loaded plugin?

Dalamud/Game/Command/CommandManager.cs
goaaats: Yeah, we definitely still want this. The intent isn't to remove it, but to replace it with something nicer.

PR #2248, opened 2025-04-15
Add MinimumDalamudVersion validate at install, update and load

goaaats:


Dalamud/Plugin/Internal/Types/PluginManifest.cs
```diff
@@ -75,6 +75,10 @@ internal record PluginManifest : IPluginManifest
     [JsonConverter(typeof(GameVersionConverter))]
     public GameVersion? ApplicableVersion { get; init; } = GameVersion.Any;
 
+    /// <inheritdoc/>
+    [JsonProperty]
+    public Version? MinimumDalamudVersion { get; init; }
```
KazWolfe, line 80: How is this going to react with stg tagging? I guess we're going to be using AssemblyVersions, but those often aren't really bumped for stg and all.

In general, how big of a case is "user is not running updated Dalamud" anyways, given we force an auto-update?

goaaats: I don't think solving this for stg is desirable or worthwhile. We would have to add additional versioning information and it feels like there is no real use-case.

The specific case that prompted us to add this was a plugin causing crashes with a newly merged version of CS, while also causing crashes with the old version of CS when compiled against the updated one. The developer had to add a version check for Dalamud themselves, since otherwise, people in-game were to auto-update to the updated version that would be incompatible with their Dalamud version and crash.

KazWolfe (approved):

PR #2251, opened 2025-04-16
Read .text section bytes from file instead of process memory

Nukoooo:
Reopen of #2192


KazWolfe:
Paging @Soreepeong to sign off on this, please. See notes from prior PR #2192.

Soreepeong:
While I'd rather rework on SigScanner, the functionality implemented here looks fine; I can try to make it a bit cleaner at a later time.

On relocations: copies inside SigScanner is not exposed elsewhere, and people should not be including data that will be modified by relocations in their signatures so that's fine.

PR #2253, opened 2025-04-20
Mark main thread in Framework.Destroy detour

Haselnussbomber:
When we're in the detour of Framework.Destroy we are on the main thread.
Functions guarded with `ThreadSafety.AssertMainThread();` should work in this context.

### DIFF
--- a/Dalamud/Game/Framework.cs
+++ b/Dalamud/Game/Framework.cs
@@ -474,6 +474,8 @@ internal sealed class Framework : IInternalDisposableService, IFramework
 
     private unsafe bool HandleFrameworkDestroy(CSFramework* thisPtr)
     {
+        ThreadSafety.MarkMainThread();
+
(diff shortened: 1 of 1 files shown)


KazWolfe:
Eh? Would this not already be the case? It was my understanding that `MarkMainThread` sets a ThreadStatic to true, meaning the rest of that thread's lifetime is considered the main thread.

If `Framework.Dispose` happens on the same thread, why would that thread static get un-set?

Haselnussbomber:
Hm, yeah it should be true. Maybe the issue is somewhere else. 🤔

Haselnussbomber:
Oh, you know what.. it's probably somewhat the same thing as in #2232.
The services get disposed asynchronously, which includes the PluginManager on shutdown.
Because at that time, RunOnFrameworkThread is skipping the RunOnTick function, so they are not disposed on the main thread.

PR #2264, opened 2025-04-29
Move AntiDebug to xivfixes

Haselnussbomber:
**Summary:**

This PR moves the AntiDebug patch to xivfixes and enables it by default.

**Motivation:**

Often times, the game will crash with an unexpected error on shutdown due to having a debugger attached. This is caused by the AntiDebug service disposing and reverting the patch when Dalamud unloads.
Furthermore, I can't think of any reason to disable this, so the toggle is quite useless. There are no downsides to always have it patched when the game starts, since it's just skipping some code. 🤷‍♂️

**Changes:**

- Added `disable_game_debugging_protection` to xivfixes
- Removed the AntiDebug service
- Removed "Disable Debugging Protections" from the dev menu
- Removed the `IsAntiAntiDebugEnabled` property from DalamudConfiguration


KazWolfe:
I agree with the premise, but I'm not sure how I feel about crossing the line to disabling a "protection" SE placed upon the game, ineffective as it actually is. Having it as an option vs "default on" at least requires the user to opt in to the bypass.

There _is_ prior art for this in the form of the ACL patch, but that's also doable by just running the injector as an admin (although that's a bad idea for other reasons). I don't suppose it's a big deal, but it seems relevant enough to mention.

Dalamud.Boot/xivfixes.cpp
```diff
@@ -648,6 +648,41 @@ void xivfixes::symbol_load_patches(bool bApply) {
     }
 }
 
+void xivfixes::disable_game_debugging_protection(bool bApply) {
+    static const char* LogTag = "[xivfixes:disable_game_debugging_protection]";
+    static const std::vector<uint8_t> patchBytes = { 0x31, 0xC0, 0x90, 0x90, 0x90, 0x90 };
```
KazWolfe, line 653:
```suggestion
    static const std::vector<uint8_t> patchBytes = { 
        0x31, 0xC0, // XOR EAX, EAX
        0x90,       // NOP
        0x90,       // NOP
        0x90,       // NOP
        0x90        // NOP
    };
```

Dalamud.Boot/xivfixes.cpp
```diff
@@ -648,6 +648,41 @@ void xivfixes::symbol_load_patches(bool bApply) {
...
+    if (!g_startInfo.BootEnabledGameFixes.contains("disable_game_debugging_protection")) {
+        logging::I("{} Turned off via environment variable.", LogTag);
+        return;
+    }
+
+    const char* matchPtr = utils::signature_finder()
+        .look_in(utils::loaded_module(g_hGameInstance), ".text")
+        .look_for_hex("FF 15 ?? ?? ?? ?? 85 C0 74 13 41")
+        .find_one()
+        .Match.data();
```
KazWolfe, line 667:
```suggestion
    // Find IsDebuggerPresent in Framework.Tick()
    const char* matchPtr = utils::signature_finder()
        .look_in(utils::loaded_module(g_hGameInstance), ".text")
        .look_for_hex("FF 15 ?? ?? ?? ?? 85 C0 74 13 41")
        .find_one()
        .Match.data();
```

KazWolfe (approved):
Optional comments so we know what we're looking at.

KazWolfe:
> What about enabled if devplugins are loaded?

This would require we add devplugin logic to Dalamud.Boot, which seems somewhat unreasonable. The goal is to apply this as a "full" fix which applies through the entire life of the process.

There are also, of course, people who work on Dalamud proper without devplugins or otherwise have justification for electing to disable debugging.

MidoriKami:
I was more trying to suggest that it would be default off, but when you added a dev plugin, it would automatically toggle on.
As a majority of the people that are going to be needing to attach a debugger will be those that have a dev plugin loaded.

There should still be an actual toggle to enable or disable.

KazWolfe:
We would still need to add the requisite logic to `Dalamud.Boot` to read from DalamudConfig if this is kept in Boot. As far as I can tell, the intent of this change is to move _away_ from the anti-debug checks being in Dalamud proper so they're not torn down when the game closes.

Now, we could leak hooks in Dalamud proper, but I'm not sure that'd be a sane idea. If you have any arguments for why we _should_ keep this as a configurable setting, I'm open to hearing them though.

KazWolfe (approved):
BDFL approval received, technical changes good (per prior approval).

PR #2266, opened 2025-04-29
Do not throw in Hook.IsEnabled and Disable

Haselnussbomber:
I noticed that accessing `IsEnabled` and calling `Disable` on a disposed Hook throws `ObjectDisposedException`.
What it should do instead, in my opinion, is to provide a safe way to access `IsEnabled` and to be able to call `Disable`, the same way `Dispose` itself handes it - by checking `IsDisposed` and return early if `true`.

I also added the info that `Enable` can throw `ObjectDisposedException` to its xmldocs, and moved checks inside the locks.

### DIFF
--- a/Dalamud/Hooking/Hook.cs
+++ b/Dalamud/Hooking/Hook.cs
@@ -90,6 +90,7 @@ public abstract class Hook<T> : IDalamudHook where T : Delegate
     /// <summary>
     /// Starts intercepting a call to the function.
     /// </summary>
+    /// <exception cref="ObjectDisposedException">Hook is already disposed.</exception>
     public virtual void Enable() => throw new NotImplementedException();
 
     /// <summary>
--- a/Dalamud/Hooking/Internal/FunctionPointerVariableHook.cs
+++ b/Dalamud/Hooking/Internal/FunctionPointerVariableHook.cs
@@ -115,14 +115,7 @@ internal class FunctionPointerVariableHook<T> : Hook<T>
     }
 
     /// <inheritdoc/>
-    public override bool IsEnabled
-    {
-        get
-        {
-            this.CheckDisposed();
-            return this.enabled;
-        }
-    }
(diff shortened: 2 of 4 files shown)


KazWolfe:
Hm... there are some semantic things here. First, `.IsEnabled` not throwing `ObjectDisposedException` makes sense, but it's also not in a state where it _can_ be enabled. I don't think this would cause too much confusion since I don't think people are calling that prior to checking a hook enable.

I'm not sure how I feel about `.Disable()` not throwing an exception, since the hook is (by definition) no longer mutable given it's been disposed. Granted, this is a no-op as well.

Is there a compelling argument for `.Disable()` to not throw, or for a disposed hook to be explicitly disabled?

Haselnussbomber:
I can't tell you *how* it happened, and I also don't have the code anymore to check it (because it was temporary test code), but it somehow threw `ObjectDisposedException` on calling `Disable` in my Dispose function, causing my other services to not dispose anymore. It's somewhat an issue with `Microsoft.Extensions.DependencyInjection`, where it just stops disposing all other services after an exception. Jules also mentioned it a while ago. That's why I would prefer it to just be a no-op in the case it's already disposed. 🤷‍♂️

wolfcomp:
`Microsoft.Extensions.DependencyInjection` is very aggressive on how it deals with `IDisposable` and even can manage to take Dalamud handled disposes with it, to this end I had to reflect out its disposable container to check if the service was Dalamud or not and call dispose based on that.

Ref: https://github.com/wolfcomp/MonsterDex/blob/master/DeepDungeonDex/Main.cs#L122-L130

Critical-Impact:
I don't think Microsoft provides any guidance on this particular case so no help there.

Personally I think that if this exception is thrown then it's an indication you have your code disabling/disposing in the wrong order.

I don't have an issue with the change as such as I do think the logic is sound(we know it's disabled because it's disposed)

goaaats (approved):
I think this makes sense. I generally don't like it when properties throw.

PR #2268, opened 2025-05-01
Convert PluginStatWindow to ImRaii

foophoof:
Currently the Hooks tab asserts because of a missing ImGui.EndTabItem. Instead of just adding that, I took the opportunity to convert everything to use ImRaii instead.


Dalamud/Interface/Internal/Windows/PluginStatWindow.cs
```diff
@@ -198,158 +203,156 @@ public override void Draw()
...
+                            ImGui.Text($"{handlerHistory.Value.Average():F4}ms");
+                        }
+                    }
                 }
             }
-
-            ImGui.EndTabItem();
         }
 
-        if (ImGui.BeginTabItem("Hooks"))
```
foophoof, line 261: This one was missing an `ImGui.EndTabItem`.

KazWolfe (approved):

PR #2274, opened 2025-05-24
Add Completion module

salanth357:
Dalamud and plugin commands will now be tab-completable in the ChatLog.

This PR relies on a CS >= 413d5aa0b01202f6763f0b43fa01020f6284dc33


Dalamud/Game/Internal/Completion.cs
```diff
@@ -0,0 +1,284 @@
...
+        if (mod == null) return null;
+
+        var basePtr = mod->TextInput.TargetTextInputEventInterface;
+        if (basePtr == null) return null;
+
+        if (this.wantedVtblPtr == 0)
+        {
+            this.wantedVtblPtr =
+                Service<TargetSigScanner>.Get().GetStaticAddressFromSig(
+                    "48 89 01 48 8D 05 ?? ?? ?? ?? 48 89 81 ?? ?? ?? ?? 48 8D 05 ?? ?? ?? ?? 48 89 81 ?? ?? ?? ?? 48 8B 05 ?? ?? ?? ?? 48 8B 48 68",
```
goaaats, line 93: Is there an issue with moving this vtable sig to CS? Mostly asking this since we generally don't have good tooling to keep signatures up-to-date inside Dalamud, so we try to limit it to few that we know are stable. This one looks a bit chonky.

Dalamud/Game/Internal/Completion.cs
```diff
@@ -0,0 +1,284 @@
...
+    [ServiceManager.ServiceDependency]
+    private readonly Framework framework = Service<Framework>.Get();
+
+    private readonly EntryStrings dalamudCategory = new("【Dalamud】");
+    private readonly Dictionary<string, EntryStrings> cachedCommands = [];
+    private readonly ConcurrentQueue<string> addedCommands = [];
+
+    private Hook<CompletionModule.Delegates.GetSelection>? getSelection;
+
+    private volatile bool needsClear;
```
goaaats, line 40: Why does this need to be volatile? Could you leave a comment explaining why?

goaaats (commented):
Just a few small things. Looks great, good job :)

Dalamud/Game/Internal/Completion.cs
salanth357: not strictly, I just don't think CS has a good place to store it- this is the address of the vtable for a secondary base class on `AtkComponentTextInput`, and CS doesn't really have anywhere to store that right now, afaik.

Let me check it against a couple older versions to see how stable it is- that signature maps to the dtor, iirc, so it should be _reasonably_ stable

salanth357:
Updated with a comment for the volatile (and one explaining the sig and the pointer arithmetic there because it's _very_ C++).
No changes between 7.21 and 7.25 for functionality (tested locally), so this is once again ready for review

PR #2275, opened 2025-05-24
UIDebug2 Timeline Labelset support

MidoriKami:
Adds support for displaying data with regards to a timelines Labelset

This element defines the parents controller for child timelines

image


MidoriKami:
Needs more null checks, if it catches a frame where things are being reloaded, it explodes at the moment.

PR #2287, opened 2025-05-30
Allow to access all of Status span even if "NumValidStatuses" is 30

wolfcomp:
The span is always allocated with 60 so introducing `SpanLength` that returns the valid count to index from the methods but retaining the `Length` to be the old accessor with a different comment.

This shouldn't change any behaviors unless people have hardcoded their lengths for index accessing to be 60 at all times.

### DIFF
--- a/Dalamud/Game/ClientState/Statuses/StatusList.cs
+++ b/Dalamud/Game/ClientState/Statuses/StatusList.cs
@@ -35,12 +35,12 @@ public sealed unsafe partial class StatusList
 
     /// <summary>
     /// Gets the amount of status effect slots the actor has.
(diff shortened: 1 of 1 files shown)


Caraxi:
There should never be any data at all inside the upper 30 unless the Valid value is set to 60. If the client ever sets Index 30+ it will automatically set the higher limit, so there should be no reason to bypass the "valid" limit

Caraxi:
Just because the game processes useless data doesn't mean plugins should

MidoriKami:
Probably safe to close this, because as Caraxi noted, when NumValidStatus is 30, it means you'll never actually see any valid data in indexes 31..60

goaaats:
Alright. If this comes back and makes sense again, please reopen!

PR #2302, opened 2025-06-21
[ChatGui] Remove unnecessary conversion step for performance improvement

Infiziert90:
Just a small improvement by removing the `EncodeWithNullTerminator()` at the start of `HandlePrintMessageDetour`, as we can get the null terminated byte span directly from the Utf8String instance.


Infiziert90:
Requires dalamud payload encoding/decoding to be fixed.

PR #2305, opened 2025-06-25
Completion rewrite

Haselnussbomber:
An almost complete rewrite of the Dalamud Completion system.

The suggestions list is now completely rebuild immediately before the auto-completion dropdown is opened, thus only displaying the Dalamud category and commands when needed - in the ChatLog.


Haselnussbomber:
I rebased the branch and squashed a3b8e6c, a4161db, 3db43b2 into one commit (368b6e3).
In the last force-push I renamed the service to DalamudCompletion, so it somewhat aligns with DalamudAtkTweaks and changed the setup from using the Update event to use RunOnTick. ([compare](https://github.com/goatcorp/Dalamud/compare/368b6e37acfe084acc5fa31d64e37d13641f6eb0..8106943f0b63906981b5d2da0da1e6737d1f9c6a)).

PR #2308, opened 2025-07-01
Add ITextureProvider.CreateTextureFromSeString

Soreepeong:
image
image


KazWolfe:
What's the practical benefit of this?

MidoriKami:
I don't remember all the details of the discussion but I believe this came about from a conversation about Mappy and being able to generate textures from strings so that we could generate super high resolution textures without needing to persist an entire font. This was a while ago so my memory is a bit fuzzy.

MidoriKami:
I also believe I wanted to be able to take FontAwesome icons, and inject them into native ui, but to do that I need to be able to get them as a texture.

Soreepeong:
Rebased.

goaaats (approved):

PR #2318, opened 2025-07-20
[imgui-bindings] Fix ImGui.MenuItem calls

Haselnussbomber:
Arguments of functions are now generated in order, which means for functions like `bool MenuItem(const char* label, const char* shortcut = NULL, bool selected = false, bool enabled = true)` it will generate an overload `bool MenuItem(string label, bool selected)`, instead of `bool MenuItem(string label, bool enabled)`.


Haselnussbomber:
I removed the commit that converted strings across the whole codebase to u8 literals.
Just get these Menuitem changes in here for the dev menu.

PR #2322, opened 2025-07-26
[Api13] Update ChatLinkHandler functions

Haselnussbomber:
- Move functions from `IDalamudPluginInterface` to `IChatGui`
- Switch type of `CommandId` from `uint` to `Guid`, so they are unique
- Generate `CommandId` automatically when registering a command, so plugins don't have to specify them

### DIFF
--- a/Dalamud/Game/Gui/ChatGui.cs
+++ b/Dalamud/Game/Gui/ChatGui.cs
@@ -50,7 +50,6 @@ internal sealed unsafe class ChatGui : IInternalDisposableService, IChatGui
     [ServiceManager.ServiceDependency]
     private readonly DalamudConfiguration configuration = Service<DalamudConfiguration>.Get();
 
-    private uint dommandIdCounter = 0;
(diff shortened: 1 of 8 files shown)


Dalamud/Game/Gui/ChatGui.cs
```diff
@@ -49,7 +50,8 @@ internal sealed unsafe class ChatGui : IInternalDisposableService, IChatGui
     [ServiceManager.ServiceDependency]
     private readonly DalamudConfiguration configuration = Service<DalamudConfiguration>.Get();
 
-    private ImmutableDictionary<(string PluginName, uint CommandId), Action<uint, SeString>>? dalamudLinkHandlersCopy;
+    private uint dommandIdCounter = 0;
```
KazWolfe: dead var?

Haselnussbomber: lol. that was an idea i had before i switched to Guid and forgot to remove it again (because not unique). also, nice typo me

PR #2325, opened 2025-07-26
IDtrBar Add Additional Click Events

MidoriKami:
https://github.com/user-attachments/assets/dbc5e199-f92c-4adf-9730-e0394144ea35

Might also consider doing something different with HasClickAction


MidoriKami:
Intended API Breakage, awaiting a API13 branch to re-target this PR

PR #2328, opened 2025-07-27
IObjectTable Helpful Enumerables

MidoriKami:
Adds some very simple enumerables to allow devs to more easily iterate the sections of the object table they actually care about

### DIFF
--- a/Dalamud/Game/ClientState/Objects/ObjectTable.cs
+++ b/Dalamud/Game/ClientState/Objects/ObjectTable.cs
@@ -168,7 +168,7 @@ internal sealed partial class ObjectTable : IServiceType, IObjectTable
     {
         for (var index = 0; index < 200; index += 2)
         {
-            if (this[index] is IBattleChara gameObject)
+            if (this[index] is IBattleChara { ObjectKind: ObjectKind.Player } gameObject)
             {
                 yield return gameObject;
             }
@@ -179,7 +179,7 @@ internal sealed partial class ObjectTable : IServiceType, IObjectTable
     {
         for (var index = range.Start.Value; index <= range.End.Value; index++)
         {
-            if (this[index] is { ObjectKind: ObjectKind.Player } gameObject)
+            if (this[index] is { } gameObject)
             {
                 yield return gameObject;
             }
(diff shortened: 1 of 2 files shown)


Dalamud/Game/ClientState/Objects/ObjectTable.cs
```diff
@@ -66,6 +66,24 @@ public unsafe nint Address
     /// <inheritdoc/>
     public int Length => objectTableLength;
 
+    /// <inheritdoc/>
+    public IEnumerable<IBattleChara> PlayerObjects => this.GetPlayerObjects();
+
+    /// <inheritdoc/>
+    public IEnumerable<IGameObject> CharacterManagerObjects => this.GetObjectsInRange(..199);
```
KazWolfe, line 73: Are these IDs documented somewhere (else) so we can refer to them more directly?

(e.g. as part of CS)

MidoriKami: https://github.com/aers/FFXIVClientStructs/blob/53adaf4c021c3001172ed268201cb81d77903c24/FFXIVClientStructs/FFXIV/Client/Game/Object/GameObjectManager.cs#L17

Haselnussbomber: - [`CharacterManager.Instance()->BattleCharas.Length`](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/Character/CharacterManager.cs#L12)
- [`ClientObjectManager.Instance()->BattleCharas.Length`](https://github.com/aers/FFXIVClientStructs/blob/main/FFXIVClientStructs/FFXIV/Client/Game/Object/ClientObjectManager.cs#L14)

etc.. but we don't have a EventObjectManager yet.
And the MJI/WKSManager have their own, so you'd have to pick one, I guess.

MidoriKami: So in theory, this could later be updated to be more automatic in nature? Fortunately, that wouldn't effect the api surface, so could be done at any time.

PR #2330, opened 2025-07-30
[Api13] Add native wrapper structs

Haselnussbomber:
Adds ClientStructs-independent wrappers around `AtkUnitBase`, `AtkValue`, `AgentInterface` and `UIModule`, providing basic properties (well, except for UIModulePtr which is just a placeholder for now).
The `IGameGui` functions were updated appropriately.
The wrappers implicitly cast from and to `nint`, but the address is also exposed as readonly field, so adapting to it should be pretty trivial.


Haselnussbomber:
It's a struct because it's a wrapper around the address. That's how the new ImGui bindings and CStringPointer in CS work too.

PR #2331, opened 2025-08-03
Provide native linux compatibility for Dalamud.Bindings.ImGui

Critical-Impact:
I'm aware this PR probably doesn't add anything of value to most plugin developers but this allows for the bindings to run outside of the normal plugin environment(currently I rely on being able to use .so's instead of .dll's for DalaMock). This was handled automatically before but it looks like we've hardcoded in dlls.

This checks the platform(Win32NT on wine/windows) and for the existence of the .so files before loading them.

Open to suggestions if there's a smarter way to handle this

### DIFF
--- a/imgui/Dalamud.Bindings.ImGui/ImGui.cs
+++ b/imgui/Dalamud.Bindings.ImGui/ImGui.cs
@@ -20,10 +20,18 @@ namespace Dalamud.Bindings.ImGui
             {
                 InitApi(new NativeLibraryContext(Process.GetCurrentProcess().MainModule!.BaseAddress));
             }
+
+            var linuxPath = Path.Combine(Path.GetDirectoryName(Assembly.GetCallingAssembly().Location)!, GetLibraryName() + ".so");
+            var windowsPath = Path.Combine(Path.GetDirectoryName(Assembly.GetCallingAssembly().Location)!, GetLibraryName() + ".dll");
+
+            // This shouldn't affect wine as it'll be reported as Win32NT
+            if (System.Environment.OSVersion.Platform == PlatformID.Unix && File.Exists(linuxPath))
+            {
+                InitApi(new NativeLibraryContext(linuxPath));
+            }
             else
             {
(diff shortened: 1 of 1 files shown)


imgui/Dalamud.Bindings.ImGui/ImGui.cs
```diff
@@ -20,10 +20,18 @@ static ImGui()
             {
                 InitApi(new NativeLibraryContext(Process.GetCurrentProcess().MainModule!.BaseAddress));
             }
+
+            var linuxPath = Path.Combine(Path.GetDirectoryName(Assembly.GetCallingAssembly().Location)!, GetLibraryName() + ".so");
+            var windowsPath = Path.Combine(Path.GetDirectoryName(Assembly.GetCallingAssembly().Location)!, GetLibraryName() + ".dll");
```
wolfcomp, line 25: Mac which is if I recall `osx` is using a file format called `.dylib` which can be found here <https://github.com/goatcorp/Hexa.NET.ImGui/tree/dalamud/Hexa.NET.ImGui/native/osx-x64> in the repo. *I am guessing this is also copied over when building but please verify*

Critical-Impact: Just so it's on the record and not only in discord

I'm not sure those are the same ones that dalamud uses(at least not yet)
Specifically,
https://github.com/goatcorp/gc-cimgui/tree/backends
https://github.com/goatcorp/gc-cimguizmo
https://github.com/goatcorp/gc-cimplot/

I have forks of each that build a linux specific version

I believe that's because we are still targeting the old imgui + we have some changes to imgui for assert callbacks

That's not to say it wouldn't be possible to generate the .dylib files but atm dalamud as is only generates the DLLs

I think it would be nice to have dalamud generate the imgui bindings for the platforms just in case somebody does want to use the same imgui dalamud is but it'd require a bit of reworking of how it builds which I think would be best done after the patch

KazWolfe: We can safely add dylib support once someone asks.

PR #2332, opened 2025-08-03
[API 13] Add Lumina.Excel as submodule

WorkingRobot:
Hopefully this should work without problems. Moving Lumina.Excel into a submodule allows it to be updated alongside CS. All breaking changes in the schema will be put in the .Experimental namespace for plugin developers to make use of (additionally marked with the Experimental attribute). Lumina.Excel will only ever update the main namespace on major releases, so this shouldn't have any effect on the stable API surface.


KazWolfe (approved):
Seems OK to me, though going to page @goaaats for a specific review in case there are other concerns re API surface here. We'll likely need to set up automations; I think we already have apicompat working though.

Critical-Impact:
This is kind of unrelated but seeing Lumina has generally moved in lock step with dalamud, will there be a different release schedule for the nuget lumina now?

PR #2334, opened 2025-08-04
[API13] Fire ActivePluginsChanged after a plugin loaded/unloaded

### LINKED ISSUE #2300: ActivePluginsChanged only fires when plugins are updated
It seems `DalamudPluginInterface.ActivePluginsChanged` only gets fired when plugins are updated (or the update button in the plugin installer is pressed, even if there are no plugins with updates.)

Looks like `PluginManager.UpdatePluginsAsync` is the only thing that (indirectly) calls it.

Haselnussbomber:
I kept it very simple: all plugins are notified whenever a plugin loaded or unloaded.
For this I added two new `PluginListInvalidationKind` values: `Loaded` and `Unloaded`.
No changes to the behavior of `Update` or `AutoUpdate`, so these events will be triggered too.

Closes #2300 (actually it doesn't because "The pull request must be on the default branch." 🙄 )

### DIFF
--- a/Dalamud/Plugin/ActivePluginsChangedEventArgs.cs
+++ b/Dalamud/Plugin/ActivePluginsChangedEventArgs.cs
@@ -5,15 +5,27 @@ namespace Dalamud.Plugin;
 /// <summary>
 /// Contains data about changes to the list of active plugins.
 /// </summary>
-public class ActivePluginsChangedEventArgs(PluginListInvalidationKind kind, IEnumerable<string> affectedInternalNames) : EventArgs
+public class ActivePluginsChangedEventArgs : EventArgs
 {
+    /// <summary>
+    /// Initializes a new instance of the <see cref="ActivePluginsChangedEventArgs"/> class
+    /// with the specified parameters.
+    /// </summary>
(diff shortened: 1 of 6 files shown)


KazWolfe:
General thought: as this is API13, would it make sense to redeclare the type for `ActivePluginsChangedDelegate` to include the specific plugin(s) that actually were updated?

Otherwise, should we continue consumers to just check the complete list?

Haselnussbomber:
That wouldn't be a problem. We can easily forward the `IEnumerable<string> affectedInternalNames` parameter from `PluginManager.NotifyPluginsForStateChange`.
Edit: I have the code ready and just have to push.

MidoriKami:
So say someone changes a collection, I'd get an event that says "Loaded", and a list of InternalNames of plugins that are loaded? That sounds pretty great.

Haselnussbomber:
> So say someone changes a collection, I'd get an event that says "Loaded", and a list of InternalNames of plugins that are loaded? That sounds pretty great.

Sorry I didn't answer that. Yes and no - the Loaded event is triggered for each plugin individually.

PR #2339, opened 2025-08-04
Add IPlayerState service

Haselnussbomber:
This PR adds a `IPlayerState` service, providing most of the things that the PlayerState in CS has to offer, including some more unlock functions from UIState, but it doesn't support *all* unlockable things yet (there is a list in the `UpdateUnlocks` function).

I had this in a branch locally and polished it up last minute (added the PluginScoped service). Not sure why/where I stopped working on it.
Maybe someone can look though and express opinions on this? *caugh* @KazWolfe *caugh*

### DIFF
--- a/Dalamud/Game/PlayerState/PlayerAttribute.cs
+++ b/Dalamud/Game/PlayerState/PlayerAttribute.cs
@@ -1,0 +1,489 @@
+namespace Dalamud.Game.PlayerState;
+
+/// <summary>
+/// Represents a player's attribute.
+/// </summary>
+public enum PlayerAttribute
+{
+    /// <summary>
+    /// Strength.
+    /// </summary>
+    /// <remarks>
+    /// Affects physical damage dealt by gladiator's arms, marauder's arms, dark knight's arms, gunbreaker's arms, pugilist's arms, lancer's arms, samurai's arms, reaper's arms, thaumaturge's arms, arcanist's arms, red mage's arms, pictomancer's arms, conjurer's arms, astrologian's arms, sage's arms, and blue mage's arms.
+    /// </remarks>
+    Strength = 1,
+
+    /// <summary>
+    /// Dexterity.
+    /// </summary>
+    /// <remarks>
+    /// Affects physical damage dealt by rogue's arms, viper's arms, archer's arms, machinist's arms, and dancer's arms.
+    /// </remarks>
+    Dexterity = 2,
+
+    /// <summary>
+    /// Vitality.
+    /// </summary>
+    /// <remarks>
+    /// Affects maximum HP.
+    /// </remarks>
+    Vitality = 3,
+
+    /// <summary>
+    /// Intelligence.
+    /// </summary>
+    /// <remarks>
+    /// Affects attack magic potency when role is DPS.
+    /// </remarks>
+    Intelligence = 4,
+
+    /// <summary>
+    /// Mind.
+    /// </summary>
+    /// <remarks>
+    /// Affects healing magic potency. Also affects attack magic potency when role is Healer.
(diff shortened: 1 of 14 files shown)


Dalamud/Game/PlayerState/PlayerStateAddressResolver.cs
```diff
@@ -0,0 +1,24 @@
...
+    public nint PerformMateriaActionMigration { get; private set; }
+
+    /// <summary>
+    /// Scan for and setup any configured address pointers.
+    /// </summary>
+    /// <param name="sig">The signature scanner to facilitate setup.</param>
+    protected override void Setup64Bit(ISigScanner sig)
+    {
+        // RaptureHotbarModule.PerformMateriaActionMigration
+        this.PerformMateriaActionMigration = sig.ScanText("E8 ?? ?? ?? ?? 48 8D 0D ?? ?? ?? ?? E8 ?? ?? ?? ?? 48 8B 8B ?? ?? ?? ?? 48 8B 01");
```
KazWolfe, line 22: This should absolutely be in CS.

Dalamud/Game/PlayerState/PlayerState.cs
```diff
@@ -0,0 +1,153 @@
...
+
+using FFXIVClientStructs.FFXIV.Client.UI;
+using FFXIVClientStructs.FFXIV.Client.UI.Misc;
+
+namespace Dalamud.Game.PlayerState;
+
+/// <summary>
+/// This class represents the state of the local player.
+/// </summary>
+internal unsafe partial class PlayerState : IInternalDisposableService, IPlayerState
```
KazWolfe, line 18: While I get we're trying to reflect what CS does here, I'm not sure it's the best idea to have one giant class that handles _all_ of `PlayerState` - especially since (iirc) playerstate ultimately does more that we haven't discovered.

Is there a compelling reason to have everything as partials vs having a subclass for various "notable" components to create a more cohesive API surface?

Dalamud/Game/PlayerState/PlayerAttribute.cs
```diff
@@ -0,0 +1,489 @@
+namespace Dalamud.Game.PlayerState;
+
+/// <summary>
+/// Represents a player's attribute.
+/// </summary>
+public enum PlayerAttribute
```
KazWolfe, line 6: Aren't these attributes somewhere in EXD?

Dalamud/Game/PlayerState/PlayerStateAddressResolver.cs
Haselnussbomber: Sure, we can add that. It just isn't really anthing people should call and might be inlined some day.
Maybe we can hook the `RaptureAtkModule.Update` function in Dalamud, since I always had trouble doing that in my plugins (it caused deadlocks).

Dalamud/Game/PlayerState/PlayerState.cs
Haselnussbomber: You kinda answered your own question. I made it partial because it was so much and grouped different things (core/hooks/events, wrappers, unlocks) into their own files.
I was thinking of an IUnlockState service, but there is also not really a compelling reason to *not* have it in a single IPlayerState service, imo.

Dalamud/Game/PlayerState/PlayerAttribute.cs
Haselnussbomber: Yes, in the BaseParam sheet.
The reason I added this was to have a nicer API, e.g. `GetAttribute(PlayerAttribute.Craftsmanship)`.

PR #2346, opened 2025-08-06
Fix/imgui

Soreepeong:
* Passing `ReadOnly`/`Span<byte>.Empty`, `ReadOnly`/`Span<char>.Empty`, `default(ReadOnly`/`Span<byte>)`, or `default(ReadOnly`/`Span<char>)` will be effectively passing `null`, rather than an empty string, to an `ImU8String`. `""` and `""u8` passes the empty string.
* Added explicit overload for `ImU8String.AppendFormatted(object)`.
* Changed to explicitly release any focus upon clicking on the game window that is not ImGui related.


Dalamud/Interface/Internal/DalamudInterface.cs
```diff
@@ -575,16 +575,6 @@ private void OnDraw()
 
             if (this.isCreditsDarkening)
                 this.DrawCreditsDarkeningAnimation();
-
-            // Release focus of any ImGui window if we click into the game.
-            var io = ImGui.GetIO();
```
goaaats, line 580: This is still desirable though, right? We want ImGui windows to unfocus if you click anything that is not an ImGui window, so that keyboard focus is lost.

goaaats (approved):

PR #2352, opened 2025-08-07
Work around ActiveIdUsingKeyInputMask having a wrong type

Soreepeong:
`ActiveIdUsingKeyInputMask` should be of the following type:
```cpp
typedef ImBitArray<ImGuiKey_NamedKey_COUNT, -ImGuiKey_NamedKey_BEGIN>    ImBitArrayForNamedKeys;
```

But it was generated as a `nuint`.

This PR changes the filter script so that it replaces `nuint` to the generated type `ImBitArrayImGuiKeyNamedKeyCOUNTLessImGuiKeyNamedKeyBEGIN`. Proper fix PR should be made to upstream Hexa.

All other re-filtered files are stripping down on double newline endings.


goaaats (approved):

Soreepeong:
Also did a simple workaround for the invalid error message (it does compile but IDE complains):

image

Soreepeong:
Added overloads for `ImGui.InputText` so that the callback can take `ImGuiInputTextCallbackDataPtr` instead of `scoped ref ImGuiInputTextCallbackData`.

```cs
ImGui.InputText("test1", ref test1, 800);
ImGui.InputText("test2", ref test2, 800, ImGuiInputTextFlags.CallbackAlways, _ => 0);
ImGui.InputText("test3", ref test3, 800, ImGuiInputTextFlags.CallbackAlways, (scoped ref ImGuiInputTextCallbackData data) => 0);
```

Soreepeong:
* `RegisterClassExW` should no longer collide with other addons that uses ImGui apart from Dalamud (should fix OptiScaler).
* It should no longer show random triangles when buffers have to be resized.

Dalamud/Interface/ImGuiBackend/Renderers/Dx11Renderer.cs
```diff
@@ -264,7 +264,7 @@ private void RenderDrawDataInternal(
             this.indexBuffer.Dispose();
         if (this.indexBuffer.Get() is null)
         {
-            this.indexBufferSize = drawData.TotalIdxCount + 5000;
+            this.indexBufferSize = drawData.TotalIdxCount + 16384;
```
goaaats, line 267: Where do these numbers come from?

goaaats (approved):

Dalamud/Interface/ImGuiBackend/Renderers/Dx11Renderer.cs
Soreepeong: There's no justification for this really but since it's 32KB of more VRAM usage I thought, why not.

PR #2366, opened 2025-08-11
Set ClientState.TerritoryType on load

Haselnussbomber:
This fixes an issue where TerritoryType reports 0 after Dalamud is injected while the player is already logged in.
TerritoryType was only updated when a InitZone packet was received.

Setting it in the ctor isn't possible, because DataManager isn't initialized at that point and we need it for the IsPvP lookup. So I moved it off to the first Framework tick.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -184,7 +184,7 @@ internal sealed class ClientState : IInternalDisposableService, IClientState
 
     private unsafe void Setup()
     {
-        this.SetTerritoryType((ushort)GameMain.Instance()->CurrentTerritoryTypeId);
+        this.TerritoryType = (ushort)GameMain.Instance()->CurrentTerritoryTypeId;
     }
 
     private unsafe void SetupTerritoryTypeDetour(EventFramework* eventFramework, ushort territoryType)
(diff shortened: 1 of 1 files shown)


Dalamud/Game/ClientState/ClientState.cs
```diff
@@ -180,8 +182,22 @@ void IInternalDisposableService.DisposeService()
         this.networkHandlers.CfPop -= this.NetworkHandlersOnCfPop;
     }
 
+    private unsafe void Setup()
+    {
+        this.SetTerritoryType((ushort)GameMain.Instance()->CurrentTerritoryTypeId);
```
KazWolfe: This will fire the TerritoryChanged callback instantly on a post-login injection.

This *seems* acceptable, but could cause confusion if someone is *only* expecting this event to fire when the user actually navigates to a new territory - or, rather, in response to that hook firing. I suppose this should already happen with normal logins.

I suppose the crux of the question is whether a post-inject login should count as a territory "change" - in some ways, I don't think so, since the territory was never actually (truly) zero. In other ways though, it seems to make logical sense to actually fire that event ASAP so plugins can react rather than following the "subscribe and read" pattern.

Haselnussbomber: Given that other events aren't fired immediately when injecting post-login, I changed it so that doesn't happen with `TerritoryChanged` to align expectations.

KazWolfe (approved):

KazWolfe:
Ref #2108. See outdated comment for an API quirk; we may want to go back to discuss that at some point.

Dalamud/Game/ClientState/ClientState.cs
MidoriKami: Current behavior when not injecting after load has TerritoryChanged occuring after login, if this change causes a TerritoryChanged event to happen before login, then at least in my code, things are likely to break.

I generally setup things in OnLogin, and use TerritoryChanged to reset/invalidate cached data.

PR #2367, opened 2025-08-11
Avoid c++ exceptions in import hook

marzent:
This fixes crashes in wine in conjunction with DllInject.


KazWolfe (approved):
thats a lot of avoiding try-catch

PR #2370, opened 2025-08-12
ISeStringEvaluator: Add ReadOnlySpan<byte> support

Haselnussbomber:
- Adds an overload for `ISeStringEvaluator.EvaluateMacroString` taking `ReadOnlySpan<byte>`.
  - Supported by [ReadOnlySeString.FromMacroString](https://github.com/NotAdam/Lumina/blob/master/src/Lumina/Text/ReadOnly/ReadOnlySeString.cs#L305)
- Adds an implicit cast from `ReadOnlySpan<byte>` to `SeStringParameter`.

### DIFF
--- a/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
+++ b/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
@@ -122,6 +122,15 @@ internal class SeStringEvaluator : IServiceType, ISeStringEvaluator
     }
 
     /// <inheritdoc/>
+    public ReadOnlySeString EvaluateMacroString(
+        ReadOnlySpan<byte> macroString,
+        Span<SeStringParameter> localParameters = default,
+        ClientLanguage? language = null)
+    {
+        return this.Evaluate(ReadOnlySeString.FromMacroString(macroString).AsSpan(), localParameters, language);
+    }
+
+    /// <inheritdoc/>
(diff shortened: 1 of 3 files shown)


KazWolfe:
Is it just me, or is the explosion of `ReadOnlySpan<byte>` everywhere getting kind of unsettling? 

It's not _wrong_, but it does feel akin to allowing `void*` (or, really, any data type representable by an array of bytes) and something the type checker ***should*** be screaming bloody murder at.

Where is the line between "oh, it's easy to write" and "devs need to explicitly declare things"?

Soreepeong:
`ReadOnlySpan<byte>` is C#'s way of representing UTF-8 strings at least according to [this doc](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals#u8-suffix-on-string-literals) so it should be fine. While anything can be cast to `ReadOnlySpan<byte>` it does require going through `MemoryMarshal.Cast` or some `Unsafe.` functions so it's different from `void*` in that you must explicitly express your intent to reinterpret-cast something.

KazWolfe:
Sometimes it [feels like MS makes bad decisions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-11.0/utf8-string-literals#wait-for-utf8string-type).

PR #2373, opened 2025-08-15
Reduce usage of exceptions from Boot

Soreepeong:
Incorporates #2372.

Also changed error dialogs.

image
image


Soreepeong:
Wine specific: Changed xivfixes to run on its own thread, just in case it makes any difference on exceptions. Main thread still blocks for that thread to complete.

KazWolfe (dismissed):

KazWolfe:
> Wine specific: Changed xivfixes to run on its own thread, just in case it makes any difference on exceptions. Main thread still blocks for that thread to complete.

I'd like verification from @marzent about this, so they're required for PR approval.

marzent (approved):
LGTM and works as expected

KazWolfe (approved):

PR #2377, opened 2025-08-19
feat: Identify the plugin causing an assertion failure

KazWolfe:
Modifies the assertion error to blame a specific plugin that did something wrong, when available:

image


KazWolfe:
image

image

goaaats (approved):

PR #2378, opened 2025-08-20
[master] Update ClientStructs

bleatbot:


Critical-Impact (approved):
LGTM, if this does cause issues we'll either rollback or deploy additional fixes

Issue #2383, opened 2025-08-24
[API14] IPC Redesign


KazWolfe:
The IPC system in Dalamud has, historically, been far from perfect with a number of pitfalls and other annoyances that make it difficult to work with.

With API 14, we should take the opportunity to rework the IPC surface to be better/more useful. This Issue serves to (a) catalogue possible ideas for the IPC surface, and (b) allow IPC consumers to provide feedback to these proposals.

KazWolfe:
At minimum, I want to see the following for an IPC rework:

* Support for `async` methods.
* Namespacing for IPC so that plugins can have better control over who they call.
  * As an example, two plugins can register `MyIPC.Endpoint`. Consumers can call `PluginA::MyIPC.Endpoint` to specify which plugin specifically receives the IPC.
  * Plugins that don't care about who handles an IPC call should still be able to use `MyIPC.Endpoint`.
* Passing IPC context to the called method, so plugins know who/what is calling them.
  * Can be implemented now with `IIPCProvider#GetContext()` and thread-locals, most likely.
* Clarify the difference between RPC and getting PubSub events from other plugins.

haroldiedema:
## Centralized Service Locator Concept

I would like to propose an IPC mechanism based on shared assemblies rather than the traditional IPC methods, to eliminate the need for complex serialization and deserialization processes. This approach would involve having plugin developers creating an intermediary assembly (an SDK of sorts) that both the owning plugin and the consuming plugin would reference.

The model I'm proposing is a form of a **Service Locator** or **Dependency Injection** pattern, where Dalamud acts as the central hub. Instead of plugins directly referencing and instantiating a shared library, they ask Dalamud for the "official" instance of that library.

Here's how it would work conceptually:

1. **Provider Plugin (Plugin A):** This plugin is the owner of the shared library. It implements the library's functionality and, crucially, registers a singleton or a specific instance of that library with Dalamud's API. This registration might include metadata like the library's name, version, and the API surface it exposes.
2. **Consumer Plugin (Plugin B):** This plugin wants to use the functionality provided by Plugin A. It doesn't instantiate the shared library itself. Instead, it queries Dalamud's API for the registered instance of that library, specifying the version it expects.
3. **Dalamud's Role:** Dalamud acts as the intermediary. When Plugin B requests the shared library, Dalamud checks its internal registry.
    - If the library has been registered by Plugin A, Dalamud performs **version and compatibility checks** to ensure that Plugin B can safely use it.
[...]

PR #2386, opened 2025-08-25
feat: Add IPC Context

KazWolfe:
Extremely rough first draft of the IPC context. Contexts allow receiving plugins to understand who is calling them without walking the stack. This is useful for logging/diagnostic purposes, displaying warnings to end users (e.g. if a plugin is using an outdated API), subtle behavioral/compatibility changes, and similar.

See #2383 for the larger discussion.

### DIFF
--- a/Dalamud/Plugin/Ipc/Internal/CallGateChannel.cs
+++ b/Dalamud/Plugin/Ipc/Internal/CallGateChannel.cs
@@ -2,6 +2,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 
 using Dalamud.Plugin.Internal.Types;
 using Dalamud.Plugin.Ipc.Exceptions;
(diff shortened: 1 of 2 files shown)


Dalamud/Plugin/Ipc/IpcContext.cs
```diff
@@ -0,0 +1,15 @@
...
+/// </summary>
+public class IpcContext
+{
+    /// <summary>
+    /// Gets the plugin that initiated this IPC call.
+    /// </summary>
+    public IExposedPlugin? SourcePlugin { get; init; }
+
+    /// <inheritdoc/>
+    public override string ToString() => $"<IpcContext SourcePlugin={this.SourcePlugin?.Name ?? "null"}>";
```
goaaats, line 14: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.debuggerdisplayattribute?view=net-9.0

NanaKhide:
Would be incredibly useful for me.
At the moment I'd need to walk the stack to add a transparency feature to my recording plugin.
WIth this it would be easier to show the user which plugin requested mine to record the screen via ipc.

PR #2400, opened 2025-09-05
Update Settings Window

Haselnussbomber:
- Made all classes in the `Dalamud.Interface.Internal.Windows.Settings` namespace internal, as they should be.
- Fixed that the window title as well as setting names and descriptions were not updated when the language changed.
  - Introduced an internal `LazyLoc` type which fetches the localized text when its used (in `ToString()`).
- Fixed that the search bar overlapped the tab bar in certain languages by simply moving it above them.
- Fixed that the currently opened tab would switch to Auto-Updates on language change.
- Disabled the search bar while credits are scrolling in the About tab, so the tab isn't overwritten with the search results.
- Added new localizable text keys:
  - `DalamudSettingsSearchPlaceholder` = `Search for settings...`
  - `DalamudSettingsSearchResults` = `Search Results`
  - `DalamudSettingsNoSearchResultsFound` = `No results found...`
- Due to how LazyLoc works, I had to replace the placeholders in `DalamudSettingsPluginTestHint` with the actual text.
- `Localization` now loads Dalamuds locales before firing off the `LanguageChanged` event, so that listeners can get the updated locs.
- Other small updates:
  - Moved tab and search results logic into their own methods.
  - Removed unused `IsVisible` property in `SettingsTab`.
  - Added `Kind` getter to SettingsTab to remove hardcoded `select` in `SetOpenTab(SettingsOpenKind)`.
  - Applied code suggestions by Visual Studio. 🤷‍♂️

https://github.com/user-attachments/assets/21f651ab-9ff1-423d-b388-cd2078ac47a8


KazWolfe (approved):
yolo merge

PR #2401, opened 2025-09-05
[API15] Window system upgrades

Critical-Impact:
This PR makes adjustments to how dalamud's windowing system works.  

- Specifically it interfaces both Window and WindowSystem, allowing for developers to define their own classes instead of needing to inherit from Window directly.

- It creates a wrapper class called WindowHost that wraps the IWindow implementation the plugin developer provides, keeping track of it's state.

- Interfacing WindowSystem was probably not necessary but could allow us to potentially provide alternate versions of the WindowSystem(one suggestion was overlays vs regular windows)

- These are breaking changes and would require to be released with an API bump.

Things that I still haven't resolved and I may resolve in a separate PR:

 - There needs to be a window management service, something like IWindowManager or IWindowService that dalamud provides. We should take ownership of WindowSystem instead of letting users construct it directly. You should be able to request a WindowSystem from the service

 - WindowSystem has 3 statics, HasAnyWindowSystemFocus, FocusedWindowSystemNamespace and TimeSinceLastAnyFocus. These sit outside the abstraction and should be moved into above service.

 - I'm not 100% happy about how IWindow.IsFocused works, it needs to be public so that the WindowHost can set it but it also means the plugin developer can set it. It won't actually focus the window and the WindowSystem will set it back next frame.

### DIFF
--- a/Dalamud/Interface/Windowing/IWindow.cs
+++ b/Dalamud/Interface/Windowing/IWindow.cs
@@ -1,7 +1,3 @@
-// <copyright file="IWindow.cs" company="PlaceholderCompany">
-// Copyright (c) PlaceholderCompany. All rights reserved.
-// </copyright>
-
 using System.Collections.Generic;
 using System.Numerics;
 
@@ -27,7 +23,7 @@ public interface IWindow
     string WindowName { get; set; }
 
     /// <summary>
-    /// Gets a value indicating whether the window is focused.
+    /// Gets or sets a value indicating whether the window is focused.
     /// </summary>
     bool IsFocused { get; set; }
 
@@ -137,7 +133,7 @@ public interface IWindow
     bool IsOpen { get; set; }
 
     /// <summary>
-    /// Gets or sets a value indicating whether this window will request focus from the window system next frame
+    /// Gets or sets a value indicating whether this window will request focus from the window system next frame.
     /// </summary>
     public bool RequestFocus { get; set; }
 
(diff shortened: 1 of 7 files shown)


Critical-Impact:
I've updated this to include the changes added to Window/WindowSystem since I first made this PR.

- IsPinned and IsClickthrough both suffer from the same issue as IsFocused. I don't really have a solution, realistically it should be something you query the WindowSystem for.
- ShouldInhibitAtkCloseEvents should probably move to the WindowService when it gets created

Critical-Impact:
I think as a first step this would be enough to provide a decoupled window and window system for people to use.
The changes are fairly minimal on plugin developers, the only change is where WindowSizeConstraints live so it's just a matter of adding a new using line.

A WindowService would be nice in the future but at least in my opinion this would be good enough to merge on it's own but I imagine @goaaats will need to weigh in

MidoriKami:
Are you sure you want this targeting `api15-rollup` instead of `api15` branch?

Critical-Impact:
Yes, api15-rollup is 10 commits ahead, 0 commits behind
api15 is 9 commits ahead, 8 commits behind

Infiziert90:
> Yes, api15-rollup is 10 commits ahead, 0 commits behind api15 is 9 commits ahead, 8 commits behind

Shouldn't it still be API 15 as rollup is intended to pick up everything and combine it? So directly PRing to it seems strange

On top of that, wouldn't it lead to undetectable merge conflicts if someone PRed changes to the normal spi15 branch in the same file

Critical-Impact:
Yeah fair, that makes sense
Just looks odd when viewing it on GH is all

MidoriKami:
This PR seems to be including a bunch of random non-window things as well, I suspect its grabbed the rollup branch stuff and has incorporated it into the PR directly.

Haselnussbomber:
> This PR seems to be including a bunch of random non-window things as well, I suspect its grabbed the rollup branch stuff and has incorporated it into the PR directly.

Because it includes the changes from the api15-rollup branch and is now comparing it to the api15 branch.
Should solve itself when the rollup PR is merged.

Critical-Impact:
Fixed, now properly based on api15

PR #2406, opened 2025-09-08
Use long running task for plugin ctor

Soreepeong:


Critical-Impact (approved):
This does what the old code was doing bar the thr.Join() which could cause a deadlock under certain conditions(an exception is thrown by a loading plugin)
@goaaats Happy to merge this?

Issue #2409, opened 2025-09-11
New TerritoryChanged event breaks detecting instance and room changes


Abyeon:
Recently the ``ClientState.TerritoryChanged`` event was changed to trigger only when the ``TerritoryID`` is not the same. While the new behavior is accurate to the name of the event, it does break plugins that relied on it to detect instance changes (i.e. Where Am I Again, Hunt Helper), and changing rooms within an estate (i.e. my plugin Racingway).

It would be convenient if there were an event introduced that reflected the original behavior of ``TerritoryChanged``.

Haselnussbomber:
I made PR #2410 which adds a new `PublicInstanceId` property and new events `MapChanged` and `PublicInstanceChanged` in order to address this issue. Would that be sufficient?

Abyeon:
That sounds like it should work. Does MapChanged or PublicInstanceChanged fire when switching rooms within an estate/apartment?

Thank you for the hard work Hasel 💜

Haselnussbomber:
> Does MapChanged or PublicInstanceChanged fire when switching rooms within an estate/apartment?

MapChanged does, yes.
PublicInstanceChanged is for when they add more instances to a zone after patch to reduce congestion.

Abyeon:
Perfect! Thanks a ton! I can close this whenever I get on my computer

Abyeon:
Looks like MapChanged wouldn't fire in the case where you swap rooms from inside a room, since the MapID would be the same. Unless I misunderstood. Sorry I didn't catch that earlier

Haselnussbomber:
It does change from room to room. I tested it.

https://github.com/user-attachments/assets/f97a9c98-3fde-4a61-9e67-3bc9015d0002

This is AgentMap's CurrentMapId field. The field from GameMain doesn't change.
And yes, it also works when the Map is closed.^^

Abyeon:
Oh, sorry there has been a bit of a miscommunication. I meant specifically chambers / apartment rooms. Not the floor of a house.

Abyeon:
Here, I attached an example of what I mean using your PR

https://github.com/user-attachments/assets/c9ca64de-c0db-49d2-bd6d-3d584895d713

MidoriKami:
That seems like a really specialized case where you would need to know that changed, do you have a justification for why a room changed event would be necessary in Dalamud? Are there multiple plugins with this usecase?

Haselnussbomber:
[...]

PR #2410, opened 2025-09-12
Update ClientState

Haselnussbomber:
To address issue #2409, this PR adds the following to ClientState:

- A new `ZoneInit` event, fired when the game handles the ZoneInit packet. This essentially restores the previous `TerritoryChanged` behavior, but with additional info.
- A new `MapIdChanged` event, fired whenever `MapId` changes.
- A new `Instance` property, returning the current zone instance used to reduce congestion.
- A new `InstanceChanged` event, fired whenever `Instance` changes.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -15,7 +15,6 @@ using Dalamud.Utility;
 
 using FFXIVClientStructs.FFXIV.Application.Network;
 using FFXIVClientStructs.FFXIV.Client.Game;
-using FFXIVClientStructs.FFXIV.Client.Game.Event;
 using FFXIVClientStructs.FFXIV.Client.Game.UI;
 using FFXIVClientStructs.FFXIV.Client.Network;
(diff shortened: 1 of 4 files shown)


Dalamud/Game/ClientState/ClientState.cs
```diff
@@ -428,6 +486,8 @@ internal ClientStatePluginScoped()
     void IInternalDisposableService.DisposeService()
     {
         this.clientStateService.TerritoryChanged -= this.TerritoryChangedForward;
+        this.clientStateService.MapChanged -= this.MapChangedForward;
+        this.clientStateService.PublicInstanceChanged -= this.PublicInstanceChangedForward;
```
MidoriKami: These need corresponding `this.clientStateService.MapChanged = null`
statements below to ensure that the objects get GC'd, we had an issue with this in the past which is why we always try to set the delegates to null afterwards.

Haselnussbomber:
Added a `ZoneInit` event to bring back the old behavior of `TerritoryChanged` with some more info.
Also I've reworked the properties to raise events when their values change, similar to how `INotifyPropertyChanged` works.

PR #2411, opened 2025-09-20
Update SeStringEvaluator, Payload, AutoTranslatePayload

Haselnussbomber:
- [`bf4fc786`](bf4fc786) The auto translate code of the fixed macro parser didn't consider multiple `col-` entries in the LookupTable of the Completion sheet, as seen on RowId 2213 (TextCommand) for example. If the first column is empty (in the case of TextCommand, that is the column Alias), it takes the second one (in the case of TextCommand, that is the column Command).
- [`327ebf3b`](327ebf3b) We noticed in the updated `AutoTranslatePayload` (later down this list) that the `endOfStream` parameter of `DecodeImpl` is 1 byte too short. Further proof in `RawPayload` where it gets corrected.
- [`93a44842`](93a44842) Optimized `Payload.DecodeChunk` by skipping to the end of the payload instead of reading the remaining bytes (prevents allocation).
- [`3f037e5d`](3f037e5d) Updated `AutoTranslatePayload` to pass-through the original packet when decoded and encoded again, and to use SeStringEvaluator to get the properly evaluated text.

### DIFF
--- a/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
+++ b/Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
@@ -1629,7 +1629,9 @@ internal class SeStringEvaluator : IServiceType, ISeStringEvaluator
             return true;
 
         var isNoun = false;
-        var col = 0;
+
+        Span<int> cols = stackalloc int[2];
+        cols.Clear();
 
         if (ranges.StartsWith("noun"))
         {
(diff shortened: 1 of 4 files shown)


Dalamud/Game/Text/Evaluator/SeStringEvaluator.cs
```diff
@@ -1629,19 +1629,34 @@ private bool TryResolveFixedAutoTranslation(
             return true;
 
         var isNoun = false;
-        var col = 0;
+
+        Span<int> cols = stackalloc int[2];
```
wolfcomp: I know currently the max size col is 2, but wouldn't it be an idea instead of span to be able to use an expanding list to never have to worry about this again?

Haselnussbomber: I was avoiding the allocation for `List<int>` here and I wanted to get it done and working, without thinking about it more. I don't think we need more than the 2 entries the game processes for now.

wolfcomp: Could allocate an array of n lengths dependent on what the n count of `col-` is with something like

```cs
int count;
int index;

while((index = ranges.IndexOf("col-", index)) != -1)
{
    count++;
    index += 4;
}
Span<int> cols = new int[count];
```

Haselnussbomber: I just pushed an update and bumped it to 8. That should be enough for now. If not, that's a problem for the future.

goaaats (approved):

PR #2418, opened 2025-09-29
fix: dont nag maintainers when nothing needs to be done

KazWolfe:


KazWolfe:
tested: https://github.com/goatcorp/Dalamud/actions/runs/18111822122/

PR #2420, opened 2025-10-01
Add events based on AgentUpdateFlag

Haselnussbomber:
The [AgentUpdateFlags](https://github.com/aers/FFXIVClientStructs/blob/69cc320/FFXIVClientStructs/FFXIV/Client/UI/RaptureAtkModule.cs#L180) are used to notify agents that something has changed and that they should update their addons.

This PR adds the following event to `IGameGui`:

```cs
/// <summary>
/// Fired when the game sets one or more <see cref="AgentUpdateFlag"/> values,
/// used by agents to conditionally update their addons.
/// </summary>
event Action<AgentUpdateFlag> AgentUpdate;
```

The `NameplateUpdate` flag was not included, because the name is somewhat incorrect. It is checked for a specific thing in the nameplate code, but at this time I'm not exactly sure what it is used for.

### DIFF
--- a/Dalamud/Game/Gui/AgentUpdateFlag.cs
+++ b/Dalamud/Game/Gui/AgentUpdateFlag.cs
@@ -17,8 +17,8 @@ public enum AgentUpdateFlag : byte
     /// <summary> Set when the RetainerMarket inventory has been updated. </summary>
     RetainerMarketInventoryUpdate = 1 << 2,
 
-    /// <summary> Unknown use case. </summary>
-    NameplateUpdate = 1 << 3,
+    // /// <summary> Unknown use case. </summary>
+    // NameplateUpdate = 1 << 3,
 
     /// <summary> Set when the player unlocked collectibles, contents or systems. </summary>
     UnlocksUpdate = 1 << 4,


goaaats:
I'm not that knowledgeable about how they set this up, are these events "untyped" by default? Do they not carry any data or anything like that?

I assume it's still useful to people regardless, it just struck me as odd that there is an event that is "anything at all has changed, go figure out what it is". Or would you use that to, for example, rerender the entire inventory screen? 

Maybe these should be a single event, OnAgentUpdate, that carries the update flag - or have a comment detailing what they are actually coming from. The former would also let plugins be more flexible about what they want to handle. 

But yeah, as I said, don't know too much about use-cases so maybe I'm super off here and feel free to tell me I'm wrong.

Haselnussbomber:
They are just a flag in an enum, so they do not carry any data, yeah.
It's exactly as you said, a something has changed info for the agents to go and update their addons.
I can change it to a single event, but I wouldn't really know what to call it. AgentUpdate sounds a bit too generic.

goaaats:
I think it's fine like this then, maybe with a comment explaining that this is something the game emits. I can add one when merging if you haven't gotten around to it yet.

Haselnussbomber:
Alright, I changed it to a single event. I hope this is sufficient. :)

PR #2421, opened 2025-10-01
Add IUnlockState service

Haselnussbomber:
Rework of #2339 without the PlayerState wrapper.

---

This PR adds a new `IUnlockState` service that provides functions to check the unlock status of various content and collectibles.
It also provides an `Unlock` event, carrying the `RowRef` of the detected unlocked entry.

It's not feature complete, there are some missing unlockables, but it got the most important things covered.

### DIFF
--- a/Dalamud/Game/UnlockState/ItemActionType.cs
+++ b/Dalamud/Game/UnlockState/ItemActionType.cs
@@ -2,8 +2,6 @@ using Lumina.Excel.Sheets;
 
 namespace Dalamud.Game.UnlockState;
 
-// TODO: Switch to FFXIVClientStructs.FFXIV.Client.Enums.ItemActionType.
-
 /// <summary>
 /// Enum for <see cref="ItemAction.Type"/>.
 /// </summary>


Dalamud/Plugin/Services/IUnlockState.cs
```diff
@@ -0,0 +1,304 @@
...
+using Lumina.Excel.Sheets;
+
+namespace Dalamud.Plugin.Services;
+
+#pragma warning disable SA1400 // Access modifier should be declared: Interface members are public by default
+
+/// <summary>
+/// Interface for determining unlock state of various content in the game.
+/// </summary>
+public interface IUnlockState
```
KazWolfe, line 14: Consider marking this `[Experimental]` if you still expect API changes.

Haselnussbomber: Yeah, it's a good idea nonetheless, because something could be wrong after all.

goaaats (approved):
Very clean, thanks!

PR #2422, opened 2025-10-05
Add IPlayerState service

Haselnussbomber:
Rework of #2339 without the IUnlockState service.

---

I often see plugin devs rely on the LocalPlayer's GameObject for simple lookups like Name, Current/Home World, ClassJob, Level and so on. I've never liked seeing that, especially since the GameObject might not be available when a plugin wants to access this information (for example while logging in/switching zones), even though the information is actually statically available through other means (PlayerState, AgentLobby etc.).
So I thought it would be useful if Dalamud provides a proper `IPlayerState` service/wrapper that contains all these things and provides more information that the GameObject would not be able to provide, like experience for any of the players ClassJobs, as well as Desynthesis levels and attribute values. This is obviously just a starting point to adding more data in the future.

For this, I also moved `IClientState.LocalPlayer` to `IObjectTable.LocalPlayer`, where it belongs, and `IClientState.LocalContentId` to `IPlayerState.ContentId`. I think that's reasonable and the obolete message points out the `IPlayerState` services existence.

### DIFF
--- a/Dalamud/Game/Player/PlayerState.cs
+++ b/Dalamud/Game/Player/PlayerState.cs
@@ -13,7 +13,7 @@ using Lumina.Excel.Sheets;
 using CSPlayerState = FFXIVClientStructs.FFXIV.Client.Game.UI.PlayerState;
 using GrandCompany = Lumina.Excel.Sheets.GrandCompany;
 
-namespace Dalamud.Game.PlayerState;
+namespace Dalamud.Game.Player;
 
 /// <summary>
 /// This class contains the PlayerState wrappers.
--- a/Dalamud/Plugin/Services/IPlayerState.cs
+++ b/Dalamud/Plugin/Services/IPlayerState.cs
@@ -1,6 +1,6 @@
 using System.Collections.Generic;
 
-using Dalamud.Game.PlayerState;
+using Dalamud.Game.Player;
 
 using Lumina.Excel;
 using Lumina.Excel.Sheets;
--- a/Dalamud/Game/ClientState/Buddy/BuddyList.cs
+++ b/Dalamud/Game/ClientState/Buddy/BuddyList.cs
@@ -2,11 +2,13 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Runtime.CompilerServices;
 
+using Dalamud.Game.Player;
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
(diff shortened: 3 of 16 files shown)


Dalamud/Game/Player/PlayerState.cs
```diff
@@ -0,0 +1,227 @@
...
+
+namespace Dalamud.Game.PlayerState;
+
+/// <summary>
+/// This class contains the PlayerState wrappers.
+/// </summary>
+[PluginInterface]
+[ServiceManager.EarlyLoadedService]
+[ResolveVia<IPlayerState>]
+internal unsafe class PlayerState : IServiceType, IPlayerState
```
goaaats, line 24: This is VERY nitpicky, but maybe we could have this in a `Player` namespace instead? I kind of dislike having a namespace and classname be the same, since you have to have `PlayerState.PlayerState` everywhere.
I know ClientState is the same, I would like to change it there too.

Dalamud/Plugin/Services/IPlayerState.cs
```diff
@@ -0,0 +1,243 @@
...
+using Lumina.Excel.Sheets;
+
+namespace Dalamud.Plugin.Services;
+
+#pragma warning disable SA1400 // Access modifier should be declared: Interface members are public by default
+
+/// <summary>
+/// Interface for determining the players state.
+/// </summary>
+public interface IPlayerState
```
goaaats, line 15: IMO, we should either make it very clear here that none of the values returned by anything in this class make sense if `!IsLoaded`, or have this actually not be the service - and have a "intermediate" service that returns an `IPlayerState` if it is valid.

I know this mirrors the game quite closely, so I'm not sure what would be best. The second one seems a bit nicer to me, since it behaves the same way as LocalPlayer does - returns null if you cannot get a reference that makes any sense.

Haselnussbomber: `IsLoaded` is set to `true` when the PlayerSetup packet was received and processed, and set to `false` when you log out.
So I'm not sure if this is even that big of a deal.
The reason I added this is that it prevents reading garbage data when you log out, and instead return `default` for everything.

Dalamud/Game/Player/PlayerState.cs
Haselnussbomber: Done!

goaaats (approved):

PR #2431, opened 2025-10-12
Plugin-registerable self tests

grittyfrog:
[plugin-self-test.webm](https://github.com/user-attachments/assets/e3288dc8-ac94-477e-9930-549dc73347b6)

(apologies for the terrible framerate)

The goal of this change is to let plugins register their own self-tests. 

We do this through the `ISelfTestRegistry` interface. For a plugin it would look like this:

```csharp
[PluginService]
public ISelfTestRegistry SelfTestRegistry

// Somewhere that gets called by your plugin
SelfTestRegistry.RegisterTestSteps([
  new MySelfTestStep(),
  new MyOtherSelfTestStep()
])
```

Where `MySelfTest` and `MyOtherSelfTest` are instances of the existing `ISelfTestStep` interface.

The biggest changes are to `SelfTestWindow` and the introduction of `SelfTestWithResults`. I wanted to make sure test state wasn't lost when changing the dropdown state and I was finding it a bit annoying to work with the Dictionary now that we can't just rely on the index of the item.

To fix this I moved all the "test run" state into `SelfTestWithResults`, most of the changes to `SelfTestWindow` are derived from that, other then the addition of the combo box.

The documentation for this service is a bit sparse, but I wanted to put it up for review first before I invest a bunch of time making nice documentation. 

I'm keen to hear if we think this is useful or if any changes are needed.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/SelfTest/SelfTestWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/SelfTest/SelfTestWindow.cs
@@ -6,11 +6,13 @@ using System.Numerics;
 using Dalamud.Bindings.ImGui;
 using Dalamud.Interface.Colors;
 using Dalamud.Interface.Components;
-using Dalamud.Interface.Internal.Windows.SelfTest.Steps;
 using Dalamud.Interface.Utility;
 using Dalamud.Interface.Utility.Raii;
 using Dalamud.Interface.Windowing;
 using Dalamud.Logging.Internal;
+using Dalamud.Plugin.SelfTest;
+using Dalamud.Plugin.SelfTest.Internal;
+using Dalamud.Utility;
 using Lumina.Excel.Sheets;
 
 namespace Dalamud.Interface.Internal.Windows.SelfTest;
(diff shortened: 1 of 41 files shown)


grittyfrog:
I'll address the feedback tomorrow, thanks for taking a look! :slightly_smiling_face:

grittyfrog:
Updated the PR in response to all the feedback:

- When a plugin is unloaded the `SelfTestWithResults` class now disposes of the `ISelfTestStep` reference, this should prevent any errant plugin derived class references from sticking around
- When a plugin is unloaded the self test group is now marked as "Unloaded" in the window. We show an icon and unload all the tests, but it remains selected and available in the dropdown
- When a plugin is reloaded the self-tests will automatically repopulate
- Added docstrings to everything and an example to `ISelfTestRegistry` (since stylecop complained)

There was also a ton of stylecop errors in the previous PR which I have fixed. Please let me know if there's any other issues with this change :slightly_smiling_face:

goaaats:
Thanks, looks great!

PR #2434, opened 2025-10-20
Simpler plugin service dependency injection

Haselnussbomber:
`IDalamudPluginInterface` has methods to `Create` an object and to popualte it with services, or to `Inject` services into an existing object, but no way to get the services directly from it. This PR makes it implement `IServiceProvider`, adding a `object? GetService(Type serviceType)` method.

Additionally, this PR introduces a new `IDalamudService` marker interface, which all Dalamud services implement. This allows them to be discovered via reflection and easily registered in an `IServiceCollection` (from the `Microsoft.Extensions.DependencyInjection` package).

### DIFF
--- a/Dalamud/Plugin/Services/IPlayerState.cs
+++ b/Dalamud/Plugin/Services/IPlayerState.cs
@@ -12,7 +12,7 @@ namespace Dalamud.Plugin.Services;
 /// <summary>
 /// Interface for determining the players state.
 /// </summary>
-public interface IPlayerState
+public interface IPlayerState : IDalamudService
 {
     /// <summary>
     /// Gets a value indicating whether the local players data is loaded.


goaaats:
Haven't read through the entire thing yet, but the IoC should throw if a plugin-exposed service resolves through an interface that does not inherit from IDalamudService when discovering services. Otherwise we'll forget and get sassy messages when people can't resolve new services 😁

Haselnussbomber:
Should that also be in DEBUG only?
I tried adding an additional Debug.Assert, but that makes it not even start.
If it would throw, then we at least get an exception in the log. But if it throws, the process just gets stuck.

wolfcomp:
If throw makes it not start, I would say have an error log as a stop gap. Unless someone writes an analyzer to make Dalamud fail to compile as another method.

PR #2441, opened 2025-10-28
Disable main thread check when framework is shutting down

Haselnussbomber:
There isn't really a need to check for it anyways and it just blows up when plugins try to access something that is guarded by this during disposal when the game is shutting down.


Haselnussbomber:
goat in #dalamud-dev:
> That isn't actually thread-safe

So I'm closing this.

Plugins are better off checking `Framework.IsFrameworkUnloading` in their Dispose function.

PR #1841, opened 2024-06-15
New auto-update UX

goaaats:
Not my finest work but it'll do. One to-do left about opt-ins, but I'll try to fix that up later on.

PR #2443, opened 2025-10-29
Fix systems not working on the Free Trial

Haselnussbomber:
When playing via the Free Trial the condition `OnFreeTrial` is always active.
This PR adds this as exception for the plugin auto updates and for the idle detection.

### DIFF
--- a/Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
+++ b/Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
@@ -491,14 +491,10 @@ internal class AutoUpdateManager : IServiceType
 
     private bool CanUpdateOrNag()
     {
-        var condition = Service<Condition>.Get();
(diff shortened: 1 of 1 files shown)


Dalamud/Plugin/Internal/AutoUpdate/AutoUpdateManager.cs
```diff
@@ -497,7 +497,8 @@ private bool CanUpdateOrNag()
             condition.OnlyAny(ConditionFlag.NormalConditions,
```
goaaats: I wonder why this doesn't just use the IsClientIdle() in ClientState? I don't remember adding that but it does look exactly the same

Haselnussbomber: It was added with https://github.com/goatcorp/Dalamud/pull/1842, which was opened on the same day as https://github.com/goatcorp/Dalamud/pull/1841.
I'll change it to use that instead.

PR #2447, opened 2025-11-08
[master] Update ClientStructs

bleatbot:


Haselnussbomber:
Shows up as breaking because I replaced two byte arrays with the corresponding fields from the sheet in https://github.com/aers/FFXIVClientStructs/pull/1639.
The underlying type of the CraftFlags enum (which isn’t marked as breaking change by apicompat) was changes from uint to byte in the same PR, as the field (which was at the wrong offset anyway) would overlap with the WKS fields.

PR #2448, opened 2025-11-08
Fix KeyNotFoundException in DtrBar.RemoveNode

Haselnussbomber:
Not sure why this happens, but let's make sure it doesn't throw at least.


goaaats:
Could you maybe add a log if it's not too much trouble? This means something wasn't bookkept properly as far as I can tell, so we should check it out, no?

Haselnussbomber:
I can add a warning, yeah, but I have a feeling that this is going to crash anyways in the code that comes after.

goaaats:
Not keeping the lock forever is most important IMO, then whatever Ennea saw on Discord can at least not spam the log and cause high CPU usage anymore. I'd rather have it crash.

Haselnussbomber:
Fixed a potential use-after-free too by setting TextNode to null, making sure subsequent null-checks are working.

goaaats (approved):

PR #2458, opened 2025-11-13
[API14] Use struct enumerators/types

Haselnussbomber:
- In my benchmarks `yield` was 20% slower than a class/struct implementing `IEnumerator<T>`. The difference is probably negligible and .NET 10 seems to optimize the classes away now, as they didn't allocate anything, but still, I've added struct-based enumerators to the following services:
  - AetheryteList, BuddyList, FateTable, PartyList, StatusList, ObjectTable
- The following data-classes were converted to readonly structs, implementing `IEquatable<T>`, because they are mostly short-lived wrappers around pointers or another struct:
  - AetheryteEntry, BuddyMember, Fate, PartyMember, Status
- Status now has an interface `IStatus`.
- The unused `StatusEffect` struct was removed.
- `Fate.IsValid` was removed as it only checked for the local ContentId and wasn't related to Fates at all.
- Self-tests are passing:
Passing Self-tests


goaaats (approved):
Yeah, not sure about the perf benefits, but it's certainly cleaner :)

PR #2462, opened 2025-11-16
feat: Dalamud RPC service

KazWolfe:
A draft for a simple RPC service for Dalamud. Enables use of Dalamud URIs, to be added later on the XL side (see goatcorp/FFXIVQuickLauncher#1779).

At the moment, this RPC system does not handle type enforcement or anything, instead just preferring to deal with whatever comes over the wire, as it comes over the wire. This is by design for now - it adds unnecessary complication, and also would require publishing a common set of contracts. Given that this is basically _just_ for URIs, this seems excessive.

### DIFF
--- a/Dalamud/Networking/Pipes/Api/PluginLinkHandler.cs
+++ b/Dalamud/Networking/Pipes/Api/PluginLinkHandler.cs
@@ -6,6 +6,7 @@ using Dalamud.IoC.Internal;
 using Dalamud.Networking.Pipes.Internal;
 using Dalamud.Plugin.Internal.Types;
 using Dalamud.Plugin.Services;
+#pragma warning disable DAL_RPC
 
 namespace Dalamud.Networking.Pipes.Api;
 
--- a/Dalamud/Networking/Pipes/Internal/ClientHelloService.cs
+++ b/Dalamud/Networking/Pipes/Internal/ClientHelloService.cs
@@ -1,9 +1,12 @@
 ﻿using System.Threading.Tasks;
 
+using Dalamud.Data;
 using Dalamud.Game;
 using Dalamud.Game.ClientState;
 using Dalamud.Networking.Pipes.Rpc;
 using Dalamud.Utility;
+
+using Lumina.Excel.Sheets;
 
 namespace Dalamud.Networking.Pipes.Internal;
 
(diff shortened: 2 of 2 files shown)


Dalamud/Networking/Pipes/Api/PluginLinkHandler.cs
```diff
@@ -0,0 +1,53 @@
...
+    public void DisposeService()
+    {
+        this.OnUriReceived = null;
+        this.linkHandler.Unregister("plugin", this.HandleUri);
+    }
+
+    private void HandleUri(DalamudUri uri)
+    {
+        var target = uri.Path.Split("/").FirstOrDefault();
+        if (target == null || !string.Equals(target, this.localPlugin.InternalName, StringComparison.OrdinalIgnoreCase))
```
goaaats: Maybe this should use something like `ConsoleManagerPluginUtil.GetSanitizedNamespaceName()`? Should we move that to somewhere more generic?

Dalamud/Networking/Pipes/Internal/ClientHelloService.cs
```diff
@@ -0,0 +1,94 @@
...
+        var framework = await Service<Framework>.GetAsync();
+        var dalamud = await Service<Dalamud>.GetAsync();
+        var clientState = await Service<ClientState>.GetAsync();
+
+        var response = await framework.RunOnFrameworkThread(() => new ClientHelloResponse
+        {
+            ApiVersion = "1.0",
+            DalamudVersion = Util.GetScmVersion(),
+            GameVersion = dalamud.StartInfo.GameVersion?.ToString() ?? "Unknown",
+            PlayerName = clientState.IsLoggedIn ? clientState.LocalPlayer?.Name.ToString() ?? "Unknown" : null,
```
goaaats: What is this for? I think we should be careful about information we expose here in general. Might not matter too much in the end though, I guess, since it's not HTTP.

Dalamud/Networking/Pipes/Internal/LinkHandlerService.cs
```diff
@@ -0,0 +1,129 @@
...
+            return;
+
+        try
+        {
+            var du = DalamudUri.FromUri(uri);
+            this.Dispatch(du);
+        }
+        catch (Exception)
+        {
+            // swallow parse errors; clients shouldn't crash the host
```
goaaats: Log? Probably means that someone messed up and they need to figure out why

Dalamud/Networking/Pipes/Rpc/PipeRpcHost.cs
```diff
@@ -0,0 +1,167 @@
...
+    /// </summary>
+    /// <param name="name">The name to add.</param>
+    /// <param name="handler">The delegate that acts as the handler.</param>
+    public void AddMethod(string name, Delegate handler) => this.registry.AddMethod(name, handler);
+
+    /// <summary>Starts accepting client connections.</summary>
+    public void Start()
+    {
+        if (this.acceptLoopTask != null) return;
+        this.acceptLoopTask = Task.Run(this.AcceptLoopAsync);
```
goaaats: `TaskCreationOptions.LongRunning` or just a new thread?

Dalamud/Plugin/Services/IPluginLinkHandler.cs
```diff
@@ -0,0 +1,20 @@
+﻿using Dalamud.Networking.Pipes;
+
+namespace Dalamud.Plugin.Services;
+
+/// <summary>
+/// A service to allow plugins to subscribe to dalamud:// URIs targeting them. Plugins will receive any URI sent to the
+/// <code>dalamud://plugin/{PLUGIN_INTERNAL_NAME}/...</code> namespace.
+/// </summary>
+public interface IPluginLinkHandler
```
goaaats, line 12: `[Experimental]`?

Dalamud/Networking/Pipes/Internal/ClientHelloService.cs
KazWolfe: The intent of `ClientHelloService` is to allow for RPC clients to figure out who they're talking to.

For cases like multiboxed clients, we need to be able to allow the user to choose which game instance should handle a URI. It'd probably be helpful to give some indication in that case such that the user doesn't need to guess.

goaaats: I think that makes sense, didn't consider the multiboxing case.

Dalamud/Networking/Pipes/Rpc/PipeRpcHost.cs
KazWolfe: Either or, LongRunning is probably more ideal here?

Dalamud/Networking/Pipes/Internal/LinkHandlerService.cs
KazWolfe: Swapped to `#InvokeSafely()`.

Dalamud/Networking/Pipes/Api/PluginLinkHandler.cs
KazWolfe: Using that method for now, and yes, probably should be moved.

Dalamud/Plugin/Services/IPluginLinkHandler.cs
KazWolfe: Set for finalization in 7.41.

PR #2468, opened 2025-11-18
feat: Add unix sockets

KazWolfe:
- Unix sockets run parallel to Named Pipes
  - Named Pipes will only run on non-Wine
  - If the game crashes, the next run will clean up an orphaned socket.
- Restructure RPC to be a bit tidier

### DIFF
--- a/Dalamud/Networking/Rpc/Service/ClientHelloService.cs
+++ b/Dalamud/Networking/Rpc/Service/ClientHelloService.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System.Diagnostics;
+using System.Threading.Tasks;
 
 using Dalamud.Data;
 using Dalamud.Game;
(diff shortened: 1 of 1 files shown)


KazWolfe:
Note that AF_UNIX sockets won't be a thing until https://gitlab.winehq.org/wine/wine/-/merge_requests/7650 lands, but there's no reason to not set up AF_UNIX as well.

It might make sense to replace named pipes entirely, but that's also something we can ultimately decide later. Named pipes would - I think - be better were we to ever set up proper sandboxing, for example.

Dalamud/Networking/Rpc/UnixSocketUtil.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Threading.Tasks;
+
+using Serilog;
+
+namespace Dalamud.Utility;
+
+/// <summary>
+/// A set of utilities to help manage Unix sockets.
+/// </summary>
+internal static class UnixSocketUtil
```
goaaats: This file should probably be in the Networking/Rpc folder since it seems pretty specific to RPCs, too much stuff in this one already

KazWolfe:
> Why do we have the named pipe one if the unix one works?

I wrote the named pipe one first. Nothing really else to it. 

The best move is _probably_ doing unix sockets only, though there isn't really a cost to having both. It might be worth getting someone other than me to implement a client (XL?) to see which they like better, and then we can delete the opposite one. Or just decide we only care about UDS.

## Named Pipe Advantages

* Supports more capable authentication, richer ACLs.
* Doesn't put temp files on disk that need to be cleaned up.
* (I think?) Better AppContainer support, though we might need to invert the pipe owner anyways if we did this.
* Significantly more mature stack-wise

## UNIX Domain Socket Advantages

* Wine support just around the corner
* Better performance than Named Pipes
* Fewer things can go wrong, systems-speaking.

goaaats (approved):

PR #2471, opened 2025-11-26
Refactor Addon Lifecycle

MidoriKami:
Reworked AddonLifecycle to use a new approach of replacing all addons VirtualTables on load, and using that to redirect to dalamud controlled event handlers.

Adds Pre/Post Open, Close, Show, and Hide events. (With the ability to easily add more events)

This new approach is much simpler, much more flexible, and is no longer dependent on any custom sigs.

Only one hook is used, AtkUnitBase.Initialize, and all table redirects use ClientStructs definitions meaning there is no direct maintenance burden in Dalamud when a patch rolls around.

Added a generic `GenericAddonArgs` class for use with events that don't have any meaningful args, I recommend marking AddonDrawArgs and AddonUpdateArgs as obsolete and replaced with `GenericAddonArgs`. As that would be a breaking change, it is not included in this PR.

Removed PooledArgs, as with this new approach theres no longer any need to pre-allocate arg objects for sharing, as each addon will get their own set of args to use. There is no longer runtime heap allocation pressure, due to allocating the event args per-addon when the table is allocated.

And finally, removed AddonSetupHook, and CallHook, as these two classes were only used for AddonLifecycle and are now effectively obsolete.

### DIFF
--- a/Dalamud/Configuration/PluginConfigurations.cs
+++ b/Dalamud/Configuration/PluginConfigurations.cs
@@ -11,7 +11,7 @@ namespace Dalamud.Configuration;
 /// <summary>
 /// Configuration to store settings for a dalamud plugin.
 /// </summary>
-[Api13ToDo("Make this a service. We need to be able to dispose it reliably to write configs asynchronously. Maybe also let people write files with vfs.")]
+[Api14ToDo("Make this a service. We need to be able to dispose it reliably to write configs asynchronously. Maybe also let people write files with vfs.")]
 public sealed class PluginConfigurations
 {
     private readonly DirectoryInfo configDirectory;
--- a/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonArgs.cs
+++ b/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonArgs.cs
@@ -34,40 +34,13 @@ public abstract class AddonArgs
     public abstract AddonArgsType Type { get; }
 
     /// <summary>
-    /// Checks if addon name matches the given span of char.
-    /// </summary>
-    /// <param name="name">The name to check.</param>
-    /// <returns>Whether it is the case.</returns>
-    internal bool IsAddon(string name)
-    {
-        if (this.Addon.IsNull)
-            return false;
-
-        if (name.Length is 0 or > 32)
-            return false;
-
-        if (string.IsNullOrEmpty(this.Addon.Name))
-            return false;
-
-        return name == this.Addon.Name;
-    }
-
-    /// <summary>
-    /// Clears this AddonArgs values.
-    /// </summary>
-    internal virtual void Clear()
-    {
-        this.addonName = null;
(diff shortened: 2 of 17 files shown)


MidoriKami:
Seems the bot is complaining about the now removed ICloneable from the ArgsObjects, these functions were never intended to be used by the users, and I strongly believe it is the correct move to remove these interfaces.

MidoriKami:
re: [c525655](https://github.com/goatcorp/Dalamud/pull/2471/commits/c525655be66778cd3d092d32f6ae5aba16c0fbea)

Event registrations are no longer required to be main-threaded, as the previous system would hook ReceiveEvent on event registration, the new system doesn't touch native state at all to add or unregister listeners.

Also improved performance by using a dictionary and early returning if the invoked event type isn't being listened for at all, additionally it'll only loop over listeners of the correct type.

When potentially hundreds of listeners are registered, looping through them all for every event is very inefficient.

Dalamud/Game/Addon/Lifecycle/AddonLifecycle.cs
```diff
@@ -61,10 +61,8 @@ void IInternalDisposableService.DisposeService()
     /// <param name="listener">The listener to register.</param>
     internal void RegisterListener(AddonLifecycleEventListener listener)
     {
-        this.framework.RunOnTick(() =>
-        {
-            this.EventListeners.Add(listener);
-        });
+        this.EventListeners.TryAdd(listener.EventType, [ listener ]);
+        this.EventListeners[listener.EventType].Add(listener);
```
Haselnussbomber: Initializing the List with `[ listener ]` will cause the first listener of an EventType to be added twice.
I recommend using a `HashSet<AddonLifecycleEventListener>` instead.

MidoriKami: Nice catch! Thanks, I'll do exactly that.

MidoriKami:
I'm happy to report that my testing shows that this refactor is successfully working both in and out of duties, and through entire UI rebuilds via Aestetician. Performance should also be significantly improved, though I have not collected any data to back that up.

MidoriKami:
Alright so this last set of changes has a couple potentially controversial things...

Firstly, on EventArgs we've had an `[Obsolete("Not intended for public construction")]` on these objects for I think 1-3 API levels, it's safe to mark these as internal.

I renamed `Api13ToDo` -> `Api14ToDo` and added a `Api15ToDo`, things that I've marked with Api15ToDo, are intended to give devs more time to update their code before they are removed/changed.

AddonReceiveEventArgs `Data` property should be renamed to `AtkEventData` as soon as possible, but you might want to push it off to Api15 and mark it obsolete instead.

AddonUpdateArgs, TimeDelta was never intended to be publicly exposed, as messing with this could cause some extremely undesired effects. I have made it private.

MidoriKami:
closing this PR to re-target it to API14, and to write a better PR description

PR #2476, opened 2025-11-27
Erroring Window Style Fix

MidoriKami:
Forces dalamuds standard style onto any window that is erroring so that the error can actually be read and interacted with.

For example, with the following test code:

```cs
    private class DemoWindow() : Window("Exception Test Window", ImGuiWindowFlags.AlwaysAutoResize) {
        public override void Draw() {
            ImGui.Text("butts");
            ImGui.PushStyleColor(ImGuiCol.WindowBg, Vector4.Zero);
            throw new Exception("Boom");
            ImGui.Text("Butts2");
        }
    }
```

What is shown is:
ffxiv_dx11_aadzrrhlOL

after this PR is merged you will see:
ffxiv_dx11_7O7X9mebiO

### DIFF
--- a/Dalamud/Interface/Windowing/Window.cs
+++ b/Dalamud/Interface/Windowing/Window.cs
@@ -57,7 +57,6 @@ public abstract class Window
 
     private bool hasError = false;
     private Exception? lastError;
-    private bool isErrorStylePushed;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="Window"/> class.
@@ -426,6 +425,7 @@ public abstract class Window
                 UIGlobals.PlaySoundEffect(this.OnOpenSfxId);
         }
 
+        var isErrorStylePushed = false;
         if (!this.hasError)
         {
             this.PreDraw();
@@ -434,7 +434,7 @@ public abstract class Window
         else
         {
             Style.StyleModelV1.DalamudStandard.Push();
-            this.isErrorStylePushed = true;
+            isErrorStylePushed = true;
         }
 
         if (this.ForceMainWindow)
@@ -697,7 +697,7 @@ public abstract class Window
         }
         else
         {
-            if (this.isErrorStylePushed)
+            if (isErrorStylePushed)
             {
                 Style.StyleModelV1.DalamudStandard.Pop();
             }


Dalamud/Interface/Windowing/Window.cs
```diff
@@ -448,10 +457,28 @@ internal void DrawInternal(WindowDrawFlags internalDrawFlags, WindowSystemPersis
         var flags = this.Flags;
 
         if (this.internalIsPinned || this.internalIsClickthrough)
-            flags |= ImGuiWindowFlags.NoMove | ImGuiWindowFlags.NoResize;
+        {
+            if (!this.hasError)
```
goaaats: I wonder if it would make more sense to force a "known good" set of flags when we have an error, always, instead of just when we are pinned or clickthrough?

MidoriKami: I think your right, do you have a recommendation of a good set to use?

MidoriKami: Would ImGuiWindowFlags.None be a good option? It would disable all special functions and give us just a normal window.

MidoriKami:
Updated to set flags to `None`
It's probably also a good idea to disable any size constraints, and uncollapse the window.

Though maybe uncollapse is not necessary because it wont draw while collapsed.

Dalamud/Interface/Windowing/Window.cs
```diff
@@ -425,8 +426,16 @@ internal void DrawInternal(WindowDrawFlags internalDrawFlags, WindowSystemPersis
...
-        this.ApplyConditionals();
+        if (!this.hasError)
+        {
+            this.PreDraw();
+            this.ApplyConditionals();
+        }
+        else
+        {
+            Style.StyleModelV1.DalamudStandard.Push();
+            this.isErrorStylePushed = true;
```
goaaats: What unsets this again? Maybe it should be a local variable?

MidoriKami: Good catch, I initially made it a member because I was going to apply it in PreDraw and clear it in PostDraw, but then realized it is best to not call PreDraw or PostDraw at all when erroring.

MidoriKami: It would be a good idea to convert the StyleModel to using ImRaii so we could use dispose pattern to push the style, and ensure there's no issues popping it correctly.

For now I'll convert to local variable, and push/pop as needed.

PR #2480, opened 2025-11-30
Improve third-party repo UX

Eauldane:
A surprisingly common complaint I keep coming across is that people have trouble adding third-party repos. The current workflow requires  pressing the "plus" button to add an empty row after pasting the last URL you want to add in. This PR changes the behaviour so that a user can simply paste in the URL and hit the save button without the unnecesary row add.

Some changes to the saving logic were needed to accomodate this, since validity checks were tied to the add button - the save button would otherwise close the window, hiding the error message if there is one. The window will stay open now if there's an error, or close as before if everything is good. 

I've tested this with changing both the repo list and adding dev DLLs and it appears to work as intended.

### DIFF
--- a/Dalamud/Interface/Internal/Windows/Settings/Widgets/ThirdRepoSettingsEntry.cs
+++ b/Dalamud/Interface/Internal/Windows/Settings/Widgets/ThirdRepoSettingsEntry.cs
@@ -269,8 +269,20 @@ internal class ThirdRepoSettingsEntry : SettingsEntry
     }
 
     private static bool ValidThirdPartyRepoUrl(string url)
-        => Uri.TryCreate(url, UriKind.Absolute, out var uriResult)
-        && (uriResult.Scheme == Uri.UriSchemeHttps || uriResult.Scheme == Uri.UriSchemeHttp);
+    {
+        if (!Uri.TryCreate(url, UriKind.Absolute, out var uriResult))
+            return false;
+
+        if (uriResult.Scheme != Uri.UriSchemeHttps && uriResult.Scheme != Uri.UriSchemeHttp)
+            return false;
+
+        if (string.IsNullOrWhiteSpace(uriResult.Host))
+            return false;
+
+        var hostNameType = Uri.CheckHostName(uriResult.Host);
+        return hostNameType != UriHostNameType.Unknown
+               && Uri.IsWellFormedUriString(url, UriKind.Absolute);
+    }
 
     private bool ValidateTempRepoInput()
     {


Dalamud/Interface/Internal/Windows/Settings/Widgets/ThirdRepoSettingsEntry.cs
```diff
@@ -269,4 +266,39 @@ public override void Draw()
...
+
+    private bool TryAddTempRepo()
+    {
+        if (string.IsNullOrWhiteSpace(this.thirdRepoTempUrl))
+            return false;
+
+        this.thirdRepoTempUrl = this.thirdRepoTempUrl.Trim();
+        if (this.thirdRepoList.Any(r => string.Equals(r.Url, this.thirdRepoTempUrl, StringComparison.InvariantCultureIgnoreCase)))
+        {
+            this.thirdRepoAddError = Loc.Localize("DalamudThirdRepoExists", "Repo already exists.");
```
goaaats, line 300: Is there a reason you are clearing the errors after a while? Seems unintuitive, since the input is still not valid after they are cleared, but there is no more feedback

Dalamud/Interface/Internal/Windows/Settings/Widgets/ThirdRepoSettingsEntry.cs
```diff
@@ -229,33 +242,17 @@ public override void Draw()
         ImGui.Text(repoNumber.ToString());
         ImGui.NextColumn();
         ImGui.SetNextItemWidth(-1);
-        ImGui.InputText("##thirdRepoUrlInput"u8, ref this.thirdRepoTempUrl, 300);
+        if (ImGui.InputText("##thirdRepoUrlInput"u8, ref this.thirdRepoTempUrl, 300))
```
goaaats, line 250: Would it not make more sense to run the validations every time this input changes? Then you don't have to add the double-check SettingsWindow and people still get immediate feedback. Just have Save() write the new repo automatically.

Eauldane: Fair point - I usually work on systems where not waiting for user confirmation before validation is frowned upon due to resource constraints or convention, so I was still in that mindset when I wrote that. Fixed in the next commit!

Eauldane:
Requested changes have been made - errors don't clear after a few seconds now, and input is validated as it happens. The latter led me to discover that the existing validation would consider an entry of just `http://` or `https://` to be valid, so I took the liberty of adjusting that to require at least a host be set as well. This could be extended to actually poll the URL and check for the content type being JSON, but that feels out of scope for this PR.

PR #2481, opened 2025-11-30
[API14] AddonLifecycle Refactor

MidoriKami:
This PR reworks how AddonLifecycle works at a fundamental level, this new approach replaces all addons virtual tables when they are loaded with a custom virtual table that redirects the calls to to this service to invoke listeners event handlers.

By replacing the virtual table entirely, we can now inform users of addon events, that are called from other parts of the game, or even from other plugins. This allows us to now include Open, Close, Show, Hide, OnMove, OnMouseOver, OnMouseOut, OnFocus events, and potentially any other AtkUnitBase virtual function that we might want to provide.

We no longer need any CallHook's, nor the very hacky custom AddonSetup hook classes, as we are attaching the redirects directly to the addons themselves.

And finally, there's no longer any need to manage ReceiveEvent hooks, as we are intercepting the VirtualTable entry itself. This has a major benefit in reducing this services complexity, as we no longer have to care about specific addons sharing a receive event function, we only redirect an event if that event is actually being sent to the target addon.

## Breaking Changes

1. Removed IClonable from AddonArgs, and derived classes

These functions were never intended to be used by plugin developers and were only included to facilitate the function of the AddonPooledArgs system, which is no longer necessary.

AddonPooledArgs was designed to reduce heap allocation pressure, as the hooked functions draw and update get called very often, and allocating objects inside each call was leading to high GC pressure. With the new system, each addon gets its own set of args allocated when the virtual table is generated, removing the need for the arg pool.

2. Removed AddonDrawArgs, AddonFinalizeArgs, and AddonUpdateArgs

These objects provide no functionality and have been removed. Additionally removed their entries from AddonArgsTypes.

Use the base class AddonArgs instead.

Note: AddonUpdateArgs did expose the update delta, however I believe this was done in error, as users should not be needing to interact with the update delta. If a user has a use case where that matters, they are advised to directly hook Update themselves.

3. Removed AddonLifecycleAddressResolver

This new approach only uses a single hook, AtkUnitBase.Initialize, this address is provided by ClientStructs directly, so there aren't any sigs that dalamud would need to scan for itself.

4. Sealed AddonArgs, and derived classes

The constructor is no longer public, and the obsolete tag has been removed. These were marked obsolete in December of 2023, I think it's been long enough to be able to make them internal.

5. Renamed `Data` to `AtkEventData` in AddonReceiveEventArgs

When this class was originally written, we didn't have a definition for the event data, so the property was just named `Data`, but now that we know its correct type, I want the property to be named the thing that people will have to cast it into to be able to use it.

#### Set all dalamud projects to Language Version 14

This was necessary to enable auto backing fields for AddonArgs to more easily and cleanly save addon pointer and save addon name.

#### API TODO
Renamed Api13ToDo to Api14ToDo

Added Api15ToDo, used to mark things that should be done but don't need to be done soon.

#### ImGui
Fixed language change preferring ReadOnlySpan over Span, causing the cast method to generate compiler errors.

### DIFF
--- a/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonRefreshArgs.cs
+++ b/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonRefreshArgs.cs
@@ -4,6 +4,7 @@ using Dalamud.Game.NativeWrapper;
 using Dalamud.Utility;
 
 using FFXIVClientStructs.FFXIV.Component.GUI;
+using FFXIVClientStructs.Interop;
 
 namespace Dalamud.Game.Addon.Lifecycle.AddonArgTypes;
 
@@ -54,7 +55,7 @@ public class AddonRefreshArgs : AddonArgs
                 AtkValuePtr ptr;
                 unsafe
                 {
-                    ptr = new AtkValuePtr((nint)this.AtkValueSpan[i].Pointer);
+                    ptr = new AtkValuePtr((nint)this.AtkValueSpan.GetPointer(i));
                 }
 
                 yield return ptr;
--- a/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonSetupArgs.cs
+++ b/Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonSetupArgs.cs
@@ -4,6 +4,7 @@ using Dalamud.Game.NativeWrapper;
 using Dalamud.Utility;
 
 using FFXIVClientStructs.FFXIV.Component.GUI;
+using FFXIVClientStructs.Interop;
 
 namespace Dalamud.Game.Addon.Lifecycle.AddonArgTypes;
 
@@ -54,7 +55,7 @@ public class AddonSetupArgs : AddonArgs
                 AtkValuePtr ptr;
                 unsafe
                 {
-                    ptr = new AtkValuePtr((nint)this.AtkValueSpan[i].Pointer);
+                    ptr = new AtkValuePtr((nint)this.AtkValueSpan.GetPointer(i));
                 }
 
                 yield return ptr;


Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonRefreshArgs.cs
```diff
@@ -31,19 +32,7 @@ public AddonRefreshArgs()
     /// <summary>
     /// Gets the AtkValues in the form of a span.
     /// </summary>
+    [Obsolete("Pending removal, unsafe to use when using custom ClientStructs")]
```
goaaats: Can you note here what the replacement is, or what else users should be doing? We shouldn't remove useful API for the three people that bring custom CS. Same for the other args classes that expose AtkValues.

Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonHideArgs.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    /// <summary>
+    /// Gets or sets the flags that the window will set when it Shows/Hides.
+    /// </summary>
+    public uint SetShowHideFlags { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether something for this event message.
+    /// </summary>
+    internal bool UnknownBool { get; set; }
```
goaaats, line 31: We shouldn't expose unknowns like this, IMO. Let's add it once we actually know what it is.

MidoriKami: Its internal, so its not a part of the API currently.

Dalamud/Game/Addon/Lifecycle/AddonArgTypes/AddonRefreshArgs.cs
MidoriKami: I'm happy to keep these here, but there has been 2 instances of people using custom CS running into issues that took them some time to trace back to these helpers.

I can either remove the obsolete notice entirely and keep them as is.

Or

I can edit the Obsolete note to: "Pending removal, use 'new Span<AtkValue>(AtkValues, (int)AtkValueCount)'"

goaaats: Could you elaborate on the specific problems it causes? I'm curious, haven't heard of it before

Haselnussbomber: You should know? It's a type conflict between the `AtkValue` type coming from Dalamuds CS and the type in a plugins custom CS. I've ran into this myself, trying to access this Span and getting an error, without understanding it for a good 10 minutes. I thought we have a unspoken rule to not expose ClientStructs types?

What we could do, to make it safe, would be to copy the `IEnumerable<AtkValuePtr> AtkValues` from https://github.com/goatcorp/Dalamud/blob/master/Dalamud/Game/NativeWrapper/AtkUnitBasePtr.cs#L119.

goaaats: Ah okay, I remember now. I'm trying to figure out what the impact of removing this would be, and what the effort would be to make it work. Do you think it's fine to leave it as-is for now until we can try a solution for the next bump?

MidoriKami: I really like Hasels idea of adding the IEnumerable<AtkValuePtr>, I will adjust the code shortly.

Haselnussbomber: The impact... well, code using it needs to be changed.
In my opinion, Dalamud should prefer to use `AtkValuePtr` over `AtkValue`. It's much safer to use `AtkValuePtr.TryGet<T>` for getting a value due to the type safety, and for writing to an AtkValue you'd need to get the pointer anyhow, which is not exposed via the Span (you'd use the `GetPointer` extension from CS).
`AtkValuePtr` already has an `Address` property that can be easily cast to `AtkValue*` if you really need to.
I think changing it is the best option. It's a nicer API.

PR #2487, opened 2025-12-05
[API14] Update Nuke

Haselnussbomber:
Only small changes here. CI still builds everything.^^


goaaats:
Looks like CI warnings are broken on api14, but it seems like they are broken without this as well :(

PR #2493, opened 2025-12-07
[API14] Update packages

Haselnussbomber:
Updated most nuget packages.
Build works, injection works, logs work, unit tests successful.

Removed unused packages:

- System.Collections.Immutable
- System.Drawing.Common
- System.Resources.Extensions

Not updated:

- Lumina to 7.1.0 needs Lumina.Excel submodule bump
- HexaGen.Runtime had a bunch of refactors, so I didn't touch it
- Reloaded.Memory and Reloaded.Memory.Buffers require some changes in Injector due to removed classes. (The rust rework happened in v3.0.0 of Reloaded.Memory.Buffers.)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,35 +6,35 @@
 
   <ItemGroup>
     <!-- Analyzers -->
-    <PackageVersion Include="Microsoft.CodeAnalysis.BannedApiAnalyzers" Version="3.3.4" />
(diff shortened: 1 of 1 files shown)


goaaats:
I consider Reloaded effectively pinned, at least for now. At this point we have so many changes in there that I don't think we will ever do the upgrade to the Rust versions if we don't have to do major rewrites to keep things working, the only reason we still have the official packages referenced is because the injector needs cross-process buffers for inject mode.
I removed support for that in the version Dalamud uses because it was buggy and causing problems with in-process buffers.

PR #2503, opened 2025-12-12
feat: Refactor CJK font resources into a TTC collection

Loskh:
Currently, Dalamud does not switch CJK font family according to the UI language, causing certain characters to display incorrectly under Simplified Chinese and Traditional Chinese UIs.
Need to manually select Noto Sans in the settings , as the game's font is based on a Japanese typeface.
Additionally, this change also simplifies maintenance for both the Simplified Chinese and Traditional Chinese server branches and makes it easier for plugin developers to support multiple languages.
Before Fix:
PixPin_2025-12-12_16-35-49
PixPin_2025-12-12_16-36-28
After Fix:


Soreepeong:
If this is to be merged after an API version bump, then aliases with Obsolete attributes need to be added to DalamudAsset enum.

Loskh:
> If this is to be merged after an API version bump, then aliases with Obsolete attributes need to be added to DalamudAsset enum.

Done! Added the [Obsolete] aliases to DalamudAsset as suggested.

Dalamud/Interface/FontIdentifier/IFontSpec.cs
```diff
@@ -24,6 +24,11 @@ public interface IFontSpec
     /// </summary>
     float LineHeightPx { get; }
 
+    /// <summary>
+    /// Gets the font no.
+    /// </summary>
+    int FontNo { get; }
```
goaaats, line 48: Could you describe what this means? Is it the set of CJK glyphs used? Why is it an int, and not an enum, and can multiple sets of CJK glyphs be used in a single font spec?

Soreepeong: This is for the font face index in a TrueType Collection file; it effectively is a concatenated multiple TTF files.

goaaats: Oh, interesting, thanks. Wasn't aware that that is a thing, would be nice if we could note it here just in case someone comes across this in the future.

PR #2519, opened 2025-12-18
Add git hash/scm version properties to DalamudVersionInfo

Critical-Impact:


goaaats:
Could you maybe leave these internal and add the ones we need to IDalamudVersionInfo instead? That one is directly exposed, then plugins don't need to rely on random statics from our end

Critical-Impact:
Should be good now

goaaats (approved):

PR #2522, opened 2025-12-18
fix: EventItem name for Japanese client.

Loskh:
ItemPayloadSelfTest will fail in a Japanese client.
Only English, German, and French have a Name field.For other languages, the Name is an empty string, and the Singular field should be used instead.
image


Haselnussbomber:
Thanks, I didn't think of this.
For reference, the game uses this bitmask to check this in "GetEventItemName" (CS naming): `0b0001_1010_0010`
Using Luminas [Language](https://github.com/NotAdam/Lumina/blob/master/src/Lumina/Data/Language.cs#L5) enum, this would be true for the following languages:
- Japanese (1)
- ChineseSimplified (5)
- Korean (7)
- TraditionalChinese (8)

PR #2525, opened 2025-12-18
[master] Update ClientStructs

bleatbot:


Haselnussbomber:
There is a mistake in this. PlaceFieldMarker should have Vector3* as arg. I already fixed it - pull again.

PR #2526, opened 2025-12-18
Remove RPC Entirely

KazWolfe:
Removes the RPC system, as it's breaking plugin builds and creating a mess:

```
0>Microsoft.Common.CurrentVersion.targets(2437,5): Warning MSB3277 : Found conflicts between different versions of "WindowsBase" that could not be resolved.
There was a conflict between "WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" and "WindowsBase, Version=8.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35".
    "WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" was chosen because it was primary and "WindowsBase, Version=8.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" was not.
    References which depend on "WindowsBase, Version=4.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" [C:\Program Files\dotnet\packs\Microsoft.NETCore.App.Ref\10.0.0\ref\net10.0\WindowsBase.dll].
        C:\Program Files\dotnet\packs\Microsoft.NETCore.App.Ref\10.0.0\ref\net10.0\WindowsBase.dll
          Project file item includes which caused reference "C:\Program Files\dotnet\packs\Microsoft.NETCore.App.Ref\10.0.0\ref\net10.0\WindowsBase.dll".
            C:\Program Files\dotnet\packs\Microsoft.NETCore.App.Ref\10.0.0\ref/net10.0/WindowsBase.dll
    References which depend on or have been unified to "WindowsBase, Version=8.0.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35" [].
        C:\Users\Kaz\AppData\Roaming\XIVLauncher\addon\Hooks\dev\Microsoft.VisualStudio.Threading.dll
          Project file item includes which caused reference "C:\Users\Kaz\AppData\Roaming\XIVLauncher\addon\Hooks\dev\Microsoft.VisualStudio.Threading.dll".
            Dalamud
```

Nobody uses this piece of crap anyways, so safe to remove. It was a poorly thought out idea, and it's not worth breaking builds/causing dev consternation over what is, ultimately, an extremely stupid idea.


goaaats:
Could we just replace the library or is it load-bearing? I think this is a nice capability to have, would be a real shame to see it go. It sucks that the broken dependency is nested so deep

KazWolfe:
We could use a different library or just implement our own RPC. As the implementation is currently written though, it is pretty tightly coupled as it’s responsible for actually managing the protocol itself.

I still have no idea why it’s trying to pull in WPF or other systems though, the library and all transients should be cross platform compatible.

nathanctech:
It appears we're not the only ones impacted by this bug: https://github.com/MicrosoftEdge/WebView2Feedback/issues/4959
Microsoft has an open bug on the problem too: https://github.com/dotnet/sdk/issues/41909

It's also hitting folks that use WinForms and forcing them to either suppress the warning or include Wpf, which isn't ideal. The issue appears to be the `Microsoft.VisualStudio.Threading.dll` dependency which is brought in by this Json library. We could always suppress the warning in the Dalamud SDK as it doesn't appear to actually break anything, though.

KazWolfe:
The problem is that we can't suppress just a single instance. I'm not a huge fan of suppressing what is otherwise a legitimate warning - especially if other libraries could be affected in a way that a developer would *want* to know.

PR #2528, opened 2025-12-19
Prevent ImGui text box methods from cloning unchanged input every frame

CMDRNuffin:
The overloads taking a `ref string` for the input text of the various
ways to display a text box would all take the input string, copy it into
a buffer for imgui and then unconditionally produce a new string once
the imgui call returned. Now we only create a new string when the return
value of the native function actually indicates that the text changed.

This makes the GC happy, and also users like me who like to make the GC
happy.

Other side effects: The assumption that the reference doesn't change if
the method returns false, which is very reasonable IMO, is now correct.


goaaats:
Hey, I had to revert this because it was causing issues with certain input flags. Check #dalamud-dev on discord, there's some discussion about it. Sorry!

PR #2533, opened 2025-12-19
Add new themes and update themed path logic

wolfcomp:


goaaats:
I wonder if we could read this from some sheet instead

wolfcomp:
The uld locations are stored in the exe file through `ui/uld/` or `ui/uld/img{:2d}` but actual uld names might be in a sheet somewhere

PR #2540, opened 2025-12-21
Use RowRef in ZoneInitEventArgs

Limiana:
To prevent crashes if there is unexpected data

### DIFF
--- a/Dalamud/Game/ClientState/ZoneInit.cs
+++ b/Dalamud/Game/ClientState/ZoneInit.cs
@@ -3,6 +3,7 @@ using System.Text;
 
 using Dalamud.Data;
 
+using Lumina.Excel;
 using Lumina.Excel.Sheets;
 
 using Serilog;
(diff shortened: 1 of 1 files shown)


goaaats:
When does this happen?

Limiana:
When forcing zone load artificially and if you accidentally supply wrong/outdated parameters Lumina throws exceptions because GetRow is used  - end users likely won't run into this issue, mostly just to prevent unneeded crash and restarts while experimenting

(can live without it to be honest, if you don't really want it)

MidoriKami:
While that may prevent a crash, won't that leave the rest of the service, and even the game itself still in a invalid and undefined state?

Limiana:
Game not really, you don't modify the packet, service - probably, but it's fixed by just receiving a new packet from the server simply by changing zone

Limiana:
Actually changed this PR to use RowRef instead of try-catching, this won't throw and is in line with how Lumina sheets are handled elsewhere in Dalamud

PR #2550, opened 2025-12-22
Add event for LogMessages being added to the chat

RedworkDE:
Adds an easy way for reading system events from the chat, my being able to match chat entries that originate from entries in the `LogMessage` sheet (which is most system messages) by id instead of having to mess around with regex and the various localization issues.

Requires https://github.com/aers/FFXIVClientStructs/pull/1683

<hr/>

For example this could be used to largely replace gatherbuddys multiple files of regex and hacks (https://github.com/Ottermandias/GatherBuddy/blob/main/GatherBuddy/FishTimer/Parser/FishingParser.Data.cs https://github.com/Ottermandias/GatherBuddy/blob/main/GatherBuddy/FishTimer/Parser/FishingParser.Matches.cs https://github.com/Ottermandias/GatherBuddy/blob/main/GatherBuddy/FishTimer/Parser/FishingParser.Regexes.cs) to match a few events with:

```cs
private void OnLogMessage(ILogMessage message)
{
    if (message.LogMessageId == 1110 && message.TryGetIntParameter(0, out var placeName)) // you cast your line <SheetReferenceEn("PlaceName", 2, lnum(1), 2, 1)>.
    {
        if (placeName == 950) // Undiscovered Fishing Hole
        {
            BeganFishing?.Invoke(null);
        }
        else
        {
            var fishingSpot = GatherBuddy.GameData.FishingSpots.Values.FirstOrDefault(s => s.FishingSpotData?.PlaceName.RowId == placeName);
            BeganFishing?.Invoke(fishingSpot);
        }
    }
    else if (message.LogMessageId == 1121) // you recast your line with the fish still hooked.
    {
        BeganMooching?.Invoke();
    }
    else if (message.LogMessageId == 3513 && message.TryGetIntParameter(0, out placeName)) // The fishing hole <SheetReferenceEn("PlaceName", 2, lnum(1), 2, 1)> is added to your fishing log.
    {
        var fishingSpot = GatherBuddy.GameData.FishingSpots.Values.FirstOrDefault(s => s.FishingSpotData?.PlaceName.RowId == placeName);
        IdentifiedSpot?.Invoke(fishingSpot);
    }
}

```


Dalamud/Game/Chat/LogMessage.cs
```diff
@@ -0,0 +1,220 @@
...
+    public SeString FormatLogMessageForDebugging()
+    {
+        var logModule = RaptureLogModule.Instance();
+
+        // the formatting logic is taken from RaptureLogModule_Update
+
+        var utf8 = new Utf8String();
+        SetName(logModule, this.SourceEntity);
+        SetName(logModule, this.TargetEntity);
+        logModule->RaptureTextModule->FormatString(this.GameData.Value.Text.ToDalamudString().EncodeWithNullTerminator(), &ptr->Parameters, &utf8);
```
Infiziert90: Use the Lumina ReadOnlySeString directly, going back to Dalamud shouldn't be done as the SeString system is rather fragile.

Dalamud/Game/Gui/ChatGui.cs
```diff
@@ -41,10 +42,12 @@ internal sealed unsafe class ChatGui : IInternalDisposableService, IChatGui
 
     private readonly Queue<XivChatEntry> chatQueue = new();
     private readonly Dictionary<(string PluginName, uint CommandId), Action<uint, SeString>> dalamudLinkHandlers = new();
+    private readonly List<IntPtr> seenLogMessageObjects = new();
```
Infiziert90: Dalamud uses the `nint` keyword almost everywhere, so this should also follow the same format.

Dalamud/Game/Gui/ChatGui.cs
```diff
@@ -493,6 +502,46 @@ private void HandleLinkClickDetour(LogViewer* thisPtr, LinkData* linkData)
...
+                }
+            }
+
+            this.handleLogModuleUpdate.Original(thisPtr);
+
+            // record the log messages for that we already called the event, but are still in the queue
+            this.seenLogMessageObjects.Clear();
+            foreach (ref var item in thisPtr->LogMessageQueue)
+            {
+                this.seenLogMessageObjects.Add((IntPtr)Unsafe.AsPointer(ref item));
```
Infiziert90: Same here

Infiziert90:
As a more general note on SeString and Lumina ReadOnlySeString, i personally think that SeString shouldn't be used here, as users shouldn't edit these objects.

Dalamud/Game/Gui/ChatGui.cs
```diff
@@ -493,6 +502,46 @@ private void HandleLinkClickDetour(LogViewer* thisPtr, LinkData* linkData)
...
+    {
+        try
+        {
+            foreach (ref var item in thisPtr->LogMessageQueue)
+            {
+                var logMessage = new Chat.LogMessage((LogMessageQueueItem*)Unsafe.AsPointer(ref item));
+
+                // skip any entries that survived the previous Update call as the event was already called for them
+                if (this.seenLogMessageObjects.Contains(logMessage.Address))
+                    continue;
```
MidoriKami, line 516: Are there any cases where the game normally sends 2 of the same log messages in a row that would be valid to trigger on?

Dalamud/Game/Chat/LogMessage.cs
```diff
@@ -0,0 +1,222 @@
...
+    /// </summary>
+    /// <remarks>This can cause side effects such as playing sound effects and thus should only be used for debugging.</remarks>
+    /// <returns>The formatted string.</returns>
+    ReadOnlySeString FormatLogMessageForDebugging();
+}
+
+/// <summary>
+/// This struct represents a status effect an actor is afflicted by.
+/// </summary>
+/// <param name="ptr">A pointer to the Status.</param>
```
nebel: This doc comment seems to be copy-pasted from a different struct.

Dalamud/Game/Gui/ChatGui.cs
RedworkDE: I am not sure if I understand your issue correctly. While there could be multiple items that reference the same `LogMessage` in consecutive frames (or even in one frame), this doesn't check the LogMessageId of the leftover items, but the actual address of the object in the queue. No other object can get allocated at the same address before the current queue item gets removed from the queue by a future call to `Update` (when the item gets actually shown in the chat).

The assumptions this makes for correctness are:
1. The queue never moves objects in the queue, the C++ standard guarantees this for std::deque.
2. `RaptureLogModule.Update` never adds items to the queue, if it did they would never get picked up by this event.
3. Only `RaptureLogModule.Update` removes items from the queue, otherwise we may have called the event for messages that actually never make it into the chat and there could be address collisions that cause us to skip items that were actually not seen before.

Dalamud/Game/Chat/LogMessage.cs
Haselnussbomber: This was not addressed.
```suggestion
        using var rssb = new RentedSeStringBuilder()
        logModule->RaptureTextModule->FormatString(rssb.Builder.Append(this.GameData.Value.Text).GetViewAsSpan(), &ptr->Parameters, &utf8);
```

Dalamud/Game/Chat/LogMessage.cs
```diff
@@ -0,0 +1,222 @@
...
+        // the formatting logic is taken from RaptureLogModule_Update
+
+        var utf8 = new Utf8String();
+        SetName(logModule, this.SourceEntity);
+        SetName(logModule, this.TargetEntity);
+        logModule->RaptureTextModule->FormatString(this.GameData.Value.Text.ToDalamudString().EncodeWithNullTerminator(), &ptr->Parameters, &utf8);
+
+        var result = new ReadOnlySeString(utf8.AsSpan());
+        utf8.Dtor();
+        return result;
```
Haselnussbomber: If you add a `using` to `utf8` above, you can change this here to:

```suggestion
        return new ReadOnlySeString(utf8.AsSpan());
```

Dalamud/Game/Chat/LogMessage.cs
Infiziert90: Shouldn't need the Builder, GameData is a Lumina RowRef, so the data should already be a ReadOnlySeString

Haselnussbomber: Need it for the null terminator. GetViewAsSpan adds it.

Dalamud/Game/Chat/LogMessageEntity.cs
```diff
@@ -0,0 +1,96 @@
...
+
+/// <summary>
+/// Interface representing an entity related to a log message.
+/// </summary>
+public interface ILogMessageEntity : IEquatable<ILogMessageEntity>
+{
+    /// <summary>
+    /// Gets the name of this entity.
+    /// </summary>
+    SeString Name { get; }
```
Haselnussbomber: This should also be ReadOnlySeString.

Haselnussbomber:
I've added a [comment](https://github.com/goatcorp/Dalamud/pull/2550#discussion_r2642858941) to Infis first review, which does not show up, because it was marked as resolved and didn't reopen. Just calling it out here so it doesn’t get overlooked. :)

Dalamud/Game/Chat/LogMessage.cs
RedworkDE: sorry I misread this comment, also I have seen that dalamud has its own formatting service now that could be used, would it make sense to switch to that instead?

Haselnussbomber: In my opinion, if possible, it would be preferred, yes.

1) Avoids adding unnecessary functions to CS.
2) Is for debugging purposes anyway.
3) Uses the UI Language set in the Dalamud settings by default, which is at the same time good for debugging again.
4) Another test-case for the ISeStringEvaluator service in some way. :D

Haselnussbomber: Edit: Wait, scratch that. You can of course use the functions you added to the PR already. I'm in too deep, sorry.

RedworkDE: Made the switch, but this doesn't really change anything about (1) as the mess in SetName still has to stay the same as SeStringEvaluator can't directly accept values for global parameters. I'll have a look at how much of an issue adding support is, but that is probably best left for a separate change.

Haselnussbomber: Sorry to have you sent down that path. It’s fine. I rather have you call CS functions to set global parameters than modifying it directly or making changes to SeStringEvaluator.

Dalamud/Game/Gui/ChatGui.cs
MidoriKami: Oh alright, thanks for the explanation!

Dalamud/Game/Chat/LogMessage.cs
RedworkDE: It is now just your suggestion (and I replaced one more spot so that SeString is no longer used anywhere in this PR)

Dalamud/Game/Gui/ChatGui.cs
RedworkDE: Also for performance considerations of using a simple List.Contains:

In a random sample of running a Dungeon and Alliance Raid (for all those juicy battle log messages), it took a total of 157666 frames, which saw 46086 LogMessages at 23226 frames (including duplicates from loading, with a maximum of 27 in a single frame), had 1399 messages that survived an update call over 555 distinct frames with a maximum of 19 in a single frame, so we are looking at the every occasional frame with up to a thousand comparisons on that line.

I don't think this is enough to justify using a different collection type, but if there is a case where this does actually cause performance issues, it is *probably* possible switch this to a logic only stores the last item that wasn't processed yet and assume that Update never skips over items.

Dalamud/Interface/Internal/Windows/SelfTest/Steps/ChatSelfTestStep.cs
```diff
@@ -34,20 +39,63 @@ public SelfTestStepResult RunStep()
...
-                    this.subscribed = false;
-                    return SelfTestStepResult.Pass;
+                    this.subscribedChatMessage = false;
+                    this.step++;
                 }
 
                 break;
+
+            case 2:
+                ImGui.Text("Teleport somewhere...");
```
KazWolfe: Please replace this test with something that doesn't incur a cost/penalty. It's generally an anti-pattern that we need to waste resources on tests, and one I'd really like to start cleaning up.

KazWolfe (changes requested):

Dalamud/Interface/Internal/Windows/SelfTest/Steps/ChatSelfTestStep.cs
RedworkDE: Switched it to using a mount, if that requires too much MSQ progression for the test chars, I could switch it to the enter/leave sanctuary messages but that would loose out on testing the parameters and I am not sure what messages with parameters can be easily be triggered without any requirements.

Haselnussbomber: Perhaps doing an emote? Those are LogMessages and have parameters (global ones).
Or /itemsearch (LogMessage#1629,1630 with local parameters)?

RedworkDE: Remembered that the battle log exists and changed it to the use action message, less effort to press sprint (or such) once then typing out that command.

goaaats (dismissed):
Very clean, thanks for contributing!

goaaats:
Some warnings in the CI build, would appreciate if you could have a look but otherwise I think this is good to go

vedantmgoyal9:
> Some warnings in the CI build, would appreciate if you could have a look but otherwise I think this is good to go

@goaaats sorry for reaching out here... could you please take a look at https://github.com/goaaats/KeySharp/pull/4
 whenever you get a chance? And if you’re no longer actively maintaining the project, I’d be happy to help keep it going and can take over the repository if you’d prefer. 😊

vedantmgoyal9:
> @goaaats sorry for reaching out here... could you please take a look at [goaaats/KeySharp#4](https://github.com/goaaats/KeySharp/pull/4) whenever you get a chance? And if you’re no longer actively maintaining the project, I’d be happy to help keep it going and can take over the repository if you’d prefer. 😊

@goaaats please reply 🙏🏻

PR #2552, opened 2025-12-24
boot: Turn opt header sum mismatch into warning, thanks NVIDIA

0x0ade:
Launching the game via the injector via NVIDIA Nsight currently results in the optional header checksum check failing. Whatever NVIDIA is doing, it's setting the checksum to 0:

image

Everything else still matches though, the game boots injected and Nsight works too.

I've moved the checksum check to be the last thing and turned it into a warning, so that if (however unlikely it is to happen) two images were to reach that point to only fail afterwards on the sections check, the message only gets logged once on the match that is returned by `get_mapped_image_base_address` anyway.

### DIFF
--- a/Dalamud.Boot/rewrite_entrypoint.cpp
+++ b/Dalamud.Boot/rewrite_entrypoint.cpp
@@ -192,25 +192,27 @@ void* get_mapped_image_base_address(HANDLE hProcess, const std::filesystem::path
                 read_process_memory_or_throw(hProcess, static_cast<char*>(mbi.BaseAddress) + compare_dos_header.e_lfanew + offsetof(IMAGE_NT_HEADERS32, OptionalHeader), compare_nt_header32.OptionalHeader);
                 if (compare_nt_header32.OptionalHeader.SizeOfImage != exe_nt_header32.OptionalHeader.SizeOfImage)
                     continue;
-                if (compare_nt_header32.OptionalHeader.CheckSum != exe_nt_header32.OptionalHeader.CheckSum)
-                    continue;
 
                 std::vector<IMAGE_SECTION_HEADER> compare_section_headers(exe_nt_header32.FileHeader.NumberOfSections);
                 read_process_memory_or_throw(hProcess, static_cast<char*>(mbi.BaseAddress) + compare_dos_header.e_lfanew + sizeof compare_nt_header32, &compare_section_headers[0], sizeof IMAGE_SECTION_HEADER * compare_section_headers.size());
                 if (memcmp(&compare_section_headers[0], &exe_section_headers[0], sizeof IMAGE_SECTION_HEADER * compare_section_headers.size()) != 0)
                     continue;
 
+                if (compare_nt_header32.OptionalHeader.CheckSum != exe_nt_header32.OptionalHeader.CheckSum)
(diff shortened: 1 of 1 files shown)


KazWolfe:
Isn’t this an NVIDIA bug? If so, I’m not sure it’s on us to fix it.

While I freely admit that I’m not quite sure what NSight does (nor why we are checking these checksums - iirc Windows only cares for “critical processes”), it implies that *something* subtle is being changed or is otherwise breaking.

Do you know if 0 would otherwise be considered a special value to indicate there’s no checksum? If so, this would feel slightly safer to merge in. Otherwise, I’m led to wonder what else could explode.

Soreepeong:
Realistically the only two DLLs that should have been loaded into the barely created process should be the target executable itself and `ntdll.dll`, and the game process and `ntdll` are sufficiently different enough to not require checksum comparison. The comparison code is there just in case somehow other DLLs got loaded first (possibly in an illegitimate way.)

0x0ade:
I'm not entirely sure if it's an NVIDIA bug. For context, if you've heard of RenderDoc, it's like that. If you've not heard of that either, Nsight is a graphics dev tool which allows you to perform frame captures and analyze draw call and the render state, among other things. To analyze a game with it, it needs to be launched through it, and it also injects itself into child processes. Dunno how that injection happens though, but the checksum being zeroed does strike me as odd too.

Here's what the memory info looks like on the suspended process at the time of scan:

image

For comparison, here's the extra crap it loads after the process is unsuspended (different PID cus separate game instance):

image 

I'll do some more digging on what exactly it would need to change the checksum to 0 for.

0x0ade:
Turns out NVIDIA is indeed messing with the headers to inject Nsight into child processes in a somewhat funny way. The game process launches suspended with a zeroed checksum (still not sure how), which then gets replaced back by the injector some time after unsuspend, which makes this even funnier IMO.

image

I also went through an ImHex diff between the unloaded disk image and suspended mem image. The image base is obviously different, and as already established, so is the checksum. What's _also_ different is DataDirectory[1], which according to MSDN (at offset 120 for PE32+) is the import table... which makes sense, I guess. It also matches what pe-sieve reports: only the headers changed, nothing else.

I'm not too inclined to go at NVIDIA's stuff with a nutcracker right now, as interesting as this is. I do hope that this is enough info to explain this all though 😅 if still absolutely necessary for whatever reason, I guess I could add additional checks for that specific case to this PR. Want me to-

image

KazWolfe:
So, to confirm, this only really exists to allow interop with NVIDIA's NSight and other debug tools?

Given the context here I'd be more willing to accept (though I'd want to run by @goaaats just to make sure), but I'm also stuck wondering how critical this would be and/or whether we can clean up the boot system to be a bit better overall.

0x0ade:
Yup, and personally I don't deem it too critical - IMO anyone doing graphics debugging should have the necessary tools to build Dalamud installed, but _insert XKCD 2501 here_, and they'd need to either find this PR or come to the same conclusion.

I do wonder if Wine still needs special handling, as from what I understand, K32GetMappedFileName has been implemented with 6.4 back in 2021, and surely nobody is stuck on a Wine version from before that... right?

PR #2558, opened 2025-12-29
feat: Add configurable plugin load order with sync/async overrides

msdx321:
This adds a Dev Tools “Plugin Load Order” page where you manually build the ordered list by selecting enabled plugins from the dropdown and clicking “Add to load order,” then drag rows to reorder and use “Remove” to delete entries; each row shows the plugin’s default load flavor and lets you override it via the dropdown.

On boot, plugins in the list load in that order, keeping sync plugins in the sync path and async plugins in a sequential async path.

When a plugin is disabled it falls out of the enabled list, so it is automatically removed from the load order and any override entry is cleaned up. 

image

### DIFF
--- a/Dalamud/Configuration/Internal/DalamudConfiguration.cs
+++ b/Dalamud/Configuration/Internal/DalamudConfiguration.cs
@@ -338,6 +338,16 @@ internal sealed class DalamudConfiguration : IInternalDisposableService
     public List<string>? DtrOrder { get; set; }
 
     /// <summary>
+    /// Gets or sets the plugin load order, by internal name.
+    /// </summary>
+    public List<string>? PluginLoadOrder { get; set; }
+
+    /// <summary>
+    /// Gets or sets the plugin load order override mode, by internal name.
+    /// </summary>
+    public Dictionary<string, PluginLoadOrderMode>? PluginLoadOrderOverrides { get; set; }
+
+    /// <summary>
     /// Gets or sets the list of ignored DTR elements, by title.
     /// </summary>
     public List<string>? DtrIgnore { get; set; }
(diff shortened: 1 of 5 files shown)


goaaats:
Hi, thanks for your contribution but this is not something we will support. Plugins that depend on other plugins need to coordinate their dependency through IPC, since all plugins must be able to load/unload in any order at runtime. 

If you require this and you can't solve your use case through IPC, please elaborate on that.

msdx321:
> Hi, thanks for your contribution but this is not something we will support. Plugins that depend on other plugins need to coordinate their dependency through IPC, since all plugins must be able to load/unload in any order at runtime.
> 
> If you require this and you can't solve your use case through IPC, please elaborate on that.

Thanks for the response.

My issue is not that one plugin depends on another plugin in terms of functionality or shared state. I agree that such dependencies should be handled via IPC and that plugins must be loadable and unloadable independently.

The problem I am running into is that multiple plugins can hook the same method, and in practice there are third-party plugins that perform unconventional hooking. These can break other plugins purely due to hook ordering, even when there is no logical dependency between them.

Because of this, my motivation for a load order control mechanism is not to express plugin dependencies, but to provide a way to mitigate hook conflicts.

Additionally, this PR would only affect the loading order of plugins whose order has been explicitly overridden. For all other plugins, the existing default loading behavior would remain exactly the same.

goaaats:
Could you give us an example of where this is a problem in practice with links to hooks/delegates?

KazWolfe:
Going to echo a -1 on load order - this is not something I feel like we should be supporting in general for many reasons, including the fact that plugins can be loaded/unloaded at any time during the game's execution.

If hook conflicts continue to be a problem and there's no way to amicably/collaboratively resolve these issues between plugin developers (n.b. custom repo plugin devs _really_ should not be doing Weird Things), then we should probably look at a more robust solution.

In general, too, I don't think it's a great idea to allow plugins to clobber and control order (and certain plugins _will_ reflect into this to set their priorities). We've been resistant to any sort of hook priority up until now because that does negate the entire "collaboration" side of things.

That said, this has come up a few times now (cc @marzent with #1843, and @Glorou since they discussed this in the Discord now). I believe it's time we set up a proper DIP to resolve this particular issue.

PR #2559, opened 2025-12-29
Make color struct human readable

Glorou:
This is to support https://github.com/goatcorp/gc-cimguizmo/pull/1 with human readable color structs, I haven't verified them all yet but these are 1:1 with the C structs


goaaats:
Did you generate this? Would it not be generated by the bindings generator?

Glorou:
I generated and then edited it. When generated, it will output the color names as Color0-Color14, so I modified the struct to line up with the original color definitions.

Critical-Impact:
Is there a way we can alter the generator to do this instead of having to do it by hand each time?

Glorou:
> Is there a way we can alter the generator to do this instead of having to do it by hand each time?

With respect to this, the custom generated partial struct will override the generator, and so long as these variables dont change on the imguizmo side (they havent), they wont need to be re-generated. I recompiled just to triple check as well.

Glorou:
Hey @goaaats, saw you merged the other PR so I wanted to bump this one too.

Thanks!

Glorou:
Just bumping this, hoping to get it in during the hot hotfix patch @goaaats if you can

PR #2565, opened 2026-01-04
Add CPU info to crash log

wolfcomp:
Added a function to retrieve CPU vendor and brand information.

From my system it would add:
```
CPU Vendor: AuthenticAMD
CPU Brand: AMD Ryzen 7 5800X 8-Core Processor
```
to the log
image

### DIFF
--- a/DalamudCrashHandler/DalamudCrashHandler.cpp
+++ b/DalamudCrashHandler/DalamudCrashHandler.cpp
@@ -698,8 +698,6 @@ void get_cpu_info(wchar_t *vendor, wchar_t *brand)
     std::array<int, 4> cpui;
     int nIds_;
     int nExIds_;
-    std::string vendor_;
-    std::string brand_;
     std::vector<std::array<int, 4>> data_;
     std::vector<std::array<int, 4>> extdata_;
     size_t convertedChars = 0;


wolfcomp:
Should work fine on wine as they also use this function in their own debug helper. https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/dbghelp/minidump.c#L589

DalamudCrashHandler/DalamudCrashHandler.cpp
```diff
@@ -941,6 +998,9 @@ int main() {
         const bool is_external_event = exinfo.ExceptionRecord.ExceptionCode == CUSTOM_EXCEPTION_EXTERNAL_EVENT;
 
         std::wostringstream log;
+        wchar_t vendor[0x20];
+        wchar_t brand[0x40];
```
KazWolfe, line 1000: Can these not be std::string?

wolfcomp: I didn't test that but know the `std::format` is very sensitive for what types can be used with the `L""` format.

wolfcomp: cannot be `std::string` no:

`error C2039: 'parse': is not a member of 'std::formatter<std::_Compile_time_parse_format_specs::_FormattedType,std::_Compile_time_parse_format_specs::_CharT>'`

KazWolfe: We already use `std::wstring dumpError;`, so I suspect it should work. I'm really just trying to avoid having to list a fixed buffer and all.

wolfcomp: still would need the fixed buffer as it gets returned from `__cpuid` as `char` and it would need to be transfered over to `wchar_t`

KazWolfe: Is `std::string vendor_;` and `std::string brand_;` just unused then?

wolfcomp: yes forgot to remove it from my test code and copied the entire function which contained them

KazWolfe (approved):

PR #2572, opened 2026-01-05
fix: respect system dark mode setting

Loskh:
This PR fixes the behavior of `WindowIsImmersive `to match its hint and i18n description.
The setting currently forces a fixed dark/light title bar, but the hint says it should follow the system theme and change title bar mode.
This change makes the title bar respect the system’s current light/dark mode when enabled
The `WindowIsImmersive` property is somewhat ambiguous, and I'm unsure whether the name should be changed.

### DIFF
--- a/Dalamud/Memory/MemoryHelper.cs
+++ b/Dalamud/Memory/MemoryHelper.cs
@@ -265,6 +265,31 @@ public static unsafe class MemoryHelper
     }
 
     /// <summary>
+    /// Compares a UTF-16 character span with a null-terminated UTF-16 string at <paramref name="memoryAddress"/>.
+    /// </summary>
+    /// <param name="charSpan">UTF-16 character span (e.g., from a string literal).</param>
+    /// <param name="memoryAddress">Address of null-terminated UTF-16 (wide) string, as used by Windows APIs.</param>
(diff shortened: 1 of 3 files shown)


Dalamud/Memory/MemoryHelper.cs
```diff
@@ -264,6 +264,31 @@ public static bool EqualsZeroTerminatedString(
...
     }
 
+    /// <summary>
+    /// Compares a UTF-16 character span with a null-terminated UTF-16 string at <paramref name="memoryAddress"/>.
+    /// </summary>
+    /// <param name="charSpan">UTF-16 character span (e.g., from a string literal).</param>
+    /// <param name="memoryAddress">Address of null-terminated UTF-16 (wide) string, as used by Windows APIs.</param>
+    /// <returns><see langword="true"/> if equal; otherwise, <see langword="false"/>.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static unsafe bool EqualsZeroTerminatedWideString(
```
Haselnussbomber, line 274: I think this is unnecessary as you can just use `new string((char*)address).Equals(str)`?

https://learn.microsoft.com/en-us/dotnet/api/system.string.-ctor?view=net-10.0#system-string-ctor(system-char*)

Edit: Actually, nevermind. That would copy the string and this function just reads the memory, which is better. Sorry.

PR #2577, opened 2026-01-09
Pause after init

Glorou:
This is an attempt to patch in renderdoc support as its extremely picky about when it can be injected, I want to add in support for passing this param via XL, but I as of yet cant figure it out.


Glorou:
PR in https://github.com/goatcorp/FFXIVQuickLauncher/pull/1822 adds support for this via XL

PR #2582, opened 2026-01-11
fix: prevent AsmHook from corrupting ReadJmpCallSig

Loskh:
Read bytes from the copied module to avoid corruption by AsmHook.  
Also, can CallHook be added back? It was available before but seems to have been removed. https://github.com/goatcorp/Dalamud/blob/731d7e0f6e8206a0137bcd1ffc4e6c93766f62b5/Dalamud/Hooking/Internal/CallHook.cs 
Many functions are now inlined, so this approach has become necessary.


MidoriKami:
Call hook was internal and only used for AddonLifecycle, as the new system doesn't need callsite hooking it wasn't needed anymore and was removed.

Loskh:
Thanks for the comment.  Some plugins already implement their own call hooks, so adding CallHook back isn't really necessary right now.  But those hooks can cause other plugins' ScanText to return incorrect results. The PR is just to keep signature scanning reliable. Hope this fix can be merged!

goaaats:
If we add CallHook back, we should make sure that both CallHook and AsmHook are instantiated through IGameInteropProvider so that we can track their lifetime.

PR #2586, opened 2026-01-13
Remove UiDebug V1 in favor of V2

Infiziert90:
After talking about it in plugin-dev the majority of voices has been using v2 without issues, and if anyone comes up with missing features after this removal they shouldn't be hard to add.

(Had to close the older PR because unresolvable issues from Hasel's other PR)


Haselnussbomber:
Might as well drop the 2 from everything.^^

Dalamud/Interface/Internal/Windows/Data/Widgets/AddonInspectorWidget.cs
```diff
@@ -1,35 +0,0 @@
...
-
-/// <summary>
-/// Widget for displaying addon inspector.
-/// </summary>
-internal class AddonInspectorWidget : IDataWindowWidget
-{
-    private UiDebug? addonInspector;
-    
-    /// <inheritdoc/>
-    public string[]? CommandShortcuts { get; init; } = ["ai", "addoninspector"];
```
KazWolfe, line 11: At the very least, add this shortcut to the new debugger. Either that, or rename everything from 2 (and move 2's shortcuts into here).

Infiziert90:
Removed 2 from everything

PR #2591, opened 2026-01-19
feat(network): Add safe packet handling and pointer validation

RoseOfficial:
## Summary
- Add NetworkPointerValidator for validating packet pointers with bounds checking and user-mode address space limits
- Add SafePacket wrapper providing lifetime safety, bounds checking, and sensitive data clearing on dispose
- Fix integer overflow vulnerabilities in bounds checking (use safe subtraction pattern instead of addition)
- Fix pointer validation timing in GameNetwork to validate after adjustment
- Add deprecation documentation to IGameNetwork interface
- Add runtime check for scoped service access outside plugin context

### DIFF
--- a/Dalamud/Game/Network/GameNetwork.cs
+++ b/Dalamud/Game/Network/GameNetwork.cs
@@ -16,6 +16,26 @@ namespace Dalamud.Game.Network;
 [ServiceManager.EarlyLoadedService]
 internal sealed unsafe class GameNetwork : IInternalDisposableService
 {
+    /// <summary>
+    /// Offset from the data pointer to the start of the packet header.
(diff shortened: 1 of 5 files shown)


Haselnussbomber:
IGameNetwork is obsolete with error set to true, so it can’t be used at all. Any changes to it are unnecessary.

The pointer to a packet cannot be invalid or the game would crash. Packets are to be processed immediately when the detour is called. We don’t know the length of a packet, nor do we need a copy of it to extend its lifetime. The validator is also entirely useless and the SafePacket class is not even used and will never be used.

The GameNetwork service itself is only used for the network monitor widget. If anything, this should be reduced in size.

PR #2592, opened 2026-01-20
chore(network): Remove obsolete IGameNetwork interface

RoseOfficial:
## Summary
Remove the dead `IGameNetwork` interface that was already marked with `[Obsolete(..., true)]`. 

- Plugins cannot use this interface as the obsolete attribute causes compile errors
- Zero dependencies exist on this interface in the codebase
- Internal `GameNetwork` class remains unchanged for NetworkMonitorWidget and other internal tooling

This reduces the GameNetwork service footprint as suggested in #2591 review.

## Test plan
- [x] Build succeeds
- [x] Tests pass (pre-existing flaky test `IsConcurrencySafe` unrelated to this change)
- [x] NetworkMonitorWidget continues to function (uses internal GameNetwork directly)


KazWolfe:
Hi! We merged #2593 in favor of this PR as it was a more complete removal of GameNetwork things - which, generally, is something we don't want to handle too much.

I still see value in some of your debug/testing functionality, so if you want to look at what was modified and make a new PR, we'd be willing to consider it.

Thanks!

PR #2594, opened 2026-01-25
Write troubleshooting to json file

Haselnussbomber:
Occasionally logs don't include the troubleshooting payload when too much was logged afterwards.
We should write it to its own file, so that it doesn't get lost.
Requires changes to Loggy at some point.


goaaats:
We should move all the logs to a subdirectory soon, getting a bit crowded

PR #2599, opened 2026-01-27
Switch GetRow to TryGetRow in ZoneInitEventArgs 

Infiziert90:
- Use TryGetRow to prevent crashes from unresolvable keys
- Add API 15 ToDo: RowRef

#2540 workaround until API 15 can break the API


Critical-Impact:
Are we 100% sure it's not possible for part of the packet to be invalid but the rest is fine
If we built a ZoneInitEventArgs and fill in what we can parse that'd be the closest to the original behaviour and what #2540 was doing

Haselnussbomber:
Unless the packet changes, it can’t contain invalid ids. Otherwise the game would probably crash from it.
This is purely to safeguard data from a third party plugin intercepting the packet incorrectly, wich we shouldn’t support imo.

Infiziert90:
This is indeed purely a crash prevention
It enables the devs may have the issue of #2540 to check for RowId 0 instead of straight up crashing as dalamud throws Lumina errors, which the game also wouldn't do, the game can recover from invalid RowIds just fine.

Haselnussbomber:
> When forcing zone load artificially and if you accidentally supply wrong/outdated parameters Lumina throws exceptions because GetRow is used

Under normal circumstances there is no crash to be prevented.

TerritoryType can’t be 0, otherwise what are you loading into?
ContentFinderCondition has a row 0.
Weather has a row 0.
Festival has a row 0.

In my opinion, the PR to change it to RowRefs is fine, because that’s generally what the Dalamud API provides. It’s sadly a breaking change, so it’s scheduled to API 15. But this PR here is just unnecessary.

Critical-Impact:
I'm kind of in agreement with Hasel here, this is such a unlikely case to occur during normal use. If you're forcing the game to load a different zone and passing in invalid IDs then you're probably already aware of how risky that is and should accept crashes are a part of that.

I'd err on the side of waiting for the API bump and bringing in the RowRef PR unless this is affecting real people

Infiziert90:
The RowRef PR wouldn't do much different, you still have to check every ref for valid state 
Gonna close this then, even tho i think that more safety is better than none

PR #2601, opened 2026-01-28
Enhance HookVerifier to check marshaled types

wolfcomp:


Dalamud/Hooking/Internal/Verification/HookVerifier.cs
```diff
@@ -67,6 +74,7 @@ public static void Verify<T>(IntPtr address) where T : Delegate
         }
 
         var passedType = typeof(T);
+        var isAssemblyMarshaled = passedType.Assembly.GetCustomAttribute<DisableRuntimeMarshallingAttribute>() is not null;
```
Critical-Impact: This name feels like it's the opposite of the value it represents unless I'm reading it wrong. Any chance we could change it so it's clearer for people working in this file later

Critical-Impact (changes requested):
I'm fine with the PR bar the variable naming though I feel we'd want to have this on STG for a bit to catch any plugins that may cause the validator to throw or wait for an API bump

Dalamud/Hooking/Internal/Verification/HookVerifier.cs
wolfcomp: I did an oopsie and check for `is not null` when it is supposed to be `is null`

wolfcomp:
If people try and set a 4 byte return on the SendPacket hook due to how C# works the game will break trying to login anyways.

Critical-Impact (approved):
LGTM

PR #2607, opened 2026-02-02
Use plugin internal name for DataShare tracking

Critical-Impact:
- When relinquishing a DataShare inside a plugins dispose GetCallerName will not get the correct assembly name
- This replaces the StackTrace traversal with a new record called `DataCachePluginId` that contains both the InternalName and EffectiveWorkingId of the plugin
- The DataShare widget displays the InternalName and the effective working IDs on hover


Dalamud/Plugin/DalamudPluginInterface.cs
```diff
@@ -227,19 +227,19 @@ public IDalamudVersionInfo GetDalamudVersion()
 
     /// <inheritdoc/>
     public T GetOrCreateData<T>(string tag, Func<T> dataGenerator) where T : class
-        => Service<DataShare>.Get().GetOrCreateData(tag, dataGenerator);
+        => Service<DataShare>.Get().GetOrCreateData(tag, this.plugin.InternalName, dataGenerator);
```
goaaats: Would it not be better to use the WorkingPluginId here? We can have two plugins with the same internal name.

goaaats (changes requested):

Dalamud/Plugin/DalamudPluginInterface.cs
Critical-Impact: I've used the effective working ID but I've had to use it in combination with the internal name so that it's easier for the user to see what plugin created the data and is potentially stopping it from being relinquished.

I've updated the DataShare widget to reflect this + updated the PR description with the extra changes made.

goaaats:
Broken API is exceptions and doesn't seem used anywhere so I think it's safe to take it now

PR #2609, opened 2026-02-04
Adding access to the windows Thumbnail cache

Glorou:
This adds support for accessing the windows thumbnail cache for a specified file. This should enable reliable thumbnail generation for the Dalamud file browser in the future.


Soreepeong:
Probably worth noting thread requirements or whether it can block.

Critical-Impact:
This doesn't appear to work on WINE, I'm not sure what our general approach with Windows vs WINE but generally I'd expect it to work on both.

Realistically that probably means you'd need to have a system that handles the creation/caching/disposal of thumbnails on dalamud's side

PR #2612, opened 2026-02-06
Fix loading `.tex` file from filesystem

Soreepeong:
It needed `.LoadFile` to work correctly, but since the stuff we need are public anyway might as well use them directly.

### DIFF
--- a/Dalamud/Interface/Textures/Internal/TextureManager.cs
+++ b/Dalamud/Interface/Textures/Internal/TextureManager.cs
@@ -6,7 +6,6 @@ using System.Threading.Tasks;
 using Dalamud.Configuration.Internal;
 using Dalamud.Data;
 using Dalamud.Game;
-using Dalamud.Interface.ImGuiSeStringRenderer.Internal;
 using Dalamud.Interface.Internal;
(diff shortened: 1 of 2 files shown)


Dalamud/Interface/Textures/Internal/TextureManager.cs
```diff
@@ -361,34 +361,31 @@ internal IDalamudTextureWrap NoThrottleCreateFromTexFile(TexFile file)
...
-            tf,
-            [new LuminaBinaryReader(bytesArray)]);
-        // Note: FileInfo and FilePath are not used from TexFile; skip it.
-
-        var wrap = this.NoThrottleCreateFromTexFile(tf);
+        TexFile.TexHeader header;
+        TextureBuffer buffer;
+        fixed (byte* p = fileBytes)
+        {
+            var lbr = new LuminaBinaryReader(new UnmanagedMemoryStream(p, fileBytes.Length));
```
universalconquistador, line 383: I see a stream, does that need to be disposed?

Soreepeong: No, as the stream is a simple view of the memory region.

PR #2614, opened 2026-02-08
Add GPU Info to Crash Handler

reiichi001:
I'm sure there's a better way to do this, but I also shouldn't be allowed to touch any cpp code.

This loops through all dxgi adapters based on example code I ripped from Microsoft and StackOverflow and dumps that into the crash log.

I'm hoping it doesn't make the window too tall, so if there's a better way to list only the display adapters that are unique, I'm all for it.

### DIFF
--- a/DalamudCrashHandler/DalamudCrashHandler.cpp
+++ b/DalamudCrashHandler/DalamudCrashHandler.cpp
@@ -491,7 +491,6 @@ std::vector<IDXGIAdapter1*> enum_dxgi_adapters()
         vAdapters.push_back(pAdapter);
     }
 
-
     if (pFactory)
     {
         pFactory->Release();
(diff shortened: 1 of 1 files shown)


goaaats:
This doesn't show what GPU the game is actually running with but I assume for your purposes it's good enough?

reiichi001:
While knowing which one the game is using would be a nice to have, getting a whole list also has perks, like if someone has an iGPU and dGPU. ...I'm not sure why it doubled up in my tests though, as it listed my GPU twice. (I'm assuming because I have 2 monitors).

There's probably a better way  to get it implemented, but I don't know my way around the DX APIs or C++.

I do have a minor concern that it could make the crash window too tall, but I'm not sure where else this would fit better.

goaaats:
If it ends up being too tall we can put it only in the log on disk.

reiichi001:
That could always work. I considered trying to stuff it into a set or something so only unique values would be shot out, but I don't want to bloat the crash handler of all things.

goaaats (approved):

PR #2619, opened 2026-02-13
Fix Addon/AgentLifyCycle listener execution after unreg request

AtmoOmen:
Added an `internal bool IsRequestedToClear` to both `AddonLifecycleListener` and `AgentLifecycleListener`.

This prevents the deferred to next frame behavior from executing logic that should have already been cleared, avoiding potential leaks or errors.


Critical-Impact:
Reading back on the conversation in Discord (starts at https://discord.com/channels/581875019861328007/860813266468732938/1459763865360404604) this seems like the right approach

PR #2626, opened 2026-02-19
Add new extension and remove debug notes

Infiziert90:
- Add Utf8String extesions (copy of CStringExtension)
- Remove debug notes in Nodetree

### DIFF
--- a/Dalamud/Utility/Utf8StringExtensions.cs
+++ b/Dalamud/Utility/Utf8StringExtensions.cs
@@ -10,7 +10,7 @@ namespace Dalamud.Utility;
 /// A set of helpful utilities for working with <see cref="Utf8String"/>s from ClientStructs.
 /// </summary>
 /// <remarks>
-/// WARNING: Will break if a custom ClientStructs is used. These are here for CONVENIENCE ONLY!.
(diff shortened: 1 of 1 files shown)


Dalamud/Utility/Utf8StringExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    /// <remarks>
+    /// This should be functionally identical to <see cref="AsReadOnlySeStringSpan"/>, but exists
+    /// for convenience in places that already expect ReadOnlySeString as a type (and where a copy is desired).
+    /// </remarks>
+    /// <param name="str">The Utf8String to copy.</param>
+    /// <returns>A new Lumina ReadOnlySeString.</returns>
+    public static ReadOnlySeString AsReadOnlySeString(this Utf8String str)
+    {
+        return new ReadOnlySeString(str.AsSpan().ToArray());
```
Haselnussbomber: Shouldn't this be enough?

```suggestion
        return new ReadOnlySeString(str.AsSpan());
```

Haselnussbomber: The ctor for the Span already calls ToArray:
`public ReadOnlySeString( ReadOnlySpan< byte > data ) => Data = data.ToArray();`
Same thing, less to write.

PR #2644, opened 2026-03-05
Add easier support for live-object IPC across boundaries.

Ottermandias:
Add a wrapper interface to Dalamuds known types that is a disposable and can be used to wrap live-objects with property-accessors for basic types across app context boundaries.


Exter-N:
What if an object wants to expose methods through this mechanism? Shall it expose delegate-valued "properties"?

Ottermandias:
Added full generic invocation support to the object. Removed the `object?` overloads since this is mainly meant for auxiliary libraries that should know full well which types they need.

PR #2648, opened 2026-03-06
Fix warnings in tests

Haselnussbomber:
These are happening since #2624, because `<Nullable>enable</Nullable>` was added to the Dalamud.Test.csproj.

Passing tests


goaaats:
Thanks!

PR #2678, opened 2026-03-21
Introduce IAsyncDalamudPlugin to tidy up plugin lifecycle (v2)

goaaats:
async_lifecycle

Adds new asynchronous plugin interface that fully initializes and uninitializes off the main thread and offers a new LoadAsync(CancellationToken) function to accomplish this. The plugin is not considered loaded until this task completes successfully. The task may be cancelled after a timeout, currently set at 60 seconds (may be changed later upon feedback).

Async disposal through IAsyncDisposable is not cancellable. The plugin must resolve this task and dispose its resources in a timely fashion.


KazWolfe:
Ref `api15` please :)

MidoriKami:
> Ref `api15` please :)

Why? Isn't this in parallel/non breaking to the existing API?

KazWolfe:
> Why? Isn't this in parallel/non breaking to the existing API?

API15 is Coming Soon anyways, this gives us time to preview and hammer out issues before locking in a stable API surface.

I think a month is plenty of time to let devs play with it and provide feedback on how they want to interact with things.

PR #2679, opened 2026-03-22
Support building via MinGW on Linux

0x0ade:
I absolutely do _not_ expect this to get merged as-is, especially because of the .vcxproj -> CMakeFiles generator. But if anyone else is interested in what it takes to build Dalamud with MinGW on Linux and wants to continue from here...

Documentation: https://github.com/0x0ade/dalamud-docs/blob/feature-mingw/docs/building/mingw.md


goaaats:
I think I would be happy to merge this if we built it all through CMake and removed the MSVC projects (so that we can configure for MSVC and mingw)

A prerequisite would probably be moving the native components to another repo, building them there through CI and making releases for them that can be downloaded in CI here and with a script. Would speed up builds significantly as well!

0x0ade:
I'll get rid of the CMake generator and CMake-ify the native side of things eventually, currently too busy with other things and rebasing :sweat_smile:

PR #2703, opened 2026-04-05
Semantic Colors

Ottermandias:
Add semantically named colors (Info, Success, Warning, Error, Attention) as foreground and background versions.
By default, they are the same as TankBlue, HealerGreen, DalamudOrange, DalamudRed, DalamudOrange respectively.

The first commit merely adds them and uses them for notifications, the second commit makes a pass to apply them where the mentioned colors were used.
The second commit DOES change actual default colors in a few places by going from DalamudRed (error) to AttentionForeground (which is Orange instead of Red), but aside from those few cases should result in no changes when using default colors.

I would request a sanity check on the changes in the second commit though before merging.


Dalamud/Interface/Internal/Windows/ColorDemoWindow.cs
```diff
@@ -46,6 +46,16 @@ public ColorDemoWindow()
             ("ParsedOrange", ImGuiColors.ParsedOrange),
             ("ParsedPink", ImGuiColors.ParsedPink),
             ("ParsedGold", ImGuiColors.ParsedGold),
+            ("InfoForeground", ImGuiColors.ParsedGold),
+            ("InfoBackground", ImGuiColors.ParsedGold),
```
Exter-N:
```suggestion
            ("InfoForeground", ImGuiColors.InfoForeground),
            ("InfoBackground", ImGuiColors.InfoBackground),
```

Dalamud/Interface/Internal/Windows/ConsoleWindow.cs
```diff
@@ -305,11 +305,11 @@ public override void Draw()
             hadColor = true;
             if (this.lastCmdSuccess.Value)
             {
-                ImGui.PushStyleColor(ImGuiCol.FrameBg, ImGuiColors.HealerGreen - new Vector4(0, 0, 0, 0.7f));
+                ImGui.PushStyleColor(ImGuiCol.FrameBg, ImGuiColors.SuccessBackground - new Vector4(0, 0, 0, 0.7f));
```
Exter-N: While this differs from the original behavior, it would look saner to me:
```suggestion
                ImGui.PushStyleColor(ImGuiCol.FrameBg, ImGuiColors.SuccessBackground * new Vector4(1, 1, 1, 0.3f));
```

Any thoughts?

Dalamud/Interface/Internal/Windows/ConsoleWindow.cs
```diff
@@ -305,11 +305,11 @@ public override void Draw()
...
             }
             else
             {
-                ImGui.PushStyleColor(ImGuiCol.FrameBg, ImGuiColors.DalamudRed - new Vector4(0, 0, 0, 0.7f));
+                ImGui.PushStyleColor(ImGuiCol.FrameBg, ImGuiColors.ErrorBackground - new Vector4(0, 0, 0, 0.7f));
```
Exter-N: Likewise:
```suggestion
                ImGui.PushStyleColor(ImGuiCol.FrameBg, ImGuiColors.ErrorBackground * new Vector4(1, 1, 1, 0.3f));
```

Dalamud/Interface/Internal/Windows/ConsoleWindow.cs
Exter-N: Or maybe the reduced alpha should be baked directly into the defaults for the new `…Background` colors?

goaaats:
I think what Ny mentioned makes sense to me. The bg colors should be a very muted version of the fg colors, otherwise we should probably not have them.

Ottermandias:
image
Like this?

image
image

goaaats (approved):

PR #2708, opened 2026-04-11
[API15] Addon/Agent Lifecycle Toggles

MidoriKami:
Allows devs to toggle on and off the Addon/Agent Lifecycle Services.

This will restore the original virtual tables for all Addons/Agents so devs that need to trace the origin of calls are able to do so.

Original virtual tables are restored on the next frame after toggled off. This setting does not persist on restart.

Agent Lifecycle can be freely disabled and re-enabled without any change in behavior except an obvious pause on events being triggered.

Addon Lifecycle will unload, but not re-load the virtual tables for active addons, instead it will simply resume listening for addon initialization events. So any addons that had their virtual tables stripped will no longer call AddonLifecycle events until the service is re-enabled, and the addon finalized and re-constructed.


MidoriKami:
Also removed a few redundant `unsafe` blocks that IDE was marking.

goaaats (approved):

PR #2712, opened 2026-04-13
[API15] Update UiDebug

MidoriKami:
Adds a proper method for getting readable typenames correctly, from HaselCommon.

Fixes issue with displaying nested class typenames. (Causes errors while rendering currently if any KTK node is defined nested.)

Updates TypeMappedNodesDataShareName to be "non ktk specific".

### DIFF
--- a/Dalamud/Interface/Internal/UiDebug/UiDebug.cs
+++ b/Dalamud/Interface/Internal/UiDebug/UiDebug.cs
@@ -1,5 +1,6 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Linq;
 
 using Dalamud.Bindings.ImGui;
 using Dalamud.Game.Gui;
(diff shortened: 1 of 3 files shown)


Dalamud/Interface/Internal/UiDebug/UiDebug.cs
```diff
@@ -87,6 +86,87 @@ public void Dispose()
         Service<DataShare>.Get().RelinquishData(StringMappedNodesDataShareName,  this.dalamudInternalId);
     }
 
+    /// <summary>
+    /// Returns a string representing the type passed in.
+    /// </summary>
+    /// <remarks>
+    /// Yoinked with permission from https://github.com/Haselnussbomber/HaselCommon/blob/fe0b266defceca59a174aab3097fc387f581c575/HaselCommon/Extensions/Reflection/TypeExtensions.cs#L50.
+    /// </remarks>
```
Haselnussbomber: I don't think we need to note this down.

```suggestion
```

goaaats (approved):

PR #2718, opened 2026-04-18
Fix multiline log filtering and rendering

niklasstich:
- Renamed LogEntry to LogLine
- Added LogEntry class, only exists to have a unique reference
- LogLine has a reference to a LogEntry
- When filtering LogLines, first group the LogLines by LogEntry, if any of the lines in the grouping match the condition, all the lines in the grouping are added to the filtered collection


niklasstich:
I would appreciate some help testing and possibly fixing this if issues arise, as I am developing on Linux and therefore the workflow is a little clunky, always having to wait 10 minutes on the pipeline before I can test 🫠.

PR #2719, opened 2026-04-19
[API15] Remove SeString

Haselnussbomber:
Mostly just for fun, but also to make a point.

The implementation of SeString is really old and lacking a lot of the games macros.
ReadOnlySeString is a modern implementation, which is more low-level, but because of this, also more powerful.
It supports expressions and is safer to use.

This PR completely replaces SeString in Dalamud with Luminas ReadOnlySeString(Span).

(As I've said in the Discord, I don't expect this to be merged, especially not less than 2 weeks before 7.5.)


Exter-N:
> I don't expect this to be merged, especially not less than 2 weeks before 7.5.

Why not mark it as deprecated then, to get it removed by API 16?

MidoriKami:
> > I don't expect this to be merged, especially not less than 2 weeks before 7.5.
> 
> Why not mark it as deprecated then, to get it removed by API 16?

By depreciated, I assume you mean with Obsolete attributes, that would just frustrate devs as the new API wouldn't be available yet, and they would still get build warnings.

It was suggested that a warning be included in the API15 Changelog/Announcements-Dev, to be prepared for SeString changes for API16.

goaaats:
I can't change the base on this to master it seems, so you might have to reopen + do it yourself, sorry!

PR #2724, opened 2026-04-22
Add new overload for OpenLinkWithMap

Infiziert90:
Function is only callable with a map payload which is not very convenient

### DIFF
--- a/Dalamud/Game/Gui/GameGui.cs
+++ b/Dalamud/Game/Gui/GameGui.cs
@@ -111,6 +111,10 @@ internal sealed unsafe class GameGui : IInternalDisposableService, IGameGui
         => RaptureAtkModule.Instance()->OpenMapWithMapLink(mapLink.DataString);
 
     /// <inheritdoc/>
+    public bool OpenMapWithMapLink(uint territory, uint map, Vector3 worldPos);
+        => RaptureAtkModule.Instance()->OpenMapWithMapLink($"m:{territory},{map},{(int)worldPos.X * 1000},{(int)worldPos.Z * 1000}");
+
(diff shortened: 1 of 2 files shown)


Dalamud/Game/Gui/GameGui.cs
```diff
@@ -110,6 +110,10 @@ private GameGui(TargetSigScanner sigScanner)
     public bool OpenMapWithMapLink(MapLinkPayload mapLink)
         => RaptureAtkModule.Instance()->OpenMapWithMapLink(mapLink.DataString);
 
+    /// <inheritdoc/>
+    public bool OpenMapWithMapLink(uint territory, uint map, Vector3 worldPos);
```
Haselnussbomber:
```suggestion
    public bool OpenMapWithMapLink(uint territory, uint map, Vector3 worldPos)
```

Dalamud/Game/Gui/GameGui.cs
```diff
@@ -447,6 +451,10 @@ void IInternalDisposableService.DisposeService()
     public bool OpenMapWithMapLink(MapLinkPayload mapLink)
         => this.gameGuiService.OpenMapWithMapLink(mapLink);
 
+    /// <inheritdoc/>
+    public bool OpenMapWithMapLink(uint territory, uint map, Vector3 worldPos);
```
Haselnussbomber:
```suggestion
    public bool OpenMapWithMapLink(uint territory, uint map, Vector3 worldPos)
```

Haselnussbomber (commented):
This does not compile like that.

goaaats (approved):

PR #2726, opened 2026-04-26
Fix FollowJmp mishandling indirect jumps

0x0ade:
Required to not crash when AsmHooking instructions such as this one in dxvk:

image

Right now, it misuses the relative offset as an absolute address:

image

image

I wasn't sure whether to throw because of it being an error, or simply return early as that's the least destructive action here and fixes plugins affected by this (namely CustomResolution on Linux). I went with an early return for now though.

### DIFF
--- a/Dalamud/Hooking/Internal/HookManager.cs
+++ b/Dalamud/Hooking/Internal/HookManager.cs
@@ -135,6 +135,10 @@ internal class HookManager : IInternalDisposableService
                         newAddress = (IntPtr)inst.IPRelativeMemoryAddress;
                         newAddress = Marshal.ReadIntPtr(newAddress);
                         break;
+                    case OpKind.Memory when inst.IsJmpNearIndirect:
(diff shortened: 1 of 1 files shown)


Dalamud/Hooking/Internal/HookManager.cs
```diff
@@ -135,6 +135,10 @@ internal static IntPtr FollowJmp(IntPtr address)
                         newAddress = (IntPtr)inst.IPRelativeMemoryAddress;
                         newAddress = Marshal.ReadIntPtr(newAddress);
                         break;
+                    case OpKind.Memory when inst.IsJmpNearIndirect:
+                    case OpKind.Memory when inst.IsJmpFarIndirect:
+                        Log.Verbose($"Encountered indirect assembly jump ({kind}) from {address.ToInt64():X}, returning unresolved");
+                        return address;
```
KazWolfe, line 141: This should probably be considered a failed hook and treated the same, but for now this is probably better than nothing.

PR #2728, opened 2026-04-28
Hook verifier fix

Ottermandias:
The hook verifier currently explodes on duplicate addresses.
We probably should not have duplicate addresses, but this should at least fix the explosion when there are any.

### DIFF
--- a/Dalamud/Hooking/Hook.cs
+++ b/Dalamud/Hooking/Hook.cs
@@ -205,9 +205,10 @@ public abstract class Hook<T> : IDalamudHook where T : Delegate
         var assembly = callingAssembly ?? Assembly.GetCallingAssembly();
 
         // TODO: Only log verification exceptions for now, figure out how to handle this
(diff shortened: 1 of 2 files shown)


Ottermandias:
Woops, wrong branch initially, sorry.

epinephren:
Tested this PR (#2728) locally on api15.

Before:
HookVerifier crashed during startup with:
System.ArgumentException: An item with the same key has already been added. Key: 0
at HookVerifier.Initialize(...):line 111

After applying this PR:
- No more duplicate key crash
- HookVerifier continues execution
- Still seeing unresolved signatures (expected for patch), but no fatal

Build:
- Local build succeeds
- VS2022 toolchain
- NoLoadPlugins=true

Runtime:
- Injector proceeds past previous crash point
- Game still unstable due to unresolved hooks, but HookVerifier no longer hard-crashes

Logs available if needed.

PR #2736, opened 2026-04-28
[API15] ClientState Touchups

MidoriKami:
Fixes lumina issue by not using lumina!

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -257,6 +257,7 @@ internal sealed class ClientState : IInternalDisposableService, IClientState
         this.onLogoutHook = Hook<LogoutCallbackInterface.Delegates.OnLogout>.FromAddress((nint)AgentLobby.Instance()->LogoutCallbackInterface.VirtualTable->OnLogout, this.OnLogoutDetour);
         this.onLogoutHook.Enable();
 
+        this.IsPvP = GameMain.IsInPvPArea();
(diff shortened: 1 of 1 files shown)


Dalamud/Game/ClientState/ClientState.cs
```diff
@@ -333,6 +318,7 @@ private unsafe void UIModuleHandlePacketDetour(
                 Log.Debug($"ZoneInit: {eventArgs}");
                 this.ZoneInit?.InvokeSafely(eventArgs);
                 this.TerritoryType = eventArgs.TerritoryType.RowId;
+                this.IsPvP = GameMain.IsInPvPArea();
```
Haselnussbomber: GameMain goes through some zone transition stuff, so the row pointer might not be set here yet.

MidoriKami: I'll take a look as soon as I am able to

Dalamud/Game/ClientState/ClientState.cs
Haselnussbomber, line 261: You should probably set IsPvP here.

MidoriKami: Should it also be set before ZoneIniti is invoked? Should TerritoryType be moved up aswell?

Haselnussbomber: No it's fine. This is just the Setup function that is called on first tick. We want to fill the properties with data and have events fired here (for late injects).

PR #2737, opened 2026-04-29
Update XivChatType.cs

Infiziert90:
Taken from https://github.com/Infiziert90/ChatTwo/blob/main/ChatTwo/Code/ChatType.cs


KazWolfe:
Merged manually.

PR #2740, opened 2026-04-29
Enum generator fix

Ottermandias:
Make EnumCloneGenerator work with enums nested in classes or structs and add test for that.


Ottermandias:
again wrong base...

goaaats (approved):

goaaats:
Great, thank you!

PR #2751, opened 2026-04-30
Fix Sender/Message not being marked as modified

Haselnussbomber:
Plugin devs are able to modify the Payloads of the Sender and Message SeStrings.
I didn't think of that, when I made the new API, because I was thinking in ReadOnly mode, so those had to be set with new SeStrings.
I've changed it, so that it now saves the original sender and message as ReadOnlySeString and compares agains them.
This sadly causes allocations when accessing the Modified properties, but that was basically how it worked pre-API15.

### DIFF
--- a/Dalamud/Game/Chat/ChatMessage.cs
+++ b/Dalamud/Game/Chat/ChatMessage.cs
@@ -124,10 +124,24 @@ internal class ChatMessage : IHandleableChatMessage
     public SeString Message { get; set; }
 
     /// <inheritdoc />
-    public bool SenderModified => new ReadOnlySeString(this.Sender.Encode()) != this.OriginalSender;
+    public bool SenderModified
+    {
+        get
+        {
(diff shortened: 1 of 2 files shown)


Dalamud/Game/Chat/ChatMessage.cs
```diff
@@ -99,40 +112,36 @@ internal class ChatMessage : IHandleableChatMessage
...
-        set
+        get
         {
-            var newValue = value ?? new SeString();
-
-            if (field == null || !field.Encode().SequenceEqual(newValue.Encode()))
-                this.MessageModified = true;
-
-            field = newValue;
+            var encoded = this.Sender.Encode();
```
goaaats, line 131: I guess we could make this a one-way and cache it, in case anyone else checks this. But good enough for now

goaaats (approved):

PR #2758, opened 2026-04-30
Add 2 extension methods for the newly added XivChatType

Infiziert90:
Taken from https://github.com/Infiziert90/ChatTwo/blob/main/ChatTwo/Code/ChatTypeExt.cs#L339

All entries didn't exist in Dalamud, and now that ChatMessage returns the actual Source and Target some devs may be interested in what XivChatType support them. 
Gm one has been added for easier filtering in plugins that may have accidentally block them before.


Infiziert90:
image
image

Stylecop doesn't like either option, i can apply pragma if wanted

goaaats (approved):

PR #2767, opened 2026-05-01
Use already resolved addresses from CS for hook verifier

goaaats:


Dalamud/Hooking/Internal/Verification/HookVerifier.cs
```diff
@@ -49,64 +45,74 @@ public static unsafe void Initialize(TargetSigScanner scanner)
         foreach (var csType in csTypes)
```
Haselnussbomber: I wonder if this would be faster if we'd use `Parallel.ForEach` here? CS has a lot of types...

Dalamud/Hooking/Internal/Verification/HookVerifier.cs
```diff
@@ -49,64 +45,74 @@ public static unsafe void Initialize(TargetSigScanner scanner)
...
+            Type? delegateType = null;
             foreach (var method in methods)
             {
                 if (method.GetCustomAttribute<ObsoleteAttribute>() is { }) continue;
+
+                var memberFunctionAttribute = method.GetCustomAttribute<MemberFunctionAttribute>();
+                var staticAddressAttribute = method.GetCustomAttribute<StaticAddressAttribute>();
+
+                if (memberFunctionAttribute == null && staticAddressAttribute == null)
+                    continue;
```
Haselnussbomber: Would be better if you could use `Attribute.IsDefined` for the 3 checks here, as it's a lot faster.

Haselnussbomber (commented):
Comments for performance.

PR #2772, opened 2026-05-02
Restore old SizeContrains logic to prevent 0x0 windows

Infiziert90:
Prevents 0x0 size contrains from occuring.

If a plugin window ctor uses `ScaledVector2` it triggers undefined behaviour based on load order, with the potential result being that `ImGuiHelper.GlobalScale` being still 0 and the follow up calculation `Vector2 * 0` returning a `Vector2(0, 0)`

This PR prevents the 0 from spiraling into the `SetNextWindowSizeConstraints()`

Current behaviour:
```cs
Minimum = 0,
Maximum = Not set (default 0)

if (maximum < minimum) <--- false
    return maxValue
    
    
return (min, max) <--- both 0

<--- Window is forced to resize to 0x0 because Range = 0 <---> 0
```

Restored behaviour:
```cs
Minimum = 0,
Maximum = Not set (default maxValue)

if (maximum < minimum) <--- false
    return maxValue
    
    
return (min, max) <--- 0, maxValue

<--- Window keeps whatever Size it has because Range = 0 <---> maxValue
```

### DIFF
--- a/Dalamud/Interface/Windowing/WindowHost.cs
+++ b/Dalamud/Interface/Windowing/WindowHost.cs
@@ -543,10 +543,7 @@ public class WindowHost
 
         if (this.Window.SizeConstraints.HasValue)
         {
-            var (min, max) = this.GetValidatedConstraints(this.Window.SizeConstraints.Value);
-            ImGui.SetNextWindowSizeConstraints(
-                min * ImGuiHelpers.GlobalScale,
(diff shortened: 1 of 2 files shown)


goaaats:
I think I fixed this another way with the WindowHost alignment, but thanks for the analysis, was very helpful!

PR #2777, opened 2026-05-03
Add plugin installer filters

MapleRecall:
The current sorting method is a bit strange, `Not Installed` should be used as a filter rather than a sorting option... So I made a filter to make it easier to browse and search plugins

- Add a filter toggle to the plugin installer header that expands Installed, Unverified, and Incompatible checkbox filters.
- Apply the filters to available and installed plugin lists, while disabling Installed in the Installed view and disabling all filters for Installed Dev Plugins.
- Remove the obsolete Not Installed sort option.

6a46580158276dbff52a2c1255859ffe
<br>
1df1cf2e4be3c14320d95ce621318794

### DIFF
--- a/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
@@ -128,6 +128,11 @@ internal class PluginInstallerWindow : Window, IDisposable
     private string filterText = Locs.SortBy_Alphabetical;
     private bool adaptiveSort = true;
 
+    private bool showPluginFilters = false;
+    private bool showInstalledPlugins = true;
+    private bool showThirdPartyPlugins = true;
+    private bool showUnsupportedPlugins = true;
+
     private OperationStatus installStatus = OperationStatus.Idle;
     private OperationStatus updateStatus = OperationStatus.Idle;
 
@@ -210,7 +215,6 @@ internal class PluginInstallerWindow : Window, IDisposable
         DownloadCount,
         LastUpdate,
         NewOrNot,
-        NotInstalled,
         EnabledDisabled,
         ProfileOrNot,
         SearchScore,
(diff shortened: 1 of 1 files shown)


Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
```diff
@@ -704,11 +708,35 @@ private void DrawHeader()
...
-        ImGui.SetCursorPosX(windowSize.X - sortSelectWidth - (style.ItemSpacing.X * 2) - searchInputWidth - searchClearButtonWidth);
+        var controlsStartX = windowSize.X - sortSelectWidth - filterButtonWidth - (style.ItemSpacing.X * 3) - searchInputWidth - searchClearButtonWidth;
+        ImGui.SetCursorPosX(controlsStartX);
 
         var isProfileManager =
             this.categoryManager.CurrentGroupKind == PluginCategoryManager.GroupKind.Installed &&
             this.categoryManager.CurrentCategoryKind == PluginCategoryManager.CategoryKind.PluginProfiles;
+        var disableHeaderListControls = this.categoryManager.CurrentGroupKind == PluginCategoryManager.GroupKind.Changelog || isProfileManager;
+
+        using (ImRaii.Disabled(disableHeaderListControls))
```
goaaats, line 720: Maybe it would be nice to have this in a popup instead, so that it doesn't push down the installer contents?

goaaats (changes requested):

goaaats:
Sorry for taking a while to get to this. Looks nice!

Dalamud/Interface/Internal/Windows/PluginInstaller/PluginInstallerWindow.cs
MapleRecall: snipaste_20260515_215652

Great idea! I've updated it.

goaaats (approved):

PR #2780, opened 2026-05-04
Add Developer Mode to settings

Haselnussbomber:
This PR adds a new "Enable Developer Mode" checkbox to the settings, enabling various other deveoper-specific settings, such as:

- Dev Plugin Locations
- Enable ImGui asserts
- Always enable ImGui asserts at startup
- The Scan Dev Plugins button in the Plugin Installer

The Developer Menu option in the Title Screen Menu was only shown on tracks other than release (e.g. stg), but now it's also shown when Developer Mode is enabled. (I also made it a toggle.)


goaaats (dismissed):

goaaats:
I think this needs a check in PluginManager, we shouldn't load/scan dev plugins if we have dev plugin paths but dev mode is disabled. Looks great otherwise though!

goaaats (approved):

PR #2781, opened 2026-05-04
Add 30 seconds InputTimer check to IsClientIdle

Haselnussbomber:
Recently my plugins updated as I was teleporting and crashed.
It may have crashed because of other reasons, but it felt like a bad time to update plugins nevertheless.
So I thought it might be wise to wait a certain amount of time since the last user input.
I chose 30 seconds for now, but we can adjust this if necessary.

### DIFF
--- a/Dalamud/Game/ClientState/ClientState.cs
+++ b/Dalamud/Game/ClientState/ClientState.cs
@@ -228,7 +228,7 @@ internal sealed class ClientState : IInternalDisposableService, IClientState
         unsafe
         {
             var inputTimerModule = InputTimerModule.Instance();
-            if (inputTimerModule != null && inputTimerModule->InputTimer is not 0 and < 30)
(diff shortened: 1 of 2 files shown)


nebel:
I wonder if the condition you meant to check is `inputTimerModule->InputTimer < 30`? Looks like `AfkTimer` only changes when actually in the AFK status (and goes negative for some reason).

Haselnussbomber:
It increases at the same time for me. I'm not sure. I can switch to InputTimer.

Screenshot

nebel:
> It increases at the same time for me. I'm not sure. I can switch to InputTimer.

This seems to depend on the System Configuration -> Other Settings -> Auto-AFK Settings -> Time Before Auto-AFK. If it's set to "Off" then AfkTimer does not increment.

Haselnussbomber:
> goes negative for some reason

It sure does. 😅

Screenshot

goaaats (approved):

PR #2790, opened 2026-05-06
Don't mutate existing style when using add new style button

nebel:
Currently, when we grab the "new style" here, we just pull it from the existing list which means the styles are the same style object in memory. This causes the following line to change the name of both new and old styles.

In addition to being unexpected, this may cause a draw error when immediately exiting the window via "Close" and then re-opening the window, as `ChosenStyle` will become invalid (as it contains the name of the previously selected style which was erroneously renamed).

### DIFF
--- a/Dalamud/Interface/Internal/Windows/StyleEditor/StyleEditorWindow.cs
+++ b/Dalamud/Interface/Internal/Windows/StyleEditor/StyleEditorWindow.cs
@@ -102,7 +102,7 @@ public class StyleEditorWindow : Window
         {
             this.SaveStyle();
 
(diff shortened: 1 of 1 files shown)


goaaats:
Damn, I changed this recently because I thought it was useful that the "add style" button copied the current style. We would have to add a proper copy I guess, or serialize/deserialize. Thanks for catching that!

nebel:
Yeah I actually did it with a Serialize -> Deserialize at first, but since `StyleModelV1.Get()` copies from `ImGui.GetStyle()` I think doing it this way does effectively clone the selected style (since the selected style is always active),

goaaats (approved):

PR #2796, opened 2026-05-09
Fix resetting UI language to English when saving any Dalamud settings

MapleRecall:
Currently, if any plugin (such as Orchestrion) subscribes to the `LocalizationChanged` event, saving dalamud settings may cause this handler to execute earlier than dalamud itself, resulting in dalamud using a corrupted CheapLoc state, causing all strings to fallback to English before the next restart.

c3ac117ffc057332fbf4ffc5a5099c14

I did two things in this PR:

1. Restore Dalamud localization state before and after each `LocalizationChanged` subscriber invocation so plugin handlers cannot leak CheapLoc setup side effects into later subscribers or the final Dalamud state.
2. Avoid reapplying the language settings when the language did not change.


goaaats (approved):

goaaats:
I think this is a fine workaround for now. Ideally it wouldn't be necessary, but for as long as we are using CheapLoc it's probably the way to go.

PR #2801, opened 2026-05-09
[master] Update ClientStructs

bleatbot:


KazWolfe:
Intended API break:

```
  CP0002: Member 'void FFXIVClientStructs.FFXIV.Client.UI.Shell.RaptureShellModule.ChangeChatChannel(int, uint, FFXIVClientStructs.FFXIV.Client.System.String.Utf8String*, bool)' exists on left\FFXIVClientStructs.dll but not on right\FFXIVClientStructs.dll
```

(Wrong return type prior)

KazWolfe (approved):

PR #2804, opened 2026-05-10
feat: Allow blocking chat commands in macros

KazWolfe:
Add a new parameter to CommandInfo named `AllowedInMacros`. This parameter (currently defaulting to `true`) will permit this command to be run from within a macro context. If set to false, the command may not be run while a macro is active.

Ideally, this option will default to `false` in API16, requiring plugin developers *opt in* to macro execution. This would prevent potential edge cases where a macro can call a plugin command that in turn triggers actions that normally cannot be done within the context of a macro.

At present, this implementation doesn't actually determine if the command is *in* the macro or is just running while a macro is active. If this is desired, it should be addable without too much effort, but I wanted to keep the initial proposal simple.

### DIFF
--- a/Dalamud/Game/Command/CommandManager.cs
+++ b/Dalamud/Game/Command/CommandManager.cs
@@ -1,10 +1,14 @@
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+
+using CheapLoc;
 
 using Dalamud.Console;
 using Dalamud.Game.Gui;
+using Dalamud.Game.Gui.Toast;
 using Dalamud.Hooking;
 using Dalamud.IoC;
 using Dalamud.IoC.Internal;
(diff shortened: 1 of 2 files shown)


Dalamud/Game/Command/CommandManager.cs
```diff
@@ -97,6 +99,13 @@ public bool ProcessCommand(string content)
         if (!this.commandMap.TryGetValue(command, out var handler)) // Command was not found.
             return false;
 
+        if (!handler.AllowedInMacros && this.IsMacroRunning())
+        {
+            Log.Warning("Command {CommandName} is not allowed to be run in macros.", command);
+            Service<ChatGui>.Get().PrintError($"Command {command} cannot be run in a macro.");
```
MidoriKami: Might consider also including a Toast.PrintError, as that's typically what the game will do in the case of most command related errors.

Alternatively might consider dalamud INotification service, but that's probably a bit too much.

Haselnussbomber: Make this localizable please and thanks!

goaaats (dismissed):

goaaats (approved):

PR #2813, opened 2026-05-14
Fix CorePlugin for new IAsyncDalamudPlugin interface

Aireil:
It looks like this needs to be changed? Please confirm before merging, I did not test anything, just Rider complaining.
Edit: Well, I tested building, it does not build without the change and builds with it, but I have not done anything else! :)


goaaats:
Thanks! Yeah someone brought it up before, I just forgot about it. Appreciate it.

goaaats (approved):

PR #2824, opened 2026-05-19
Avoid error on disabled rollup workflow

Aireil:
Fixes this big red scary text!
image

Could also just be skipped instead, but I just restored the way it was disabled before API 15.


wolfcomp:
Workflow could also just be disabled from the `Actions` menu by anyone with the ability to do so. 
image

goaaats (approved):

PR #2828, opened 2026-05-24
Make ObjectTable Safe during shutdown

MidoriKami:
It's not pretty, but it works, tested access to ObjectTable in VanillaPlus and SimpleTweaks during game shutdown, and no errors were thrown.

This might have potential consequences that I'm not aware of, let me know your thoughts.

### DIFF
--- a/Dalamud/Game/ClientState/Objects/ObjectTable.cs
+++ b/Dalamud/Game/ClientState/Objects/ObjectTable.cs
@@ -51,6 +51,11 @@ internal sealed partial class ObjectTable : IServiceType, IObjectTable
     {
         get
         {
+            if (Service<Framework>.Get().IsFrameworkUnloading)
+            {
+                return nint.Zero;
+            }
+
             ThreadSafety.AssertMainThread();
 
             return (nint)(&CSGameObjectManager.Instance()->Objects);
@@ -86,6 +91,11 @@ internal sealed partial class ObjectTable : IServiceType, IObjectTable
     {
         get
         {
+            if (Service<Framework>.Get().IsFrameworkUnloading)
+            {
+                return null;
+            }
+
             ThreadSafety.AssertMainThread();
 
             return (index >= objectTableLength || index < 0) ? null : this.cachedObjectTable[index].Update();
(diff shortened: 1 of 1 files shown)


Dalamud/Game/ClientState/Objects/ObjectTable.cs
```diff
@@ -235,13 +265,28 @@ internal sealed partial class ObjectTable
     /// <inheritdoc/>
     public IEnumerator<IGameObject> GetEnumerator()
     {
+        if (Service<Framework>.Get().IsFrameworkUnloading)
+        {
+            return EmptyEnumerator();
```
Haselnussbomber: Simpler version:
```suggestion
            return Enumerable.Empty<IGameObject>().GetEnumerator();
```

Dalamud/Game/ClientState/Objects/ObjectTable.cs
```diff
@@ -235,13 +265,28 @@ internal sealed partial class ObjectTable
...
+    /// <summary>
+    /// An empty enumerator that returns the default value of IGameObject.
+    /// For use when the game is unloading to return empty object data safely.
+    /// </summary>
+    /// <returns>An enumerator of empty objects.</returns>
+    private static IEnumerator<IGameObject> EmptyEnumerator()
+    {
+        yield return default;
+    }
+
```
Haselnussbomber: Remove this then.

```suggestion
```

Alternatively, if you don't want to remove this (I wouldn't know why), at least change it to `yield break;` because otherwise you return 1 `default(IGameObject)`.

Haselnussbomber (changes requested):

Dalamud/Game/ClientState/Objects/ObjectTable.cs
Haselnussbomber: I think this requires to add a `using System.Linq;` too. Sorry.

MidoriKami:
Sorry I forgot to comment here wrt to hasels code suggestions, It was discussed in DM's.

The main concern I had with the IObjectTable enumerator is I wanted to find a way for it to return an entire object tables worth of null IGameObjects so that this code doesn't start causing exceptions to anyone using things like `.ElementAt(...)`, but I didn't find any sane way to do that, and then got distracted with other projects.

Regardless, https://github.com/goatcorp/Dalamud/pull/2857 supersedes this PR, as it will make the game still be on the main thread while shutting down, and I believe that plugin code should treat game shutdown the same as if the plugin is being unloaded normally, for simplicity and sanity.

MidoriKami:
Made unnecessary by #2857

PR #2829, opened 2026-05-24
Thread safe Agent/Addon Lifecycle registration/unregistration

MidoriKami:
Adds a lock mostly for `RemoveAll` to make it behave when called from async.


goaaats (approved):

Haselnussbomber:
Just saw this and am wondering why we use the Lock object, but not `using var scope = this.listenerLock.EnterScope();` 🤔

MidoriKami:
Because I barely know what I am doing, and didn't know that was a thing.

PR #2839, opened 2026-05-31
Use Lock.Scope throughout the project

Haselnussbomber:
While a `lock` block on a `Lock` object is translated into an `EnterScope` call on compile, doing it explicitly feels cleaner and also uncovered that a couple locks were still on `object`s/`List`s instead. Using `lock` on an `object` makes it use the legacy `Monitor` class, which has some overhead.

I suggest to review without whitespace changes. :)
Hope I didn't break anything. 😇


Haselnussbomber:
Rebased to fix merge conflicts. :)

goaaats (approved):
Thank you!

PR #2857, opened 2026-06-17
Cleaner Dalamud shutdown

Haselnussbomber:
- Use CancellationTokens throughout Dalamud to properly cancel Tasks when the game is shutting down.
- Prevent fonts from building and textures from being created using CancellationTokens.
- Unregister singletons from ServiceContainer when they are unset.
- Reworked how Dalamud disposes services: first all plugins are unloaded, then internal services. You'll notice it no longer freezes the process while it's waiting for services to be unloaded. It now returns false in HandleFrameworkDestroy until that's done, as the game does it.

### DIFF
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -10,7 +10,6 @@ using Dalamud.Common;
 using Dalamud.Configuration.Internal;
 using Dalamud.Game;
 using Dalamud.Hooking.Internal.Verification;
-using Dalamud.Interface.Internal;
 using Dalamud.Plugin.Internal;
 using Dalamud.Storage;
 using Dalamud.Utility;
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -195,6 +195,7 @@ internal sealed unsafe class Dalamud : IServiceType
     public void WaitForUnload()
     {
         this.unloadSignal.WaitOne();
+        this.unloadSignal.Dispose();
     }
 
     /// <summary>
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -10,6 +10,7 @@ using Dalamud.Common;
 using Dalamud.Configuration.Internal;
 using Dalamud.Game;
 using Dalamud.Hooking.Internal.Verification;
+using Dalamud.Interface.Internal;
 using Dalamud.Plugin.Internal;
 using Dalamud.Storage;
 using Dalamud.Utility;
@@ -33,7 +34,7 @@ namespace Dalamud;
 /// The main Dalamud class containing all subsystems.
 /// </summary>
 [ServiceManager.ProvidedService]
-internal sealed unsafe class Dalamud : IServiceType
+internal sealed class Dalamud : IServiceType
 {
     #region Internals
 
@@ -186,7 +187,15 @@ internal sealed unsafe class Dalamud : IServiceType
             Windows.Win32.PInvoke.CreateMutex(attribs, false, "DALAMUD_CRASHES_NO_MORE");
         }
 
-        this.unloadSignal.Set();
+        Task.Run(() =>
+        {
+            if (Service<PluginManager>.GetNullable() is { } pluginManager)
+                pluginManager.UnloadAllPlugins().Wait();
+
+            ServiceManager.UnloadAllServices();
+
+            this.unloadSignal.Set();
+        });
     }
 
     /// <summary>
@@ -219,7 +228,7 @@ internal sealed unsafe class Dalamud : IServiceType
     /// <summary>
     /// Helper function to set the exception handler.
     /// </summary>
-    private static nint SetExceptionHandler(nint newFilter)
+    private static unsafe nint SetExceptionHandler(nint newFilter)
     {
         var oldFilter =
             Windows.Win32.PInvoke.SetUnhandledExceptionFilter((delegate* unmanaged[Stdcall]<global::Windows.Win32.System.Diagnostics.Debug.EXCEPTION_POINTERS*, int>)newFilter);
--- a/Dalamud/Dalamud.cs
+++ b/Dalamud/Dalamud.cs
@@ -188,9 +188,6 @@ internal sealed class Dalamud : IServiceType
 
         Task.Run(() =>
         {
-            if (Service<PluginManager>.GetNullable() is { } pluginManager)
-                pluginManager.UnloadAllPlugins().Wait();
-
             ServiceManager.UnloadAllServices();
 
             this.unloadSignal.Set();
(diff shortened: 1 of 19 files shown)


Haselnussbomber:
As of commit e65766d ("Make RunOnTick work when Framework is unloading"):

- `IFramework.Update` runs until Destroy is called, just as before
- `IAsyncDalamudPlugin.DisposeAsync` is not on the main thread, as before
- `IFramework.RunOnFrameworkThread`/`RunOnTick`/`Run` will continue to work in `IAsyncDalamudPlugin.DisposeAsync` and are awaitable without deadlocking the game! 🥳

MidoriKami:
closes #2828

Dalamud/Interface/Internal/InterfaceManager.cs
```diff
@@ -352,9 +352,9 @@ void IInternalDisposableService.DisposeService()
         void ClearHooks()
         {
             this.wndProcHookManager.PreWndProc -= this.WndProcHookManagerOnPreWndProc;
+            this.dxgiSwapChainPresentHook?.Dispose();
+            this.reShadeDxgiSwapChainPresentHook?.Dispose();
```
goaaats: Why was it necessary to move these out of the Interlocked.Exchange?

Dalamud/Service/Service{T}.cs
```diff
@@ -412,7 +406,12 @@ private static Task<object[]> ResolveInjectedParameters(
     /// Pull the instance out of the service locator, waiting if necessary.
     /// </summary>
     /// <returns>The object.</returns>
-    private static Task<object> GetAsyncAsObject() => instanceTcs.Task.ContinueWith(r => (object)r.Result);
+    private static Task<object> GetAsyncAsObject() =>
```
goaaats, line 409: Does this mean you can't resolve services during shutdown?

Dalamud/Dalamud.cs
```diff
@@ -187,7 +188,15 @@ public void Unload()
             Windows.Win32.PInvoke.CreateMutex(attribs, false, "DALAMUD_CRASHES_NO_MORE");
         }
 
-        this.unloadSignal.Set();
+        Task.Run(() =>
+        {
+            if (Service<PluginManager>.GetNullable() is { } pluginManager)
+                pluginManager.UnloadAllPlugins().Wait();
```
goaaats: It's not 100% clear to me now why this needs to be here. There is a call to UnloadAllPlugins() in the PluginManager Dispose as well, the shutdown order should ensure that all plugin services are still alive until PluginManager is disposed.

Dalamud/EntryPoint.cs
```diff
@@ -56,7 +56,7 @@ public static void Initialize(IntPtr infoPtr, IntPtr mainThreadContinueEvent)
         if ((info.BootWaitMessageBox & 4) != 0)
             Windows.Win32.PInvoke.MessageBox(HWND.Null, "Press OK to continue (BeforeDalamudConstruct)", "Dalamud Boot", MESSAGEBOX_STYLE.MB_OK);
 
-        new Thread(() => RunThread(info, mainThreadContinueEvent)).Start();
+        new Thread(() => RunThread(info, mainThreadContinueEvent)) { Name = "EntryPointThread" }.Start();
```
goaaats:
```suggestion
        new Thread(() => RunThread(info, mainThreadContinueEvent)) { Name = "Dalamud EntryPoint" }.Start();
```

Dalamud/EntryPoint.cs
```diff
@@ -209,11 +197,6 @@ private static void RunThread(DalamudStartInfo info, IntPtr mainThreadContinueEv
             Log.CloseAndFlush();
             SerilogEventSink.Instance.LogLine -= SerilogOnLogLine;
         }
-
-        // If we didn't unload services correctly, we need to kill the process.
-        // We will never signal to Framework.
-        if (unloadFailed)
-            Environment.Exit(-1);
```
goaaats, line 216: This is gone now, I can't see any other place you are doing it, we should probably keep it for consistency

Dalamud/Interface/ImGuiBackend/Renderers/Dx11Renderer.cs
```diff
@@ -740,6 +741,9 @@ private void CreateFontsTexture()
         if (this.fontTextures.Count != 0)
             return;
 
+        if (Service<Framework>.GetNullable() is not { IsFrameworkUnloading: false })
```
goaaats, line 744: I'm a little suspicious of all of these checks you added in rendering code, was this something that was causing issues?

goaaats (changes requested):

Dalamud/Dalamud.cs
Haselnussbomber: Yeah it works without it too. I felt this was a better order, because it disposes some Dalamud services (including ServiceContainer) before PluginManager.

Dalamud/Interface/Internal/InterfaceManager.cs
Haselnussbomber: During development of this I managed to crash within `ReShadeDxgiSwapChainOnPresentDetour` because the hook was null and it tried to call the `Original` function in there.
I don't know why or how it happened. Just made sure the object is not set to null to avoid potential further crashes.
It's also the reason why I changed it to call `OriginalDisposeSafe` instead.

Dalamud/Service/Service{T}.cs
Haselnussbomber: This function is only called when services are constructed.
You can still resolve them, but you won't be able to instantiate new services on shutdown (due to the cancelled token, I think).

Dalamud/Interface/ImGuiBackend/Renderers/Dx11Renderer.cs
Haselnussbomber: When plugins unload, it rebuilds the FontAtlas because fonts may have been disposed.
We don't need to rebuild the font atlas anymore on shutdown.

Dalamud/EntryPoint.cs
Haselnussbomber: I will wrap `ServiceManager.UnloadAllServices` in a try/catch and still let it set `IsUnloaded = true`, so it exits out of the Framework.Destroy loop. The game will close afterwards on its own.

Edit: I did that in 48dd9d7

Aireil:
LGTM

Dalamud/Game/Framework.cs
```diff
@@ -121,8 +121,9 @@ private unsafe Framework()
     /// <inheritdoc/>
     public Task DelayTicks(long numTicks, CancellationToken cancellationToken = default)
     {
-        if (this.frameworkDestroy.IsCancellationRequested) // Going away
+        if (ServiceManager.UnloadCancellationToken.IsCancellationRequested) // Gone
             return Task.FromCanceled(this.frameworkDestroy.Token);
```
goaaats, line 123: This should probably be `return Task.FromCanceled(ServiceManager.UnloadCancellationToken.IsCancellationRequested);` since Task.FromCanceled() throws if the passed token isn't cancelled

goaaats (approved):

PR #2858, opened 2026-06-17
Fixes for windowed mode

sourpuh:
This PR fixes two unrelated windowed mode bugs. I tested both changes in local build.

1. ` ImGuiHelpers.MainViewport.Pos` is usually (0,0), but in windowed mode with multi-monitor mode enabled, it holds a non-zero position. Mouse position is its screen position, so the window's position must be subtracted to get the mouse's window position. You can see `WorldToScreen` does the inverse of this.
2. In windowed mode with multi-monitor mode disabled, Dalamud incorrectly uses ClientToScreen to compute mouse position when the mouse position is actually in screen coordinates and needs to be converted to mouse coordinates. You can see the equivalent ImGui logic here: https://github.com/ocornut/imgui/blob/9a5c070308ae97bf0f884b071d0262ae1cad26f7/backends/imgui_impl_win32.cpp#L337

### DIFF
--- a/Dalamud/Game/Gui/GameGui.cs
+++ b/Dalamud/Game/Gui/GameGui.cs
@@ -174,7 +174,7 @@ internal sealed unsafe class GameGui : IInternalDisposableService, IGameGui
             return false;
         }
 
-        var ray = camera->ScreenPointToRay(screenPos);
+        var ray = camera->ScreenPointToRay(screenPos - windowPos);
         var result = BGCollisionModule.RaycastMaterialFilter(ray.Origin, ray.Direction, out var hit);
         worldPos = hit.Point;
         return result;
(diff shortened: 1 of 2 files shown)


goaaats:
This recently came up and this one was also noted as suspicious, it probably needs to be swapped: https://github.com/sourpuh/Dalamud/blob/7a3e4c806dd05ea1edf6ec99bc0a2358fc2828f3/Dalamud/Interface/ImGuiBackend/InputHandler/Win32InputHandler.cs#L250
Also ScreenToClient in the official backend

sourpuh:
I think that one is correct because I've never had any hover position problems in all of my testing.

sourpuh:
> Also ScreenToClient in the official backend

I'm not sure what is this referring to

goaaats (approved):

PR #2859, opened 2026-06-20
Change WorldToScreen to use the active camera

Abyeon:
Control.Instance()->ViewProjectionMatrix was inaccurate with DLSS enabled, so simply calculate the view projection matrix from the active camera.


Abyeon:
I've been made aware that this change causes issues for ImGui overlays as the view matrix is ahead a frame. Ideally we would have a fix that addresses the DLSS issues and keeps the same frame delay as the original function (with a toggle to use the next frame info).

PR #2865, opened 2026-06-24
Add Name to ConfigChangeEvent

Haselnussbomber:
Using the name as string is safer to check than the enum that might not have been updated.
The IGameConfig API allows getting config options by string, but the ConfigChangeEvent only contains the enum value.
This PR adds the name as string to the ConfigChangeEvent.

I'm not sure if this is API breaking?


Haselnussbomber:
"So easy."

PR #2867, opened 2026-06-24
Optimize UnlockState.Update a tiny bit

Haselnussbomber:
This PR adds a public `Debouncer` to `Dalamud.Utilities`, including a handy function `IFramework.CreateDebouncer`, and makes use of it in `UnlockState` to prevent successive `Update` calls.

The `UnlockState.Update` function is called at least twice during login: first because of the Login event, and a second time because `AgentUpdateFlag.UnlocksUpdate` was set.

The Debouncer makes sure that the Update function is called once with a delay of 500ms after the last call.

I've also updated the `IsItemUnlocked` function to check if the item is unlockable in the first place, because falling back to the games native functions would have it to load the Item sheet pages which takes quite long and has no benefit at all anyway since the items are not unlockable.

### DIFF
--- a/Dalamud/Plugin/Services/IFramework.cs
+++ b/Dalamud/Plugin/Services/IFramework.cs
@@ -2,6 +2,7 @@ using System.Threading;
 using System.Threading.Tasks;
 
 using Dalamud.Interface.Internal.Windows.Data.Widgets;
+using Dalamud.Utility;
 
 namespace Dalamud.Plugin.Services;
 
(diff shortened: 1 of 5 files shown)


Dalamud/Plugin/Services/IFramework.cs
```diff
@@ -290,4 +291,13 @@ public interface IFramework : IDalamudService
...
     public Task RunOnTick(Func<Task> func, TimeSpan delay = default, int delayTicks = default, CancellationToken cancellationToken = default);
+
+    /// <summary>
+    /// Creates a new <see cref="Debouncer"/> instance.
+    /// </summary>
+    /// <param name="delay">The delay to wait after the last request before executing the action.</param>
+    /// <param name="action">The delegate to execute when the debounce period elapses.</param>
+    /// <returns>A new, thread-safe <see cref="Debouncer"/> instance. The caller is responsible for disposing of this instance.</returns>
+    /// <exception cref="ArgumentNullException">Thrown when <paramref name="action"/> is <see langword="null"/>.</exception>
+    public Debouncer CreateDebouncer(TimeSpan delay, Action action);
```
goaaats: Technically this should be an interface (`IDebouncer`) since we want to hide implementations as much as possible. People will not be able to mock this (though I guess they could pass in their own IFramework).

goaaats (changes requested):

KazWolfe (dismissed):

goaaats (approved):

PR #2869, opened 2026-06-25
Switch from nuke to cmake

Haselnussbomber:


build.ps1
```diff
@@ -65,5 +66,10 @@ else {
 
 Write-Output "Microsoft (R) .NET Core SDK version $(& $env:DOTNET_EXE --version)"
 
-ExecSafe { & $env:DOTNET_EXE build $BuildProjectFile /nodeReuse:false /p:UseSharedCompilation=false -nologo -clp:NoSummary --verbosity quiet }
-ExecSafe { & $env:DOTNET_EXE run --project $BuildProjectFile --no-build -- $BuildArguments }
+Set-Location $BuildDirectory
+cmake .. -A x64
+cmake --build . --config Release --parallel $env:NUMBER_OF_PROCESSORS
+Set-Location $PSScriptRoot
```
nebel: If the above commands fail (due to missing `cmake` or presumably other reasons) the shell will be left in the `\build` subdirectory forcing a manual `cd ..` to try again. Maybe something like this?

```powershell
Push-Location $BuildDirectory
try {
    cmake .. -A x64
    cmake --build . --config Release --parallel $env:NUMBER_OF_PROCESSORS
} finally {
    Pop-Location
}
```

.gitignore
```diff
@@ -281,26 +368,22 @@ node_modules/
...
-.paket/paket.exe
+**/.paket/paket.exe
 paket-files/
 
 # FAKE - F# Make
-.fake/
-
-# JetBrains Rider
-.idea/
-*.sln.iml
```
nebel, line 292: Probably want to keep these Rider ones ignored?

build.ps1
```diff
@@ -65,5 +66,10 @@ else {
 
 Write-Output "Microsoft (R) .NET Core SDK version $(& $env:DOTNET_EXE --version)"
 
-ExecSafe { & $env:DOTNET_EXE build $BuildProjectFile /nodeReuse:false /p:UseSharedCompilation=false -nologo -clp:NoSummary --verbosity quiet }
-ExecSafe { & $env:DOTNET_EXE run --project $BuildProjectFile --no-build -- $BuildArguments }
+Set-Location $BuildDirectory
+cmake .. -A x64
+cmake --build . --config Release --parallel $env:NUMBER_OF_PROCESSORS
```
nebel: Obviously would want to allow passing stuff in here eventually for target/clean etc. but can do that later.

build.ps1
Haselnussbomber: Added!

.gitignore
Haselnussbomber: Appended https://github.com/github/gitignore/blob/main/Global/JetBrains.gitignore to .gitignore.

build.ps1
Haselnussbomber: I've added that back in, but the question is do we pass it to cmake or to dotnet build??

.gitignore
nebel: Seems like with all these I still get untracked files:

```
$ git status -u
On branch cmake
Your branch is up to date with 'cmake'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        .idea/.idea.Dalamud/.idea/indexLayout.xml
        .idea/.idea.Dalamud/.idea/projectSettingsUpdater.xml
        .idea/.idea.Dalamud/.idea/vcs.xml
```

Unless we have official stuff in `.idea` I think better to just ignore it completely.

PR #2874, opened 2026-07-04
Add Native Overlays

MidoriKami:
This PR adds INativeOverlay service, NativeOverlay implementation, NativeOverlayPluginScoped implementation, IOverlayNode interface, OverlayAddon implementation.

This new service allows devs to attach and remove anything that implements INativeOverlay for atthing any kind of AtkResNode or derived type to a dalamud managed overlay layer.

Use of KamiToolKit is not required for devs to utilize this API, as you can attach any hand-made AtkResNode or derived type.

Also included are all the files required to make this possible, most of these are from KamiToolKit and tweaked and adjusted to fit better into dalamud.

All extra included files are marked internal to not pollute user space.

No clientstructs types are exposed by dalamud or this new API.

My intention with including these files directly into dalamud is threefold,
1. This allows dalamud and its maintainers full control of the API
2. Implement native overlay addons that can be shared and managed at a much more sane level within dalamud than within KTK
3. Allow dalamud to begin utilizing more advanced native ui tech for its various features, such as IDtrBar.

AddonConfig backing data has been added to the main DalamudConfiguration file as a Dictionary<string (InternalName), AddonConfig (Postion/Scale)> when any dalamud owned native windows are opened or closed they will reference this backing data.


MidoriKami:
Updated PR with AddonConfig system implemented, and improved performance when detaching nodes by not triggering excess drawlist or collision updates. (This code has been working in KTK for over a month without a single issue reported)

PR #2880, opened 2026-07-10
Add xivfix for faster decompression

Haselnussbomber:
This adds a xivfix to replace the `zlib_uncompress` function used in the `ReadSqpkChunk` function from the game with a detour doing the same thing with https://github.com/ebiggers/libdeflate instead, which is around 40-60% faster in decompressing sqpack chunks (which sounds much, but is actually like 0,2-0,5s per loading screen - depends on the hardware, I'd say).

Edit: Replaced the calling function `ReadSqpkChunk` instead. The game copys/deflates the block to a buffer and then memcpys it to the destination. Our implementation just memcpys the uncompressed data or deflates the compressed data directly to the destination.

---

There was a discussion in the Dalamud Discord that this should be be a "xivenhancement" instead of a xivfix, meaning the launcher would get a new window that makes it easy to toggle features on/off.
While I do agree that this is a good idea, it is simply out of scope for this PR.
Once this feature toggle system is added, I'm happy to submit a PR to integrate faster_decompression into it, if needed.

### DIFF
--- a/Dalamud.Boot/dllmain.cpp
+++ b/Dalamud.Boot/dllmain.cpp
@@ -509,6 +509,7 @@ BOOL APIENTRY DllMain(const HMODULE hModule, const DWORD dwReason, LPVOID lpRese
 
     switch (dwReason) {
         case DLL_PROCESS_ATTACH:
+            libdeflate_set_memory_allocator(mi_malloc, mi_free);
             g_hModule = hModule;
             break;
 
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,9 @@
 [submodule "lib/Lumina.Excel"]
 	path = lib/Lumina.Excel
 	url = https://github.com/NotAdam/Lumina.Excel.git
+[submodule "lib/mimalloc"]
+	path = lib/mimalloc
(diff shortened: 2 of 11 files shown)


KazWolfe (dismissed):

KazWolfe (approved):

Dalamud.Boot/dllmain.cpp
```diff
@@ -563,6 +509,7 @@ BOOL APIENTRY DllMain(const HMODULE hModule, const DWORD dwReason, LPVOID lpRese
 
     switch (dwReason) {
         case DLL_PROCESS_ATTACH:
+            libdeflate_set_memory_allocator(mi_malloc, mi_free);
```
goaaats, line 512: This probably should have either been in the xivfix init or have a comment. I assume you have to call it before any other libdeflare calls so that should be fine?

Haselnussbomber: Ah yeah, that would've been better in the xivfix.
But it's just setting static variables anyway: https://github.com/ebiggers/libdeflate/blob/master/lib/utils.c#L65-L66

PR #2881, opened 2026-07-13
Add NVIDIA Smooth Motion Compatibility

ConfoundedHermit:
# Add NVIDIA Smooth Motion Compatibility

## Related Issue

https://github.com/goatcorp/Dalamud/issues/2190

## Credit

Based on `goaaats` work in https://github.com/goaaats/Dalamud/tree/fix/smooth_motion_v2

## Purpose

Implement compatibility with NVIDIA Smooth Motion by decoupling Dalamud's ImGui frame construction from the actual `Present` calls that Smooth Motion may invoke from NVIDIA worker threads.

Smooth Motion can cause Dalamud's normal DXGI present hook to be entered multiple times per game tick, and not necessarily from the game thread. This PR makes Dalamud build ImGui once at a stable point in the game swap-chain present path, deep-copy the resulting draw data, and render those stable snapshots from later present calls without re-running Dalamud/plugin UI logic.

The implementation also unwraps `NvPresent` swap-chain wrappers, enables D3D11 multithread protection specifically *when Smooth Motion is detected*, serializes render-sensitive work against the native framework tick, and safely retires resources only after all render passes for the current frame have completed.

## Testing

Over 100+ hours of user testing - including various plugin setups and usages and with multiple users on different systems - has been performed on this implementation, with updated testing performed after every commit.

### Added

- Added `DrawDataSnapshot` to deep-copy ImGui draw data into Dalamud-owned unmanaged memory so render threads can read a stable frame while the next ImGui update is being prepared.
- Added `ViewportSnapshot` to capture secondary viewport/pop-out draw data and renderer handles, allowing viewport windows to be rendered without walking live ImGui platform state from Smooth Motion worker threads.
- Added an `IImGuiBackend.Step()`/`Render()` split, a `PostCopy` event, and `EnterResize()`/`ExitResize()` coordination methods so frame construction, draw-data copying, rendering, resource retirement, and resize exclusion can be managed independently.
- Added renderer support for drawing and presenting secondary viewport snapshots.
- Added a generalized COM hook unwrapper with dedicated `NvPresentUnwrapper` and `ReShadeUnwrapper` implementations.
- Added NvPresent detection/unwrapping in `SwapChainHelper`, including a separate `GameDisplaySwapChain` concept for the swap chain actually presenting to the game window.
- Added a hook for `Client::Graphics::Kernel::SwapChain::Present()` so Dalamud can step ImGui at a fixed point even when DXGI present is later driven by Smooth Motion.
- Added frame-retirement callbacks through `InterfaceManager.DeferUntilFrameRetired()` so D3D/unmanaged resources used by a captured frame are released only after that frame is no longer being rendered.
- Added shared framework/render synchronization via `ThreadSafety.NativeFrameworkRenderSyncRoot`.

### Modified

- Reworked `Dx11Win32Backend` so `Step()` runs ImGui update/build once, snapshots main and secondary viewport draw data under a write lock, and `Render()` submits the captured snapshots under a read lock.
- Serialized Smooth Motion render calls with a backend render lock and enabled D3D11 multithread protection when NvPresent is detected.
- Made swap-chain resize handling hold an exclusive backend write lock, skip render/step work during resize, clear stale viewport snapshots, and retire resources sized for the old swap chain before buffers are recreated.
- Updated `InterfaceManager.RenderDalamudDraw()` to tolerate present calls from any thread, track `IsAnyThreadInPresent`, and serialize pre-render actions, snapshot rendering, and post-render actions.
- Updated setup logic to repeatedly unwrap ReShade and NvPresent wrappers when both are present, detect Smooth Motion, hook the display swap chain, and pass Smooth Motion state into the backend.
- Replaced end-of-frame deferred disposal bags with frame-retirement queues so textures, font locks, blur callback data, and other render resources survive all Smooth Motion re-renders of the captured frame.
- Updated blur callback pooling to be thread-safe and moved blur callback data returns to frame retirement.
- Updated texture and viewport texture handling to use `IsAnyThreadInPresent` and the unwrapped display swap chain instead of assuming only the main thread/game device swap chain is presenting.
- Hardened ReShade peeling and COM unwrapping against cycles, invalid memory ranges, and stale module enumeration.
- Wrapped native framework update rendering synchronization so Smooth Motion worker renders do not overlap native UI update/finalize work.
- Ensured addon lifecycle listener invocation always clears its `isInvokingListeners` flag with `try/finally`.
- Updated current process module caching to use the newer cached-property pattern while still refreshing when notified by boot support code.


Dalamud/Utility/ThreadSafety.cs
```diff
@@ -15,6 +15,16 @@ public static class ThreadSafety
...
 
+    /// <summary>
+    /// Gets the shared lock that prevents worker-thread rendering from overlapping the original native framework
+    /// update.
+    /// </summary>
+    /// <remarks>
+    /// This reentrant monitor allows the game thread to present during the native update.
+    /// The managed framework tick runs before this lock is acquired and is not protected by it.
+    /// </remarks>
+    internal static object NativeFrameworkRenderSyncRoot { get; } = new();
```
Haselnussbomber:
```suggestion
    internal static Lock NativeFrameworkRenderSyncRoot { get; } = new();
```

This needs a `using System.Threading;` at the top of the file.

Dalamud/Game/Framework.cs
```diff
@@ -416,7 +416,13 @@ private unsafe bool HandleFrameworkUpdate(CSFramework* thisPtr)
     {
         this.RunFrameworkTick();
 
-        return this.updateHook.OriginalDisposeSafe(thisPtr);
+        // Smooth Motion can ask Dalamud to render from NvPresent worker threads while the game thread is inside
+        // the native framework tick. Keep those worker renders out of native UI update/finalize work; the game
+        // thread can still re-enter this monitor for the real present that happens during the tick.
+        lock (ThreadSafety.NativeFrameworkRenderSyncRoot)
```
Haselnussbomber:
```suggestion
        using (ThreadSafety.NativeFrameworkRenderSyncRoot.EnterScope())
```

Dalamud/Interface/ImGuiBackend/Dx11Win32Backend.cs
```diff
@@ -162,14 +235,128 @@ public void Render()
...
+    }
+
+    /// <inheritdoc/>
+    public void Render()
+    {
+        // Skip work while resize is announced; the write lock supplies snapshot exclusion.
+        if (this.resizeInProgress)
+            return;
+
+        lock (this.renderLock)
```
Haselnussbomber:
```suggestion
        using (this.renderLock.EnterScope())
```

Dalamud/Interface/Internal/InterfaceManager.cs
```diff
@@ -655,36 +683,50 @@ private unsafe void FrameworkOnUpdate(IFramework framework1)
     /// <param name="activeBackend">The scene to draw to.</param>
     private void RenderDalamudDraw(IImGuiBackend activeBackend)
     {
-        this.CumulativePresentCalls++;
-        this.IsMainThreadInPresent = true;
+        // Presentation callbacks may run on multiple threads. Serialize queued actions and live ImGui access
+        // with drawing, and exclude the original native framework update through the shared monitor.
+        // Acquire NativeFrameworkRenderSyncRoot, renderDalamudLock, backend renderLock, then the snapshot read lock.
+        lock (ThreadSafety.NativeFrameworkRenderSyncRoot)
```
Haselnussbomber:
```suggestion
        using (ThreadSafety.NativeFrameworkRenderSyncRoot.EnterScope())
```

Dalamud/Interface/Internal/InterfaceManager.cs
```diff
@@ -655,36 +683,50 @@ private unsafe void FrameworkOnUpdate(IFramework framework1)
...
+        {
+            lock (this.renderDalamudLock)
```
Haselnussbomber:
```suggestion
            using (this.renderDalamudLock.EnterScope())
```

Haselnussbomber (commented):
Hey, I can't comment on the Smooth Motion part of the PR because I don't have the expertise, I just have a small nitpick:

In the past (#2839), I've changed everything in Dalamud from using the `lock` keyword to make use of [`System.Threading.Lock`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.lock)s `EnterScope` function.

Even though the `lock` keyword on a `Lock` object is translated by the compiler into using the `EnterScope` function, it ensures we don't accidentally `lock` on an `object` and run into C# using the old, slow `Monitor.Enter` function under the hood.

Would be nice if you could change this. :)

Please note that `Dalamud/Utility/ThreadSafety.cs` needs a `using System.Threading;` at the top of the file, so it would probably be better if you change this locally in case there are any compilation errors or so. 😇

ConfoundedHermit:
Addressed the items from the last review comment.

Dalamud/Interface/Internal/InterfaceManager.cs
ConfoundedHermit: Addressed in commit https://github.com/goatcorp/Dalamud/pull/2881/commits/e577d6e59b7b8e7378d4f259797db844fae0fdde

Dalamud/Interface/Internal/InterfaceManager.cs
ConfoundedHermit: Addressed in commit https://github.com/goatcorp/Dalamud/pull/2881/commits/e577d6e59b7b8e7378d4f259797db844fae0fdde

Dalamud/Interface/ImGuiBackend/Dx11Win32Backend.cs
ConfoundedHermit: Addressed in commit https://github.com/goatcorp/Dalamud/pull/2881/commits/e577d6e59b7b8e7378d4f259797db844fae0fdde

Dalamud/Game/Framework.cs
ConfoundedHermit: Addressed in commit https://github.com/goatcorp/Dalamud/pull/2881/commits/e577d6e59b7b8e7378d4f259797db844fae0fdde

Dalamud/Utility/ThreadSafety.cs
ConfoundedHermit: Addressed in commit https://github.com/goatcorp/Dalamud/pull/2881/commits/e577d6e59b7b8e7378d4f259797db844fae0fdde

PR #2883, opened 2026-07-15
Always take frame padding into account for title bar buttons

Exter-N:
Before:
[Enregistrement d'écran_20260715_083158.webm](https://github.com/user-attachments/assets/c6b38f7e-518d-4930-8585-f3cfaed004b1)

After:
[Enregistrement d'écran_20260715_083826.webm](https://github.com/user-attachments/assets/840800c6-33d0-4640-91d6-6f544f5feb5a)

(Yes, :baguette_bread: file names)

Also exposes the `EnablePluginUiAdditionalOptions` setting so that plugins can calculate where custom title bar buttons will be laid out.

### DIFF
--- a/Dalamud/Interface/UiBuilder.cs
+++ b/Dalamud/Interface/UiBuilder.cs
@@ -238,6 +238,12 @@ public interface IUiBuilder
     bool PluginUISoundEffectsEnabled { get; }
 
     /// <summary>
+    /// Gets a value indicating whether the user has enabled the "Add a button to the title bar of plugin windows to
+    /// open additional options" setting.
+    /// </summary>
+    bool PluginUIAdditionalOptionsEnabled { get; }
+
+    /// <summary>
     /// Loads an ULD file that can load textures containing multiple icons in a single texture.
     /// </summary>
     /// <param name="uldPath">The path of the requested ULD file.</param>
@@ -588,6 +594,9 @@ public sealed class UiBuilder : IDisposable, IUiBuilder
 
     /// <inheritdoc />
     public bool PluginUISoundEffectsEnabled => Service<DalamudConfiguration>.Get().EnablePluginUISoundEffects;
+
+    /// <inheritdoc />
+    public bool PluginUIAdditionalOptionsEnabled => Service<DalamudConfiguration>.Get().EnablePluginUiAdditionalOptions;
 
     /// <summary>
     /// Gets or sets a value indicating whether statistics about UI draw time should be collected.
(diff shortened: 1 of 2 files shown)


Dalamud/Interface/UiBuilder.cs
```diff
@@ -237,6 +237,12 @@ public interface IUiBuilder
     /// </summary>
     bool PluginUISoundEffectsEnabled { get; }
 
+    /// <summary>
+    /// Gets a value indicating whether the user has enabled the "Add a button to the title bar of plugin windows to
+    /// open additional options" setting.
+    /// </summary>
+    bool PluginUIAdditionalOptionsEnabled { get; }
```
goaaats, line 244: Is this so that you can draw additional stuff around the buttons? Would it not be easier to be able to get the final positions/size of the buttons instead?

Exter-N: Here's an example of what I'd like to achieve without an ugly positioning hack that only works if the user's configuration is sufficiently similar to mine - note the bouncing chevron that calls the user's attention on the button:
[Enregistrement d'écran_20260715_013331.webm](https://github.com/user-attachments/assets/e3bf34ff-6bbf-4ecd-beb6-ddae69d1127b)

The issue with exposing `ScreenBounds` just like I had done on DTR bar entries (see excerpt below) is that title bar buttons are currently laid out and drawn after the window's `.Draw()` returns, and therefore, for buttons that are specific to a window, I'd get the bounds with a frame of latency, and for reused ones, I'd get the bounds the button has (or had) in another window.
https://github.com/goatcorp/Dalamud/blob/9fa2038de4ed6b816ca4fabfe6c9e8977825d6ea/Dalamud/Game/Gui/Dtr/DtrBarEntry.cs#L58-L61

And I'm not confident about changing the draw order to have the buttons drawn before the window's body, I don't know the implications this would have.

goaaats: Sorry for the late reply. I don't see anything inherently wrong with changing this, other than that events would be dispatched before the end of the window instead if we don't buffer them. I'm only bringing it up since I feel like it would be a lot nicer to have the final values instead of having to mirror the exact calculations we do in the plugin.

If you disagree, let me know, I haven't looked at the code. Fine merging this anyway.

Exter-N: I don't disagree, I'll have to try changing it.

One implication it would have is that you wouldn't be able to add/remove/reorder your window's TBBs within its `Draw` function. But then again, I'm not sure this is exactly a common thing to do.

A thing I can also do is make the TBB rendering two-pass: one pass that runs before the window's `Draw`, performs the layout, and populates `ScreenBounds`, and the second pass that stays after the window's `Draw`, and performs the actual drawing at the bounds calculated by the first pass. This way, the rendering order, and the event order, stay the same, and the only BC break is TBB additions/removals/reorders within the window's `Draw` function only being taken into account starting from the next frame.

PR #2891, opened 2026-07-24
Split up DalamudAtkTweaks into separate files

Haselnussbomber:
Yep. Just a little refactor.

I also moved the code for the Dalamud version and plugin count for the title screen into its own class, TitleScreenVersionInfo, but I had to get rid of this `|| !this.showTsm.Value` condition, which is seemingly for a console command that I didn't know existed.
In exchange for this inconvenience, I made it properly reset the string to its original version (pun intended) when Dalamud unloads (kinda depends on #2857 because it just freezes currently, wink wink - well, to be clear: this PR does not depend on the other PR though).


goaaats (approved):
Thanks!

PR #2895, opened 2026-07-26
Filter Source out of plugin-specific regex filter

mashirochan:
With change, only log entries actually containing the filter content in the message will display:
image

Without change, every single log entry with the plugin internal name/source will display:
image


mashirochan:
It certainly mostly deals with edge cases, but perhaps a better example were to be if the plugin were InternallyNamed "BatteryBuddy" and I was trying to search through the logs for battery-related entries. Searching "battery" wouldn't help at all with this new filter, because I'd just get (most) every line from the plugin, BatteryBuddy. The plugin filter already filters the logs by plugin, so you'd never want to use the regex filter to see all logs from the plugin!

PR #2896, opened 2026-07-27
Fix Framework shutdown

Haselnussbomber:
`ServiceManager.UnloadCancellationToken` is cancelled as soon as `ServiceManager.UnloadAllServices` is called in `Dalamud.Unload`, which causes all Framework tasks to be cancelled and no longer run during the Update tick (thanks to my previous PR #2857).
I made a new token, `frameworkDestroyed` in the Framework class that is cancelled when the service is disposed.


KazWolfe (approved):
yeah sure ok

PR #2904, opened 2026-08-02
Fix MinHook enabling

Drovolon:
I wanted to test MinHook on Linux, and found that it didn't work. It looks like it regressed in 2c735e9ec on May 9, 2025.

Tested locally with `DALAMUD_FORCE_MINHOOK=true`. Hooks appear to work with this change. /xlstats shows MinHook backend, title screen has the `mh!` indicator.


goaaats:
We decided to remove these with #2905 since we were planning on removing them anyway, but thanks!

PR #2907, opened 2026-08-04
Restore listings hidden by the price comparison window

Vycton:
A summoned retainer's listings are withheld from the price comparison window. A snapshot taken there is missing them, so Universalis diffs it against stored state and publishes listings/remove for listings that are still live. On rarely-uploaded items nothing contradicts that and the removal sticks - measured at a 1.40% failure rate over 4631 remove probes against 0.04% over 2501 add probes, across 6 worlds. It also identifies the uploader: a player adjusting prices leaves a trail of removes on their own retainer's listings, which is how the cause was originally traced to a specific seller.

When a listings batch resolves while a retainer is summoned, that item's current listings are fetched from the v2 REST API and any belonging to the retainer which the snapshot does not already carry are added back. Universalis' own rows are used rather than local knowledge because the game never tells a client what listing IDs its listings were assigned; a reconstructed ID would post as an add and a remove instead of a no-op.

The insert is keyed on listing ID, which covers two cases: a listing the snapshot already carries, and a second copy of one the API returned twice - it does that occasionally, and filtering against the snapshot alone uploaded each restored listing twice. A failed or unusable fetch logs a warning and uploads unmodified.

The retainer is read from RetainerManager rather than inferred from packets, and it is read when the batch resolves rather than inside Upload: the upload is deferred to a Task, so by the time it runs the player may have dismissed the retainer. This matches how uploaderId and worldId are already captured.

Code written by Claude. Hand-checked and tested in-game on 7.55: no retainer is reported for an ordinary market board browse, and a comparison-window snapshot uploads with the retainer's listings restored at their real listing IDs.

The equivalent fix for the ACT plugin / standalone uploader is [goaaats/universalis_act_plugin#142](https://github.com/goaaats/universalis_act_plugin/issues/142). Same approach, agreed with karashiiro; this one needs no coordination to land, since the retainer comes from RetainerManager rather than an opcode that has to be published first.

### DIFF
--- a/Dalamud/Game/Network/Internal/MarketBoardUploaders/IMarketBoardUploader.cs
+++ b/Dalamud/Game/Network/Internal/MarketBoardUploaders/IMarketBoardUploader.cs
@@ -15,8 +15,9 @@ internal interface IMarketBoardUploader
     /// <param name="item">The item request data being uploaded.</param>
     /// <param name="uploaderId">The uploaders ContentId.</param>
     /// <param name="worldId">The uploaders WorldId.</param>
+    /// <param name="activeRetainerId">The retainer summoned when the listings were observed, or null if none was.</param>
     /// <returns>An async task.</returns>
-    Task Upload(MarketBoardItemRequest item, ulong uploaderId, uint worldId);
+    Task Upload(MarketBoardItemRequest item, ulong uploaderId, uint worldId, ulong? activeRetainerId);
 
     /// <summary>
     /// Upload tax rate data.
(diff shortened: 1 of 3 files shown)


goaaats:
Hi, we don't accept changes made autonomously using generative AI and all changes must be submitted as yourself. Please rewrite the change, attribute it to yourself and resubmit it.

Vycton:
Thanks for your comment. Is it just the fact that the commit was made by Claude? Or do you have a more general no AI policy? If I rewrite the change it's going to end up being exactly the same, since I instructed Claude to write it exactly how I think it should be written.

PR #2912, opened 2026-08-10
Update EmbeddedInfoType

Haselnussbomber:
There are 2 new LinkMacroPayloadTypes: EventTutorial (14) and Emote (15).

DalamudLink currently uses 14, blocking EventTutorial links. I've moved it to 1 before LinkTerminator. This will require plugins to be rebuilt, if they use the enum value directly.

The new types were merged to Lumina (https://github.com/NotAdam/Lumina/pull/132), but there was no release yet.


KazWolfe:
Given we broke native functionality, gonna bring this in.

@Caraxi you may need to recompile Simple Tweaks per a GitHub code search, apologies!

KazWolfe (approved):

PR #2919, opened 2026-08-14
Use GetName function in GameObject

Haselnussbomber:
The Name field is mostly filled by a packet from the server, which might contain an untranslated name.
The `GetName()` function instead resolves the applicable NameId/EventHandlerNameId/TransformationNameId field using the clients language.


KazWolfe:
Would this be considered a behavioral change? If so, this might be an API bump.

Haselnussbomber:
~~I guess? I don't use this API. I just know it's not returning the name like that, because of my experiences with GameObjects. So this is more a bugfix than a behavior change.~~

Bosses can transform into other forms and change names, for example the last boss in Dohn Mheg. The Name field would have the original name, whereas the GetName function would also return the transformed name, if available.

KazWolfe (approved):

PR #2921, opened 2026-08-14
Allow plugin windows to be closed with a gamepad

Haselnussbomber:
Pressing the B (XBOX Controller)/Circle (PlayStation Controller) key will back out of children/inputs, but will not close the window.
This PR changes this to close the focused window on the top level. Excluded are windows that are pinned.

### DIFF
--- a/Dalamud/Interface/Windowing/WindowHost.cs
+++ b/Dalamud/Interface/Windowing/WindowHost.cs
@@ -555,6 +555,12 @@ public class WindowHost
             }
         }
 
+        // Allow the window to be closed with a gamepad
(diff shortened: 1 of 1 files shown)


goaaats:
I'm wondering, how does this differ from the logic we have for Escape? Why does it need to be duplicate, is it not the same action if navigation is on the window level and deserves the same handling (i.e. in case of overlays)?

https://github.com/goatcorp/Dalamud/blob/baa476a1560b231e85e7d5076c4dcd0ec3c25559/Dalamud/Interface/Windowing/WindowHost.cs#L546

Haselnussbomber:
Hitting ESC to close the focused window is a feature part of the optional focus management setting in Dalamud.
It can be done from any navigation level, except when ImGui inputs are active, I think.
If focus management is not enabled, the user can still close the window by hitting X with the mouse cursor.

But controller users simply cannot close the window *at all*, without putting down their controller and using the mouse to click the X, or to hit ESC when focus management is enabled.
I just thought it should be baseline functionality, which is why it's separate.

goaaats (approved):

PR #2922, opened 2026-08-14
Add gamepad support to NotificationPositionChooser

Haselnussbomber:
The NotificationPositionChooser could only be controlled with mouse and keyboard.
This PR adds gamepad support to it, so that the notification position can be set with the left stick, and the overlay can properly be closed either with or without saving.
Also added special instructions when gamepad mode is active and made them translatable. I would‘ve loved to use gamepad button icons in the texts, but I‘m not sure how that would work with the translations.

Example video using the mouse first, then the controller:

https://github.com/user-attachments/assets/a5c6a33f-169e-470f-82bf-2078eb207164


goaaats (approved):
Sweet!

PR #2924, opened 2026-08-14
Change GamepadModeNotifierWindow sounds

Haselnussbomber:
This backdrop overlay played the default window open/close sounds.
I've changed it to use the sounds of the games Virtual Mouse Mode for funsies.

Would probably require some feedback from controller users, as I don't use controllers to play the game.
If this is too confusing or annoying, we can also disable the sounds or pick some other ones.

Preview (Virtual Mouse Mode vs. GamepadModeNotifierWindow - sound on!):

https://github.com/user-attachments/assets/2f981347-52a2-4c6a-a6c9-5d607c258c81


Haselnussbomber:
At least one person already said they like how it is right now, so I'm closing this.
Also it would be hard to differenciate the gamepad mode enabling from the Virtual Pad Mode enabling if you're blind.

PR #2930, opened 2026-08-19
Fixed numpad return overflow behaviour

Glorou:
Virtual Key needs to be backed by a byte, not a ushort due to behaviour surrounding numpad return, which the virtual key for is 269, this is due to the fact that it is intended to overflow a byte and wrap around to 0x0D, which is return.


goaaats:
I'm not sure this is the right way to fix this problem, as it breaks the ABI and there is no real guarantee in usage code.
Where are you actually running into problems when using this type?

PR #2936, opened 2026-08-24
Log Framework.Update hitches individually

Haselnussbomber:
Currently, `Framework.Update` hitches are detected and logged as a whole, not giving any information about which plugins `IFramework.Update` handler actually caused it. Similarly, the error handler does not log which function caused the error, instead it just mentions `IFramework.Update`:

Screenshot before changes

This PR changes the internal code, so that hitches and errors are logged per-delegate:

Screenshot after changes

The error notification is also still shown and stats are only recorded when the checkbox is ticked in the /xlstats window.

---

Code used for demonstration:

```cs
public class HitchTab : DebugTab, IDisposable
{
    private readonly IFramework _framework;
    private bool _doHitch;
    private bool _doError;

    public HitchTab(IFramework framework)
    {
        _framework = framework;
        _framework.Update += OnUpdate;
    }

    public void Dispose()
    {
        _framework.Update -= OnUpdate;
    }

    private void OnUpdate(IFramework framework)
    {
        if (_doHitch)
        {
            Thread.Sleep(500);
            _doHitch = false;
        }
        if (_doError)
        {
            _doError = false;
            throw new InvalidOperationException("This is a test exception.");
        }
    }

    public override void Draw()
    {
        _doHitch |= ImGui.Button("Hitch");
        _doError |= ImGui.Button("Error");
    }
}
```

### DIFF
--- a/Dalamud/Game/Framework.cs
+++ b/Dalamud/Game/Framework.cs
@@ -619,7 +619,7 @@ internal class FrameworkPluginScoped : IInternalDisposableService, IFramework
     {
         this.frameworkService.ProfileAndInvoke(this.Update, this, (ex, handlerName) =>
         {
-            Serilog.Log.Error(ex, "[{InternalName}] Exception in event handler {EventHandlerName}", this.plugin.InternalName, handlerName);
+            Serilog.Log.Error(ex, $"[{this.plugin.InternalName}] Exception in event handler {{EventHandlerName}}", handlerName);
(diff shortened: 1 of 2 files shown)


Haselnussbomber:
Implemented a suggestion by @MidoriKami to also log hitches from Dalamud-internal subscribers.

Screenshot

KazWolfe (approved):

PR #2940, opened 2026-09-01
DataAdapter updates

Ottermandias:
- Add alive checks and disposal event to IIdDataShareAdapter
- Entirely remove INameDataShareAdapter. According to github code search, nobody uses it, and it is a lot of code duplication just to use string names instead of enums due to arity concerns.


Dalamud/Plugin/Ipc/IDataShareAdapter.cs
```diff
@@ -4,6 +4,15 @@ namespace Dalamud.Plugin.Ipc;
 /// <remarks> Implement only methods you actually need. This can then be used to create a wrapper encapsulating the actually available methods or properties, either by the provider library, or on the consumer side. </remarks>
 public interface IIdDataShareAdapter : IDisposable
 {
+    /// <summary> An event that should be invoked when the adapter is disposed. </summary>
+    event Action? Disposed;
+
+    /// <summary> Gets the version of this adapter. </summary>
+    Version Version { get; }
```
KazWolfe: Can this be explicitly nullable?

KazWolfe (dismissed):

goaaats (dismissed):

goaaats (approved):

PR #2941, opened 2026-09-01
Add more information about active plugin changes to the event.

Ottermandias:
- I avoided to use IExposedPlugin since this contains data that does not make sense for a now-uninstalled plugin, and also references to objects that should only be available in a live plugin.
- So this creates most of the possibly relevant data and stores it without references, most importantly the WorkingPluginId, which is the only way to uniquely identify a plugin across its lifetime (due to dev plugins sharing internal names, possibly).

### DIFF
--- a/Dalamud/Plugin/ActivePluginsChangedEventArgs.cs
+++ b/Dalamud/Plugin/ActivePluginsChangedEventArgs.cs
@@ -1,4 +1,7 @@
 using System.Collections.Generic;
+using System.Linq;
+
+using Dalamud.Plugin.Internal.Types;
 
 namespace Dalamud.Plugin;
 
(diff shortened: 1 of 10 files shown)


Dalamud/Plugin/IActivePluginsChangedEventArgs.cs
```diff
@@ -1,12 +1,42 @@
...
+
 namespace Dalamud.Plugin;
 
 /// <summary>
 /// Contains data about changes to the list of active plugins.
 /// </summary>
 public interface IActivePluginsChangedEventArgs
 {
+    /// <summary> Contains a subset of the data of <see cref="IExposedPlugin"/> that is available even if the plugin has been unloaded. </summary>
+    public interface IAffectedPlugin
```
KazWolfe, line 13: [API16] we should probably clean up the types to create a clear(er) inheritance path between everything.

Dalamud/Plugin/ActivePluginsChangedEventArgs.cs
```diff
@@ -10,16 +13,48 @@ public class ActivePluginsChangedEventArgs : EventArgs, IActivePluginsChangedEve
...
     /// <inheritdoc/>
-    public IEnumerable<string> AffectedInternalNames { get; }
+    public IEnumerable<string> AffectedInternalNames
+        => this.AffectedPlugins.Select(t => t.InternalName);
+
+    /// <inheritdoc/>
+    public IEnumerable<IActivePluginsChangedEventArgs.IAffectedPlugin> AffectedPlugins { get; }
+
+    /// <inheritdoc/>
+    internal sealed class AffectedPlugin(LocalPlugin plugin, Version? version) : IActivePluginsChangedEventArgs.IAffectedPlugin
```
KazWolfe, line 34: [API16] Ditto; should be in the inheritance tree.

KazWolfe (changes requested):
Couple API16 notes, please fix the indentation issues and I'd be happy to merge.

KazWolfe (dismissed):

goaaats (approved):

PR #2947, opened 2026-09-05
feat: add support for invalid VirtualKeys for plugin use

Bronya-Rand:
This PR does the following:
1. Handles keys that are not supported by XIV for plugin use only (F13-F24 among a few others).
    > This is done by attaching *WndProcHookManager* to **KeyState** and making a event function called `OnPreWndProc` that handles the key state of keys not supported by XIV. In addition a cache and buffer was made to handle input processing and key gathering as well as returning the ref value for *GetRefValue*.

3. Rewrote the KeyState widget into two tables that are structured as:

    ## Game Keybinds
     | Virtual Key | SeVirtualKey | Decimal | Hex | State |
     | :-----------: | :------------: | :--------: | :---: | :-----: | 
     | RETURN | RETURN (13) | 13 | 0x0D | Released |
     | HOME | HOME (36) | 36 | 0x24 | Pressed |
       
    ## Extended Keybinds
     | Virtual Key | Decimal | Hex | State |
     | :-----------: | :--------: | :---: | :-----: | 
     | F13 | 124 | 0x7C | Released |
     | F24 | 135 | 0x87 | Pressed |

4. Added `TryGetSeVirtualKey` to obtain the *SeVirtualKey* equivalent of a *VirtualKey* in `IKeyState`.
5. Added `GetExtendedVirtualKeys` to return all *VirtualKey*'s that are not supported by the game in `IKeyState`.

### DIFF
--- a/Dalamud/Game/ClientState/Keys/KeyState.cs
+++ b/Dalamud/Game/ClientState/Keys/KeyState.cs
@@ -42,12 +42,18 @@ internal class KeyState : IServiceType, IKeyState, IInternalDisposableService
     private static readonly ModuleLog Log = ModuleLog.Create<KeyState>();
     private readonly WndProcHookManager wndProcHookManager;
 
+    [ServiceManager.ServiceDependency]
+    private readonly Framework framework = Service<Framework>.Get();
+
     private readonly KeyStateAddressResolver addressResolver;
     private readonly IntPtr bufferBase;
     private readonly IntPtr indexBase;
 
     // Buffer to store the state of extended virtual keys that do not map to a in-game keycode.
     private readonly int[] extendedKeyState = new int[0x100];
+
+    // Stores the virtual key codes that have been pressed or released in the current frame.
+    private readonly HashSet<int> ephemeralKeys = [];
     private VirtualKey[]? validVirtualKeyCache;
 
     // Cache of valid virtual keys that do not map to a in-game keycode.
@@ -67,13 +73,16 @@ internal class KeyState : IServiceType, IKeyState, IInternalDisposableService
         this.wndProcHookManager = wndProcHookManager;
         this.wndProcHookManager.PreWndProc += this.OnPreWndProc;
 
+        // Clear ephemeral key states at the end of each frame
+        this.framework.Update += this.OnFrameworkUpdate;
+
         Log.Verbose($"Keyboard state buffer address {Util.DescribeAddress(this.bufferBase)}");
     }
 
     /// <inheritdoc/>
     public bool this[int vkCode]
     {
-        get => this.GetRawValue(vkCode) != 0;
+        get => (this.GetRawValue(vkCode) & (int)KeyStateFlags.Down) != 0;
         set => this.SetRawValue(vkCode, value ? 1 : 0);
     }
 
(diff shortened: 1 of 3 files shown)


Dalamud/Game/ClientState/Keys/KeyState.cs
```diff
@@ -121,21 +174,79 @@ private unsafe byte ConvertVirtualKey(int vkCode)
...
+        {
+            return ref *(int*)(this.bufferBase + (4 * gameVkCode));
+        }
 
-        if (vkCode == 0)
-            throw new ArgumentException($"Keycode state is only valid for certain values. Reference GetValidVirtualKeys for help.");
+        if (vkCode >= 0 && vkCode < this.extendedKeyState.Length &&
+            Enum.IsDefined(typeof(VirtualKey), (VirtualKey)vkCode))
+        {
+            return ref this.extendedKeyState[vkCode];
```
goaaats, line 231: As far as I can tell, this means that the "extended" keys do not have all of the flags the game's keys have (pressed, down, released, held). You should either document that or fix it if you actually want to emulate the game's API accurately. So set Pressed|Down on keydown, consume the bits per frame, Released on keyup.

Dalamud/Game/ClientState/Keys/KeyState.cs
```diff
@@ -101,15 +117,52 @@ public bool IsVirtualKeyValid(VirtualKey vkCode)
     public IEnumerable<VirtualKey> GetValidVirtualKeys()
         => this.validVirtualKeyCache ??= Enum.GetValues<VirtualKey>().Where(this.IsVirtualKeyValid).ToArray();
 
+    /// <inheritdoc />
+    public IEnumerable<VirtualKey> GetExtendedVirtualKeys()
```
goaaats, line 144: There are some keys in your "extended set" that will never be set in here, such as mouse keys, since you don't process them in the wndproc. You should either set them like the rest or exclude them.

Dalamud/Plugin/Services/IKeyState.cs
```diff
@@ -63,12 +65,29 @@ public interface IKeyState : IDalamudService
     /// <inheritdoc cref="IsVirtualKeyValid(int)"/>
     bool IsVirtualKeyValid(VirtualKey vkCode);
```
goaaats, line 64: No `IsExtendedVirtualKeyValid()`? Why split them up at all if they should behave the same?

Dalamud/Plugin/Services/IKeyState.cs
```diff
@@ -63,12 +65,29 @@ public interface IKeyState : IDalamudService
...
 
+    /// <summary>
+    /// Attempts to get the SeVirtualKey equivalent of a given virtual key code.
+    /// </summary>
+    /// <param name="vkCode">The virtual key code to convert.</param>
+    /// <param name="seVkCode">The resulting SeVirtualKey, if the conversion is successful.</param>
+    /// <returns>Whether the virtual key code has a valid SeVirtualKey equivalent.</returns>
+    bool TryGetSeVirtualKey(int vkCode, out SeVirtualKey? seVkCode);
```
goaaats: What do you need this for? It exposes CS API, we prefer to avoid that if there's no clear use case

Dalamud/Game/ClientState/Keys/KeyState.cs
```diff
@@ -101,15 +117,52 @@ public bool IsVirtualKeyValid(VirtualKey vkCode)
...
             this[vk] = false;
         }
+
+        Array.Clear(this.extendedKeyState);
     }
 
+    /// <summary>
+    /// Unhooks the WndProc and clears the extended key state buffer.
+    /// </summary>
+    public void Dispose()
```
goaaats: Collapse into a single `IInternalDisposableService.DisposeService()`? This doesn't implement IDisposable

goaaats (changes requested):

Dalamud/Plugin/Services/IKeyState.cs
Bronya-Rand: As for this, Midori in #dalamud-dev asked if it would be possible to basically map the equivalent of a VirtualKey to what the game's own VirtualKey table is and AFAIK the only way to do so would have been to fetch it from CS since it has said info. I at first thought this was gonna be a issue but seeing there was other areas that used CS in Dalamud I thought it would have been fine here. See KeyStateWidget.cs as its where its used.

Haselnussbomber: There are currently 7 cases where Dalamud exposes CS types. Most of them are extension functions and accessible differently. If you don't provide a way to call the function by - for example - having an overload that uses int as type, plugins with custom CS will not be able to call it due to a type conflict.

Bronya-Rand: Adjusted this to return a int which is translated back to `SeVirtualKey` in the widget.

Dalamud/Game/ClientState/Keys/KeyState.cs
Bronya-Rand: Should now support all the flags the game has for extended keys.

goaaats (approved):

PR #2962, opened 2026-09-11
Only allow mainthread access of GameObject properties

Haselnussbomber:
Safeguarding the ObjectTable wasn't enough, since plugins can cache the LocalPlayer's IPlayerCharacter object and can still access its properties in a different thread.

Not sure if this should be API16 or if we want plugins to get thrown in the face now.

See also internal channel, where access to the Name causes a crash now due to the change in #2919:
https://discord.com/channels/581875019861328007/837808848718004245/1548038060518670429

### DIFF
--- a/Dalamud/Utility/ThreadSafety.cs
+++ b/Dalamud/Utility/ThreadSafety.cs
@@ -1,5 +1,7 @@
 using System.Runtime.CompilerServices;
 
+using Dalamud.Configuration.Internal;
+
 namespace Dalamud.Utility;
 
 /// <summary>
@@ -49,6 +51,21 @@ public static class ThreadSafety
 #if DEBUG
         AssertMainThread();
 #endif
+    }
+
+    /// <summary>
+    /// Throws an exception when the current thread is not the main thread and DevMode is enabled.
+    /// </summary>
+    /// <param name="message">The message to be passed into the exception, if one is to be thrown.</param>
+    /// <exception cref="InvalidOperationException">Thrown when the current thread is not the main thread.</exception>
+    [Api16ToDo("Replace all calls with AssertMainThread")]
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void DevModeAssertMainThread(string? message = null)
+    {
+        if (!threadStaticIsMainThread && Service<DalamudConfiguration>.Get().DevMode == true)
+        {
+            throw new InvalidOperationException(message ?? "Not on main thread!");
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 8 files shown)


universalconquistador:
Has Dalamud considered a debug-only assert? That way devs get it when they test locally but it can't introduce new crashes for users

MidoriKami:
With the Dev Mode toggle, maybe that's a somewhat reasonable approach, but we have users that also use the Dev Mode toggle, not sure what to do about them, or just say yolo, you turned on dev mode, you get to deal with it.

universalconquistador:
I was suggesting have the thread checks only throw in Debug builds of the plugin, not anything related to Dev Mode.
Just like how .NET's built in `Debug.Assert` only fires in Debug builds by using
```csharp
[System.Diagnostics.Conditional("DEBUG")]
```

universalconquistador:
But that wouldn't work because the helpers get compiled into Dalamud... 🤔

universalconquistador:
Could also base it on whether the plugin is loaded as a dev DLL and not from a repo

Haselnussbomber:
I don't mind changing it to use DevMode. I wrote the code, it's one push away. I just asked in my PR message about it before doing so.
Well, okay, I didn't mention DevMode, but yeah. I got it ready.

goaaats:
> Could also base it on whether the plugin is loaded as a dev DLL and not from a repo

Finding calling plugins from a call stack is crazy expensive, so this is not feasible without giving every GameObject instance some context about which plugin requested it. 

Generally I'm not crazy about things like this since they make the plugin behave differently based on whether you are a developer or not. It's less predictable and not obvious if you are not aware of the functionality (which is probably just us in this thread).

Not saying that there isn't a tasteful way to do it, but we would have to make it very obvious somehow.

Haselnussbomber:
Well, I guess I'll push the code an we discuss again.

goaaats:
I think dev mode now and unconditionally in 16 would be a good compromise.

Really the problem is that this also affects not your plugins as a developer, so if someone else doesn't actively work on their plugin and you use it, you have to tell them about it and get them to care. Maybe that's a feature though.

Dalamud/Utility/ThreadSafety.cs
```diff
@@ -51,6 +53,21 @@ public static void DebugAssertMainThread()
...
+    /// <summary>
+    /// Throws an exception when the current thread is not the main thread and DevMode is enabled.
+    /// </summary>
+    /// <param name="message">The message to be passed into the exception, if one is to be thrown.</param>
+    /// <exception cref="InvalidOperationException">Thrown when the current thread is not the main thread.</exception>
+    [Api16ToDo("Replace all calls with AssertMainThread")]
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal static void DevModeAssertMainThread(string? message = null)
+    {
+        if (!threadStaticIsMainThread && Service<DalamudConfiguration>.Get().DevMode == true)
```
goaaats, line 65: I guess this is fine because it's a provided service, if it's slow we should just cache it and set it from somewhere on change

goaaats (approved):

